PR #315, opened 2025-08-08
chore: react to web components tile (closed)

devadula-nandan:
ref #290 

This PR replaces all `Tile` instances in the demo with `cds-custom-tile` and `cds-custom-clickable-tile` using lit wrappers.
It also removes `CitationClickableCard`, as its two use cases in the citation card (rendered in Conversational Search) one wrapped with an `<a>` tag and the other with a `<button>` are now handled by `cds-custom-clickable-tile`.

#### Changelog

**New**
Updated import paths to the `Tile`, `ClickableTile` to match the new file.

**Removed**
packages/ai-chat/src/chat/shared/components/responseTypes/util/citations/CitationClickableCard.tsx

#### Testing / Reviewing
select below response types from the dropdown and verify if they render same as base integration branch [preview](https://1.www.s81c.com/common/carbon/ai-chat/tag/latest/demo/index.html), but now with web components.

human agent
card
image
video
audio
conversational search

tour card (not sure how to test this)


devadula-nandan:
> So I think we should probably release some kind of "wrapper" component that applies the styles in `@carbon/ai-chat-components` for the tile(s) so that folks using the tiles inside the chat in their own custom responses can also get the correct styles.

i have closed this pr, and opened a new one #406 that's only scoped to the acceptance criteria of the issue.

Issue #307, opened 2025-08-06
Auto-scroll to top of incoming message, not to the user utterance


ethanwinters:
*Notes*

In `public doAutoScroll` (around here: https://github.com/carbon-design-system/carbon-ai-chat/blob/main/packages/ai-chat/src/chat/shared/containers/MessagesComponent.tsx#L287) we need to make the following change:

`shouldScrollToMessage` should return false for `isRequest*` and true for isResponse, this is the inverse of current behavior. If `message.history.silent` is true for the response (for instance they returned a user_defined response type that just kicks off some behavior and isn't meant to be visible) then we should return false.

`*there is a caveat here if this message is being loaded from history`

The next block with:
```
// Iterate backwards until we find the last message to scroll to. By default, the user's last message should be
// scrolled to. However, if the user's message was silent, the last response should be scrolled to.
```

will need a comment update (and there are other comments that need similar updates in the file), but walking up the message list from the bottom to find the next scrollable message is the correct behavior. However, we will need to deal with the following edge case:

response -> request -> response with history.silent === true. 

I think in the end it might just mean we look at `localMessageItems[localMessageItems.length - 1]` and if its something we should scroll to, we do it, and if its not we don't do any scrolling and we can get rid of the while loop entirely.

There is an asterisk here on if the messages were loaded from history. We should always scroll to the last message we can scroll to if the message came from history. So we might need to keep the while loop for that use case. From history is defined by `message.ui_state_internal?.from_history`. This last message we can scroll to *could* be a request, so something to keep in mind. 

Because we are going to be doing some complicated scroll detection in #308 and #309, we should test is just turning off scroll animation on all requests if it still provides a nice UX: https://github.com/carbon-design-system/carbon-ai-chat/blob/main/packages/ai-chat/src/chat/shared/containers/MessagesComponent.tsx#L403. Should test in chrome and safari. If it doesn't, #309 is going to get more complicated and we should raise as soon as we know,

We should make sure that doAutoScroll runs (throttled) when the chat window is re-sized (this is going to be very important for workspace stuff). There is already a resize observer for this.

*Requirements*
- Scroll to the top of the last response on doAutoScroll instead of the top of the last user utterance
- If the last response is set to silent, just don't do any scrolling automatically.
- Make sure that if the message is coming from history, we scroll to the last message we can scroll to. This could, hypothetically, be a request instead of a response.
- On resize, make sure we call doAutoScroll (throttled)
[...]

PR #355, opened 2025-08-16
fix: add indicator on scroll when messages out of view

### LINKED ISSUE #308: Have some kind of indicator that there is more content below the fold when there is more content.
We don't, yet, have the component to show from @aubrey-oneal but lets assume its one of these https://web-components.carbondesignsystem.com/?path=/docs/experimental-chat-button--overview for this writeup.

https://github.com/carbon-design-system/carbon-ai-chat/issues/307 is a pre-req for this issue.

The logic for this all will take place in https://github.com/carbon-design-system/carbon-ai-chat/blob/main/packages/ai-chat/src/chat/shared/containers/MessagesComponent.tsx. There is already a ref to `this.messagesContainerWithScrollingRef.current` to make use of.

*Notes*
- We don't need this indicator to show up if they are *almost* to the bottom of the page, only if there is a "significant" more amount of content. For instance, we don't want this button to be showing up when they are 1px away from being from the bottom of the content hiding the content they actually want to see. This might be just something we design live with @aubrey-oneal rather than get specs in advance.
- We need to throttle the scroll listener, obviously, for performance and also to avoid the button flashing in and out during auto scrolling. We can't debounce because then we won't get the button to show until auto scrolling has completed, which defeats the purpose. We should make sure we call lodash throttle with `leading: true, trailing: true` to get an initial and final call.

*Requirements*
- If there are more than X pixels left from the current scroll position to the bottom of the content, show the button (likely with z-index to hover above the content).
- This should work while either auto scrolling or manually scrolling.
- Watch for re-sizes of the chat, not just scrolling! There is already a resize observer, to make use of. 307 may make this task free.
- We need to make sure that in small views like the sidebar view the indicator can nicely fit. Its possible that we will need an alternate icon only version of the indicator on very small view sizes.

UzairJ99:
Closes #308 

Requirements:

- If there are more than X pixels left from the current scroll position to the bottom of the content, show the button (likely with z-index to hover above the content).
- This should work while either auto scrolling or manually scrolling.
- Watch for re-sizes of the chat, not just scrolling! There is already a resize observer, to make use of. 307 may make this task free.
- We need to make sure that in small views like the sidebar view the indicator can nicely fit. Its possible that we will need an alternate icon only version of the indicator on very small view sizes.

#### Changelog

**New**

- Temporary down arrow icon for the indicator (until the confirmed design is given from the designers) above the chat messages
- Only appears when the user is X pixels away from the bottom of the chat
- Appears when streaming and the scrollbar is still above the bottom as well as when the user scrolls manually away from the bottom of the chat
- Icon fits in all views nicely - may have to revisit this with the updated design if things change later
- Issue resolution works both with and without the resolution of issue #307 
- Clicking on the indicator button will scroll down to the bottom of the chat through doAutoScroll with options set to {scrollToBottom: 0}

**Changed**

- state is updated with { scrollDown: shouldRender } if the scrollbar is above the threshold X value from the bottom of the chat
- shouldRender determins if the icon should be rendered based on above mentioned threshold value

#### Testing / Reviewing

1. stream messages or code from the options and do not scroll yourself yet (to see its effect)
2. the icon should appear once the scrollbar automatically adjusts itself due to the messages rendering below the visible messages container
3. scroll however you like, the icon will appear whenever you're not at/close-to the bottom
4. clicking on the button will take you to the bottom of the messages container


packages/ai-chat/src/chat/shared/containers/MessagesComponent.tsx
```diff
@@ -959,6 +999,13 @@ class MessagesComponent extends PureComponent<MessagesProps, MessagesState> {
               notifications={notifications}
             />
             {this.renderScrollHandle(false)}
+            {unreadMessages && (
+              <div className="WAC__messages--unreadMessagesIndicator">
+                <div className="WAC__messages--unreadMessagesIndicatorIcon">
+                  <ArrowDown />
```
ethanwinters, line 1005: Should this be a scroll to bottom button?

packages/ai-chat/src/chat/shared/containers/MessagesComponent.scss
```diff
@@ -56,3 +57,21 @@
   block-size: 100%;
   inline-size: 100%;
 }
+
+.WAC__messages--unreadMessagesIndicator {
```
ethanwinters, line 61: We actually have an "unread messages indicator" on the launcher to show you have received messages while you have it minimized. Just to avoid confusion can you make up a different name here?

ethanwinters (commented):
I'm out next week so just left a couple of comments.

ChengTing315 (approved):

OikuraHoro (approved):

Issue #229, opened 2025-07-21
White labeling


ethanwinters:
White labeling has a couple of different tracks.

- [x] Visual theming of the chat
- [x] Replacement of avatar and `watsonx` name

### Visual theming of the chat

- [x] Allow folks to optionally turn off the AI theme (blue gradients)
- [x] Make sure this applies to the AI icon explainability as well.
- [x] Allow folks to provide their own Carbon theme variables like every other Carbon component https://carbondesignsystem.com/elements/color/overview/ (in additions to options where we provide them for sites not on Carbon) - included in https://github.com/carbon-design-system/carbon-ai-chat/pull/367
- [ ] The Chat tokens provided by Carbon are missing the following tokens `$chat-button-text`, `$chat-bubble-user-text`, `$chat-bubble-agent-text` (maybe more) and those need to be added to the core `@carbon/styles` package.
- [ ] Add an example to the demo site showing a "themed" chat.

#### Maybe initial release
1. We already have code in place for a "quick theme" option where folks provide a couple of tokens and then we automagically fill in the other tokens. e.g. "replace all the blue60 tokens with your brands green", etc. Its a light lift because it already exists.

### Later/maybe never
1. Align on what we are doing for typography.

PR #367, opened 2025-08-19
feat: add ability to inherit carbon theme from parent

ethanwinters:
Also, made the switch from cds-custom to cds to make this work, I'll be adding an alternate cds-custom build in a subsequent PR.

Contributes to #229 

Added ability to pick "inherit" as your Carbon Theme. This will mean we just inherit from whatever the parent theme is.

Also, changed from `cds-custom` to `cds-` for our web component usage and will make a new build in rollup like `@carbon/web-components` to handle the `cds-custom` use case.

#### Changelog

**New**

- {{ New thing }}

**Changed**

- {{ Changed thing }}

**Removed**

- {{ Removed thing }}

#### Testing / Reviewing

On the demo site there is now the ability to change the carbon theme applied to the whole page. Make sure that the watsonx icon behaves appropriately in addition to the rest of the elements on the page when switching themes when in "inherit" mode.


devadula-nandan (commented):
| global g10, widget g10 | global white, widget g10 |
|------------------------|--------------------------|
|Screenshot 2025-08-21 at 1 07 29 PM|Screenshot 2025-08-21 at 1 08 18 PM|

since there are 4 supported themes globally, and the widget has 2 (g10, g100), the carbon elements in the widget responds to the global theme, where in the right image the dropdown (and every other carbon elements or layer token usage) layering looks odd.

We may need to find a way to dynamically switch the layer levels of the elements inside the widget, based on global and widget themes.


tiles background isn't showing when widget theme is inherited from parent

Image

packages/ai-chat/src/chat/shared/containers/main/MainWindow.tsx
```diff
@@ -1215,58 +1220,60 @@ class MainWindow
...
-        <div
-          className={cx(
-            "WACMainWindow",
-            "WACWidget__FocusTrapContainer",
-            ...extraClassNames,
-          )}
-          ref={this.mainWindowRef}
-        >
-          {showGlass && <div className="WACWidget__FocusTrapGlass" />}
+        <Layer level={shouldUseLayer ? 1 : 0}>
```
devadula-nandan, line 1223: Perfect! I was thinking the same. just switch the layers once at the parent level. Then we can drop the extra wrappers in the children, and the Carbon components will update naturally.

devadula-nandan (approved):

PR #411, opened 2025-09-05
feat: make config items work like regular props

ethanwinters:
Instead of config items booting up some class, they work like standard props. With this in mind, was able to move a pile of things out of ChatInstance and just into props.

Also flattened props instead of just having one big "config" prop.

Also added a 0.5.0 to 1.0.0 migration guide


ethanwinters:
@annawen1 I'm going to make another commit that adds a bunch more PublicConfig to the demo site so we can test these things better as part of a follow up PR.

PR #414, opened 2025-09-08
chore: adding better debugging for markdown parsing with debug: true

ethanwinters:
In case 0.6.0 doesn't solve folks markdown issues, this will get us better debugging.


devadula-nandan (approved):
LGTM

PR #433, opened 2025-09-13
chore: remove redux dependency

ethanwinters:
Keep running into weird peerDependency issues with folks, trying to lower the peerDependency footprint, especially a weird one like this where there are different major versions of Redux we would want to support.

This is a small and limited subset of what redux/react-redux offer... but don't see why it isn't enough for us? No thunks or anything like that. Just actions, reducer, and useSelector.


packages/ai-chat/src/chat/components-legacy/input/Input.tsx
```diff
@@ -504,7 +504,7 @@ function Input(props: InputProps, ref: Ref<InputFunctions>) {
           </div>
 
           <div className="cds-aichat--input-container__send-button-container">
-            {isStopStreamingButtonVisible && (
+            {(true || isStopStreamingButtonVisible) && (
```
devadula-nandan, line 507:
```suggestion
            {isStopStreamingButtonVisible && (
```

we might want to revert this?
the `stop response` button always renders in the deploy preview.

PR #436, opened 2025-09-15
feat(customSendMessage): update messageSendSource and isSilent

### LINKED ISSUE #369: [Feature Request]: Pass MessageSendSource to customSendMessage
### The problem

Currently, there is no way for us to determine the source of the request in `customSendMessage`, since it only has `request`, `_requestOptions`, and `instance` parameters. It would be helpful to pass in `MessageSendSource`, which can be found in the interface `BusEventSend`.

### The solution

`BusEventSend` should be a parameter `customSendMessage` to provide extra context.

### Examples

_No response_

### Application / product

Business Automation Workflow - Workplace

No

### Business priority

Medium Priority = upcoming release but is not pressing

_No response_

### Additional contacts

Email: tyler.yan@ibm.com

### Available extra resources

_No response_

### Code of Conduct

sangeethababu9223:
Closes #369 

Pass MessageSendSource and "isSilent" to customSendMessage

#### Changelog

**New**

- Add `isSilent` and `busEventSend` as parameters for `customSendMessage`

### DIFF
--- a/packages/ai-chat/src/types/config/PublicConfig.ts
+++ b/packages/ai-chat/src/types/config/PublicConfig.ts
@@ -11,6 +11,7 @@ import { ChatInstance } from "../instance/ChatInstance";
 import { CustomSendMessageOptions } from "./MessagingConfig";
 import { MessageRequest } from "../messaging/Messages";
 import { CornersType } from "./CornersType";
+import { BusEventSend } from "../events/eventBusTypes";
 import type {
   ServiceDesk,
   ServiceDeskFactoryParameters,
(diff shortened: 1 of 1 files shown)


packages/ai-chat/src/types/config/PublicConfig.ts
```diff
@@ -266,6 +267,8 @@ export interface PublicConfigMessaging {
     request: MessageRequest,
     requestOptions: CustomSendMessageOptions,
     instance: ChatInstance,
+    busEventSend?: BusEventSend,
+    isSilent?: boolean,
```
ethanwinters, line 271: i think we should probably add things we want to add into requestOptions

sangeethababu9223: Yes, that makes sense. Thank you!
I've updated it.

PR #440, opened 2025-09-17
feat(human agent): add offset to autoscroll for disconnect banner

### LINKED ISSUE #376: When autoscrolling with human agent, need to add an offset to handle the "disconnect" banner
Image

Autoscroll scrolling too high

sangeethababu9223:
Closes #376

When autoscrolling with human agent, need to add an offset to handle the "disconnect" banner

#### Changelog

**Changed**

- Conditionally adds an offset to the scroll value in `doAutoScroll`. The extra offset only applies when conncted to human agent.


packages/ai-chat/src/chat/shared/containers/MessagesComponent.tsx
```diff
@@ -435,7 +435,17 @@ class MessagesComponent extends PureComponent<MessagesProps, MessagesState> {
             scrollElement,
             setScrollTop,
           );
-          doScrollElement(scrollElement, setScrollTop, 0);
+
+          if (this.props.humanAgentState.activeLocalMessageID) {
+            const aiChatReact = document.querySelector("cds-aichat-react");
+            const humanbanner = aiChatReact.shadowRoot.querySelector(
```
ethanwinters, line 441: Hmm... I'm not sure if just hardcoding in a value or querying the element is a better approach.

We might change the size of the banner.

We might change the classname (in fact, we do change the class name in the integration branch).

I'm not sure if one is more brittle than the other. 

I wonder if the REAL answer here is to move the banner outside of the scrolling div and above it?

ethanwinters (changes requested):

packages/ai-chat/src/chat/shared/containers/MessagesComponent.tsx
sangeethababu9223: Hey @ethanwinters,
I've looked further into this. And I realised that the issue is not caused by the banner.
But beacuse we have:
<pre>
function MESSAGE_TO_AGENT_HELP(): MockStep[] {
  return [
    {
      delay: 0,
      callback: (instance: MockServiceDesk) => {
        instance.sendMessageToUser(
          "***These messages must be sent to an agent and not to the bot.*** \n" +
            HELP_TEXT,
          instance.mockState.currentAgent.id,
        );
        instance.sendMessageToUser(
          HELP_TEXT,
          instance.mockState.currentAgent.id,
        );
      },
    },
  ];
}
</pre>

And the second instance is being considered as a sepearte message from human agent. And the scroll is being focused to that. I tested this out by combining both these messages in a single  instance message and it works.

Also, if we choose `markdown` from human agent options, which has a long text, we can see there is no scroll position issue with that either. If it was related to banner the behaviour should be consistant across all responses.
I noticed `WAC__message--agentMessage WAC__message--withAvatarLine ` class in the second section also in `help` response, which usually doesn't happen in regular response types like `card` or `image`.

This results in `doAutoScroll` function wrongly entering ` if (
            !nextItemInSameMessage &&
            this.shouldScrollToMessage(localItem, message)
          )`
          
          this condition affecting the scroll position.
          
          I've commeted out the second instance method and combined the text to first one for testign purposes. And here it works. Could we do something like this?

Issue #352, opened 2025-08-15
Allow people to make use of our custom tiles as part of their custom responses


ethanwinters:
The tiles used in AI Chat are not core carbon compliant and while Carbon 12 might bring everything into better alignment, for now we need to expose a way for folks to use tiles in their custom responses that match our tiles.

We may just want to provide a `cds-aichat-tile-style-wrapper` (but with a better name?) component that forces our styles on `cds-*-tile` components (and another one for React tiles) rather than making a `cds-aichat-expanded/clickable/etc-tile` and have to deal with worrying about api changes all the time.

Requirements:
- There is a pathway for folks writing custom responses to use tiles that look like ours
- We should also make use of this pathway internally

ethanwinters:
@devadula-nandan do you mind outlining a proposed approach here since you have been deep in this?

devadula-nandan:
In the initial exploration done in [this](https://deploy-preview-394--carbon-ai-chat-components.netlify.app/?path=/story/components-tile-container--default) preview, the most ideal way to avoid tweaking cds-tile too much is to use the following structure:

```
<cds-aichat-tile-container>
  <cds-tile> ... </cds-tile>
  <div class="footer">
    <cds-button>Action 1</cds-button>
  </div>
</cds-aichat-tile-container>
```

The idea is to avoid modifying cds-tile directly and instead provide a container that accepts `cds-tile` as a slot, placing it alongside a footer wrapped together by the custom tile container.

This however requires the styles to be written in `Light DOM` only.
we need to style parts on the slot `cds-clickable-tile` for rounded corners on focus ring. and web components doesn't support part selector on a slotted element 
eg: `::slotted(cds-clickable-tile):part(link)` ❌ won't work. more info in this [thread](https://github.com/w3c/csswg-drafts/issues/3896)
so we can't ship these styles as part of the component.

A dev implementing this as a custom tile response should write these styles [from here](https://github.com/carbon-design-system/carbon-ai-chat/pull/394/files#diff-156865c576bf9e678e5c7b80b9c04d287ea9b3b1784128d30d3946574d0f3834R31) into their code base `Light DOM`, or we can export a `tile-container.scss` or `ai-chat-tile-styles.scss` that can be `use`'d to add the styles.

Another way to style these is by using JavaScript in the updated callback to query the shadowRoot and apply styles, or by injecting the processed CSS into the adopted style sheets in the connected callback.

[...]

PR #450, opened 2025-09-22
feat(tile-container): build tile container in web components storybook

devadula-nandan:
contributes #352 

build to publish `cds-aichat-tile-container` in ai-chat-components
add story showcasing the component

#### Changelog

**New**

- storybook 9 upgrades
- code tab, interactions tab
- new welcome page
- first component `tile-container`
- storybook 9 gives the ability to add transformer for the code tab, which we have used to strip style tags in the code tab.
- added `*.snap.js` to eslint ignore and prettier ignore, as we are noticing a weird problem where lint staged formatting the snap files, causing the snapshot matching to fail.
- added "aiChat:start:storybook" script command to root package.json to make it easier to start the storybook

**Changed**

- modified ci to build chat components before testing
- 

**Removed**

- removed old placeholder extended button

#### Testing / Reviewing

- make sure the components render as expected in the storybook
- review the docs page and the new welcome page
- make sure the interactions for clickable and ai-label variants work as expected


packages/ai-chat-components/src/components/extended-button/src/extended-button.scss
```diff
@@ -10,6 +10,6 @@ $css--plex: true !default;
 @use "../../../globals/scss/vars" as *;
 @use "@carbon/web-components/scss/components/button/button";
 
-:host(#{prefix}-extended-button) {
```
ethanwinters, line 13: we can just delete the whole example "extended-button" now that we have a real component in here

packages/ai-chat-components/src/react/tile-container.ts
```diff
@@ -0,0 +1,22 @@
...
+import React from "react";
+
+// Export the actual class for the component that will *directly* be wrapped with React.
+import CDSAIChatTileContainer from "../components/tile-container/src/tile-container.js";
+
+const TileContainer = createComponent({
+  tagName: "cds-aichat-tile-container",
+  elementClass: CDSAIChatTileContainer,
+  react: React,
+});
```
ethanwinters, line 20: So the built version of this points to "import { createComponent as o } from '../node_modules/@lit/react/create-component.js';"

I think we have to fix up the dependencies or ai-chat-components to include `@lit/react`

packages/ai-chat/package.json
```diff
@@ -46,6 +46,7 @@
     "**/web-components/**/*.js"
   ],
   "devDependencies": {
+    "@carbon/ai-chat-components": "*",
```
ethanwinters, line 49: so something to keep in mind is that @carbon/ai-chat-components is currently "private". When do we make it public even if in beta?

ethanwinters (changes requested):

devadula-nandan:
new updates:
upgraded storybook to 9
added a welcome page (need to review the content, maybe add hero image like what other repos did? designers? assets?)
added some interaction story tests as an exploration (we can disable them if we want)
there could be some issues with icon on image variant which would be discussed with the design later.
lets keep this as our first iteration of ai-chat tile component, and we can improve them with a11y and design inputs gradually.

ethanwinters (approved):

ethanwinters:
Lets merge AFTER @annawen1 cuts 1.0.0.

PR #452, opened 2025-09-22
fix(scroll): disable autoscroll for streaming content

### LINKED ISSUE #309: If I user attempts to manually scroll during auto-scrolling, stop auto scrolling and let them control the scroll until the next sent message resets the process.
https://github.com/carbon-design-system/carbon-ai-chat/issues/307 and https://github.com/carbon-design-system/carbon-ai-chat/issues/308 are pre-reqs for this issue. If possible, we should be using the same scroll event listener as used in #308.

*Notes*

In https://github.com/carbon-design-system/carbon-ai-chat/blob/main/packages/ai-chat/src/chat/shared/containers/MessagesComponent.tsx update a variable when we are actively "auto scrolling". We don't want to confuse a user scroll with an auto scroll.

Since, IRL, when someone scrolls it fires many events, what we want to be doing is capturing scroll events between when the autoscrolls are being fired off and if we get one, turn off autoscrolling until the next message is sent. This is potentially a very very very small window.

Its possible we might need to have a variable on `this` that can immediately update AND have one one state work force re-renders to handle this small window. Also, of course, there are always setState callbacks at our disposal. 

*Requirements*

- In doAutoScroll, just check if auto scrolling is enabled. If not, just return and don't auto scroll.
- In doAutoScroll, set a variable that lets us know that auto scrolling is happening so we don't confuse it with a manual scroll.
- In the scroll listener, if we see a scroll when auto scrolling is not happening, we disable the doAutoScroll
- We re-enable doAutoScroll when a new message comes thru. We should be aware this can happen just by sending a message OR by someone calling `instance.messaging.insertHistory` so we might need to smarten up the logic in `componentDidUpdate`

sangeethababu9223:
Closes #309

This PR solves an issue where content is streaming and the scrollbar continues to shift downwards due to auto scroll. If user tries to scroll upward, the scroll position would "fight" against user as auto scrolling is adjusting the scroll position at the same time. By disabling the autoscroll for streaming content we solve this issue.

We also adjust the scroll positioning for messages so that the last user utterance is always in view. If the last user utterance is only one line, the entire user utterance should be shown. If the user utterance is multiple lines, we show the bottom portion of the utterance if there is not enough scroll space for the bot response.

#### Changelog

**New**

- Added min-height attribute to last scrollable message 
- new constant to capture height of one liner message

**Changed**

- Added `ONE_LINE_MESSAGE_HEIGHT` to `setScrollTop` position calculation to ensure full or part of the last user utterance is shown.

**Removed**

- Removed doautoscroll for  `<StreamingRichText>`.

#### Testing / Reviewing

- Check the demo and test all the dropdown options to ensure the user message is shown (full or at least partial message)
- for the streaming options, test by manually scrolling up when content is streaming. The scroll should go up and stay in the position you stop scrolling at instead of shifting

### DIFF
--- a/packages/ai-chat/src/chat/shared/containers/MessageComponent.scss
+++ b/packages/ai-chat/src/chat/shared/containers/MessageComponent.scss
@@ -55,6 +55,10 @@ $message-bar-width: 3px;
   inset-block-start: 0.5rem;
 }
 
+.WAC__message--withAvatarLine.WAC__message--lastMessage {
+  min-block-size: 100%;
+}
+
 .WAC__message.WAC__message--lastMessage.WAC__message--has-focus::after {
   block-size: calc(100% - #{layout.$spacing-07} + 0.5rem);
   inset-block-end: calc(#{layout.$spacing-07} - 0.5rem);
--- a/packages/ai-chat/src/chat/shared/containers/MessageComponent.scss
+++ b/packages/ai-chat/src/chat/shared/containers/MessageComponent.scss
@@ -1,6 +1,6 @@
-/* 
+/*
  *  Copyright IBM Corp. 2025
- *  
+ *
  *  This source code is licensed under the Apache-2.0 license found in the
  *  LICENSE file in the root directory of this source tree.
  */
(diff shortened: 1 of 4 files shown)


ethanwinters (approved):
such a smart implementation. super A+++. Thank you!

PR #500, opened 2025-10-01
fix(rtl): text alignment 

### LINKED ISSUE #486: RTL bugs
This is an issue to encompass various RTL bugs found while testing 1.0.0:

- [ ] inline error response
- [ ] ordered/unordered lists
- [ ] feedback panel

annawen1:
Closes #486 

In RTL, there are text alignment issues especially visible in full screen mode.

Screenshot 2025-10-01 at 11 55 23 AM

Screenshot 2025-10-01 at 2 14 30 PM

We have styles that are setting the `text-align` attribute when in RTL, which is overriding the browsers natural text alignment for components like ordered-list and regular text.

#### Changelog

**Removed**

- remove styles that are setting the `text-align` attribute.

#### Testing / Reviewing

1. Go to demo deploy preview
2. Set full screen mode and open chat
3. Set direction to RTL 
4. Test with `inline error`, `ordered list`, and `unordered list` from dropdown options

### DIFF
--- a/packages/ai-chat/src/chat/components-legacy/App.scss
+++ b/packages/ai-chat/src/chat/components-legacy/App.scss
@@ -1,6 +1,6 @@
-/* 
+/*
  *  Copyright IBM Corp. 2025
- *  
+ *
  *  This source code is licensed under the Apache-2.0 license found in the
  *  LICENSE file in the root directory of this source tree.
  */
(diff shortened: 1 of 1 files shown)


annawen1:
> Everything looks good to me.
> 
> Not specific to the issue but
> 
>  Screen.Recording.2025-10-03.at.5.33.52.PM.mov 
> This looks much better if we add `align-items: center` to it

Oh good catch! With `align-items: center`, it looks old if there's more than one line in the error message so instead I matched the RTL margin-top style to the same as what we have for the LTR inline-error-icon margin styles

Screenshot 2025-10-03 at 10 52 06 AM

Screenshot 2025-10-03 at 11 02 58 AM

PR #502, opened 2025-10-01
feat: add config item for input config

ethanwinters:
initial config item is only controlling maxInputLength

Be sure to test both react and web component versions.

http://localhost:3001/?config=setChatConfig

And then set `window.setChatConfig({ input: { maxInputCharacters: 10 }})` in the console and only 10 characters will be allowed in the input field. 

And then do same for web component version: http://localhost:3001/?settings=%7B%22framework%22%3A%22web-component%22%2C%22layout%22%3A%22float%22%2C%22writeableElements%22%3A%22false%22%2C%22direction%22%3A%22default%22%7D&config=setChatConfig


annawen1 (approved):
LGTM!

Issue #492, opened 2025-10-01
[Bug]: Launcher button is broken in mobile device


sangeethababu9223:
### Browser

Chrome, Safari

### Package version

1.0.0

### Description

The launcher button is broken in mobile device:

![Image](https://github.com/user-attachments/assets/b217a950-d345-452c-bbc0-d47d9344fa02)

### Reproduction / example

https://chat.carbondesignsystem.com/tag/next/demo/index.html

### Steps to reproduce

Open https://chat.carbondesignsystem.com/tag/next/demo/index.html in a mobile device in chrome browser and this issue can be spotted.

### Suggested severity

None

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/main/.github/CODE_OF_CONDUCT.md)
- [ ] I checked the [current issues](https://github.com/carbon-design-system/carbon-ai-chat/issues) for duplicate problems

ethanwinters:
If the previous demo site doesn't work properly, ping Ethan for how this is supposed to look

PR #510, opened 2025-10-02
fix: mobile css bugs

### LINKED ISSUE #495: [Bug]:  In Web Components framework Launcher button doesn't appear for Full Screen / Full w/hasContentMaxWidth: false options
### Browser

_No response_

### Package version

1.0.0

_No response_

### Description

When `Component framework` is chosen as `Web component` and `Layout` option is either `Fullscreen` or `Full w/hasContentMaxWidth: false`, the launcher button disappears. 
during initial load when the default layout is `Float`, the launcher button is there. But when you switch to `Sidebar` and then to `Fullscreen` or `Full w/hasContentMaxWidth: false` the button disappears.  And if you switch back to `Float` again the launcher button is still missing.

And from here if we switch back franework to `React` the launcher is still missing.

### Reproduction / example

https://chat.carbondesignsystem.com/tag/next/demo/index.html

### Steps to reproduce

In the demo app, switching framework to `Web Component` and choosing the options mentioned in the description will re create the issue.

### Suggested severity

None

### Application / product

_No response_

### Code of Conduct

- [ ] I checked the [current issues](https://github.com/carbon-design-system/carbon-ai-chat/issues) for duplicate problems

ethanwinters:
In "mobile" the chat is supposed to just take over the full screen when opened. We had a tiny CSS bug that was preventing that which led to a pile of side effect issues.

Closes #495 #492


devadula-nandan:
tested on ios and this is how it looks compared to previous version
| before | after |
|---|---|
| IMG_2897 | <video src="https://github.com/user-attachments/assets/c04cc8e6-477d-468e-ac03-dd244be98f9c"/> |

the input field is hidden for my view port, is it same for u too? do we subtract n pixels from 100vh ?

devadula-nandan (commented):
tested on ios and this is how it looks compared to previous version
| before | after |
|---|---|
| IMG_2897 | <video src="https://github.com/user-attachments/assets/c04cc8e6-477d-468e-ac03-dd244be98f9c"/> |

the input field is hidden for my view port, is it same for u too? do we subtract n pixels from 100vh ?

PR #517, opened 2025-10-06
fix: mobile launcher

ethanwinters:
The MOBILE launcher is no longer broken. You can just test by turning on "mobile" in chrome dev tools. Wait the 15 seconds or so for the mobile CTA to open.


ethanwinters:
added to https://github.com/carbon-design-system/carbon-ai-chat/pull/510

PR #544, opened 2025-10-10
fix(avatarline): assistantName does not update

### LINKED ISSUE #470: [BUG]: setting the `assistantName` in the config doesn't appear to change the avatar line
When setting the config in the demo
```
await window.setChatConfig({
   assistantName: "myBot",
});
```

The header title changes however in the avatar line, the `assistantName` doesn't appear. `ResponseUserProfile` (what you put on the actual message) is first priority, then it should be `assistantName`, then `watsonx`

Image

Looking at https://github.com/carbon-design-system/carbon-ai-chat/blob/main/packages/ai-chat/src/chat/components-legacy/MessageComponent.tsx#L388-L392, the logic populates the `responseUserProfile` with the default `watsonX` before there is a chance to check if `assistantName` exists

sangeethababu9223:
Closes #470

setting the assistantName in the config doesn't appear to change the avatar line

#### Changelog

**Changed**

- In `packages/ai-chat/src/chat/components-legacy/MessageComponent.tsx`, `actorName` is set to   `assistantName || responseUserProfile?.nickname ` instead of `responseUserProfile?.nickname || assistantName `

#### Testing / Reviewing

In local, add  the following to `demo/src/framework/demo-body.tsx`
<pre>
window.setChatConfig({
 assistantName: "myBot",
});
</pre>

### DIFF
--- a/packages/ai-chat/src/chat/components-legacy/MessageComponent.tsx
+++ b/packages/ai-chat/src/chat/components-legacy/MessageComponent.tsx
@@ -414,7 +414,10 @@ class MessageComponent extends PureComponent<
         iconClassName = "cds-aichat--message__avatar--agent";
         actorName = responseUserProfile?.nickname || "";
       } else {
-        actorName = assistantName || responseUserProfile?.nickname;
+        actorName =
+          responseUserProfile?.nickname === "watsonx" && assistantName
+            ? assistantName
+            : responseUserProfile?.nickname;
 
         let icon = <IconHolder icon={<ChatBot />} />;
 


sangeethababu9223:
> So we do want nickname to beat assistantName if it is set, the problem is in `const responseUserProfile: ResponseUserProfile = message.message_options ?.response_user_profile || { id: "watsonx", nickname: "watsonx", user_type: UserType.WATSONX, };` when we default it to watsonx without checking if another assistantName was set first.
> 
> ```
>   If nothing is set -> watsonx
>   If nickname is set -> nickname
>   If assistantName is set and nickname is not set -> assistantName
> ```

Thank you for pointing this out @ethanwinters.
I've updated the condition accordingly.
For 
<pre>
if (
        responseUserProfile.profile_picture_url &&
        responseUserProfile.user_type !== UserType.WATSONX
      )
</pre>

It was already set as `actorName = responseUserProfile?.nickname || "";`

I've updated this for rest as
<pre>
 actorName =
          responseUserProfile?.nickname === "watsonx" && assistantName
            ? assistantName
            : responseUserProfile?.nickname;
</pre>

PR #568, opened 2025-10-16
fix: html parsing

### LINKED ISSUE #567: [Bug]: HTML content for GenericItem is not rendered properly
### Browser

_No response_

### Package version

v1.0.0

v18.3.1

### Description

After migrating from v0.3.2 to v1.0.0, html content for GenericItem (either TextItem or CardItem) is no longer rendering as it previously did, resulting in the text appearing differently. There is also problem rendering carbon web component such as <cds-link> inside text.

Slack thread: https://ibm-systems-z.slack.com/archives/C07RV4MPJNL/p1760623768328569

### Reproduction / example

N/A

### Steps to reproduce

Create a TextItem or CardItem whose text contains a string that has html snippet in the middle or contains carbon web component such as <cds-link>

### Suggested severity

None

### Application / product

_No response_

### Code of Conduct

ethanwinters:
closes #567 

So when we get markdown we parse it with markdown-it. This gives us a tree of all the content. When markdown-it comes across inline html, it creates a token for the opening tag, a token for any inner text, and then a token to close it. Our problem is that unless we combing all those tokens into one token before we render that Lit can "helpfully" auto close the first html token. Which can lead to the following:

```
hello <b></b>bold.
```

instead of 

```
hello <b>bold</b>
```

as the `<b>` gets auto closed and then the trailing `</b>` just gets removed because it doesn't close anything.

We had logic in here to try to fix that already, but it was too strict in its checking and was only working if the first token received was an html tag, not if it occurred somewhere in the middle.

Added a block that starts with plain text at the begining of the "HTML" response in the dropdown of the demo site for testing.

### DIFF
--- a/demo/src/customSendMessage/constants.ts
+++ b/demo/src/customSendMessage/constants.ts
@@ -127,6 +127,8 @@ ${UNORDERED_LIST}
 `;
 
 const HTML = `
+Here is some <b>Carbon</b> information peppered with HTML!
+
 <style>
     .fun-fact { 
         background-color: #e8f5e8; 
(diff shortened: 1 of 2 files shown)


annawen1 (approved):
LGTM!

Issue #589, opened 2025-10-23
ChatActionsImpl refactor


ethanwinters:
ChatActionsImpl has too many methods that are 100-200 lines and have multiple responsibilities. We need to split up some of those methods into smaller discrete pieces with single responsibility. Methods are currently correctly marked as "private" or not and there is full test coverage. The goal would be to not change the external APIs, just add more smaller private methods and only have changes to ChatActionsImpl.

PR #590, opened 2025-10-23
fix: Restart conversation while streaming

### LINKED ISSUE #588: [Bug]: conversation restart streaming issues
### Browser

_No response_

### Package version

1.1.0

_No response_

### Description

If you enable the restart button on the demo site and then try to restart the conversation WHILE you are streaming back text by entering something like `text (stream)` you will see that the cancelled message will still display any chunks that come thru before the message is actually aborted and those chunks will not be cleared out as part of the message restart process.

### Reproduction / example

https://github.com/user-attachments/assets/6f864a44-8a49-421f-b506-96cf987f0728

### Steps to reproduce

Enable restart button on demo site
Type "text (stream)"
Hit restart button
See that inflight message is not removed

### Suggested severity

None

### Application / product

_No response_

### Code of Conduct

ethanwinters:
Closes #588 

If the conversation is actively being restarted, holds any received chunks from active messages from being processed so they don't get displayed and don't get added to the message list after it has been cleared.

There are two commits in the PR. The first adds a reducer to track if we are actively restarting and then has some minor refactoring of the tests to share a single shared helper file instead of having multiple shared helpers coming from different test folders.

The second commit adds a messageGenerations map where we basically hold message ids tied to a restart count. If a message is attached to a previous restart count we know to ignore it. We also store the abort controllers in MessageService because we can't guarantee the this.queue.current will still have the old messages anymore after a restart. There is clean up on this list, but I still have to do some work to add cleanup to the messageGenerations map when that data gets old and stale (e.g. we don't need to keep the message list from 4 restarts ago anymore) but im going to handle that in the next PR for #589 

#### Changelog

The bulk of the changes are in ChatActionsImpl where we now track if we are in a "restarting mode" and make sure we don't fully process the chunks when we are in that mode. That class has turned into a bear over time, created https://github.com/carbon-design-system/carbon-ai-chat/issues/589 to try to get some of these large multi purpose methods to be smaller pieces with single responsibility that are easier to understand.

Also, since the restart button handling is now async, disable the restart button while a restart is ongoing.

#### Testing / Reviewing

Go to current demo site and look at the old behavior when you complete the following steps:

Enable restart button on demo site
Type "text (stream)"
Hit restart button
See that inflight message is not removed

Then do the same in this branch. Also follow the same process with `code (stream)` and then mix them up. `code (stream)` is setup to NOT return a promise from customSendMessage and `text (stream)` does return a promise from customSendMessage. All our examples have you `await` customSendMessage, but you don't HAVE to and there was some extra work to make sure that worked.

### DIFF
--- a/packages/ai-chat/src/chat/components-legacy/input/Input.tsx
+++ b/packages/ai-chat/src/chat/components-legacy/input/Input.tsx
@@ -514,7 +514,7 @@ function Input(props: InputProps, ref: Ref<InputFunctions>) {
                     type: BusEventType.STOP_STREAMING,
                   });
                   // Also cancel the current message request to abort the signal
-                  serviceManager.messageService.cancelCurrentMessageRequest();
+                  await serviceManager.messageService.cancelCurrentMessageRequest();
                 }}
               />
             )}
--- a/packages/ai-chat/src/chat/events/ChatActionsImpl.ts
+++ b/packages/ai-chat/src/chat/events/ChatActionsImpl.ts
@@ -134,6 +134,18 @@ class ChatActionsImpl {
   private alreadyHydrated = false;
 
   /**
+   * Tracks the current restart generation. This is incremented each time restartConversation() is called
+   * and is used to filter out stale chunks from previous conversation generations.
+   */
+  private restartGeneration = 0;
+
+  /**
+   * Maps message IDs to the generation they were created in. This is used to filter out chunks from
+   * messages that were started before a restart.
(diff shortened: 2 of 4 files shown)


annawen1 (commented):
@ethanwinters Noticing that when I restart the conversation on a streaming option, it works, but if I select another streaming option right after, the streaming response doesn't appear

annawen1 (approved):
working great now!

PR #613, opened 2025-10-28
fix: add disclaimer accepted event

### LINKED ISSUE #611: disclaimer event missing
we need an event for the disclaimer so folks can save the state long term and not re-present

ethanwinters:
Closes #611 

Now fires an event when folks accept the disclaimer so people can save that state

#### Testing / Reviewing

On the demo site, search for `instance.on({ type: BusEventType.FEEDBACK, handler: feedbackHandler });` throw in a `instance.on({ type: BusEventType.DISCLAIMER_ACCEPTED, handler: () => console.log('Disclaimer accepted')});` there and then accept the disclaimer from the demo site and check the console


annawen1 (approved):
LGTM

PR #615, opened 2025-10-28
fix: no accordion on empty cot

ethanwinters:
Closes #604 

If a chain of thought step is empty, no need to show an accordian

#### Testing / Reviewing

The "text with chain of thought" on the demo site now includes a step with no inner data and it should be shown with no accordion to open.


annawen1 (approved):
LGTM!

PR #621, opened 2025-10-30
chore: Remove chat actions impl

ethanwinters:
Contributes to #589 

I split up the ChatActionsImpl.ts from being a 1500+ mess to a bunch of different services. Tests are still needed and I probably want to clean up those new services more, but I won't be able to loop back to this for a few sprints and keeping the branch up to date was getting difficult.

There isn't any new functionality here, just splitting up that class, so a simple regression test on the UI is the testing requirement.


devadula-nandan:
there seems to be some conflicts 👀

devadula-nandan (approved):

PR #622, opened 2025-10-30
chore: move cot/feedback to caic components

ethanwinters:
Just moving the cot and feedback components to the correct repo. No storybooks or anything, yet. Next PR is going to be a pile of storybooks!


packages/ai-chat-components/src/components/chain-of-thought/src/ChainOfThoughtElement.ts
```diff
@@ -7,25 +7,34 @@
  *  @license
  */
```
devadula-nandan: do we keep all the file names kebab case in here? and feedback?

devadula-nandan (approved):
looks good, one comment.

packages/ai-chat-components/src/components/chain-of-thought/src/ChainOfThoughtElement.ts
ethanwinters: good call, I swap it over when i write the storybooks

PR #651, opened 2025-11-06
feat: hide send button when stop button is rendered

### LINKED ISSUE #632: [Feature Request]: Disable the send message button until current response is finished processing
### The problem

Subsequent inputs during response processing causes weird behavior and responses to come out of place. Ideally, no messages should be able to be sent if response is still processing (this is the behavior of most other chatbots out there). Disabling the input field with `instance.updateInputIsDisabled(true);` is not the most optimal solution as user should still be expecting to be able to type just not able to send the message.

Image

### The solution

We should prevent user from sending anything in the chat (by disabling the send button) until the current response is finished processing. Ideally, no messages should be able to be sent if response is still processing meanwhile the user should still be able to type in the input field (this is the behavior of most other chatbots out there).
https://ibm-cloud.slack.com/archives/C07RV4MPJNL/p1761680636293349?thread_ts=1761663294.606119&cid=C07RV4MPJNL

### Examples

Image

### Application / product

Business Automation Workflow

No

### Business priority

None

_No response_

### Additional contacts

_No response_

### Available extra resources

_No response_

### Code of Conduct

devadula-nandan:
Closes #632 

this pr conditionally renders send button based on `isStopStreamingButtonVisible` prop.
and also focuses the input on stop button interaction.

#### Changelog

**Changed**

- hides send button if `isStopStreamingButtonVisible` is true
- focus input field on stop button interaction

#### Testing / Reviewing
Try to interact with input field and try an asynchronous response type.
the send button is not rendered. and only user action available is to stop the response

This behavior can be compared with other popular chat applications

https://github.com/user-attachments/assets/eb98cba4-8f16-4a35-a462-0f157a4ff2dd

### DIFF
--- a/packages/ai-chat/src/chat/components-legacy/input/Input.tsx
+++ b/packages/ai-chat/src/chat/components-legacy/input/Input.tsx
@@ -510,7 +510,7 @@ function Input(props: InputProps, ref: Ref<InputFunctions>) {
               <StopStreamingButton
                 label={input_stopResponse}
                 disabled={isStopStreamingButtonDisabled}
-                tooltipAlignment="top"
+                tooltipAlignment={isRTL ? "top-left" : "top-right"}
                 onClick={async () => {
                   const { store } = serviceManager;
                   store.dispatch(actions.setStopStreamingButtonDisabled(true));


packages/ai-chat/src/chat/components-legacy/input/Input.tsx
```diff
@@ -502,11 +506,11 @@ function Input(props: InputProps, ref: Ref<InputFunctions>) {
           </div>
 
           <div className="cds-aichat--input-container__send-button-container">
-            {isStopStreamingButtonVisible && (
+            {isStopStreamingButtonVisible ? (
```
ethanwinters: we should put focus onto the stop streaming button when it is made visible so screen reader users are made aware of it.

Approved with that one change.

ethanwinters: Oh, we should also make sure that when the stop button disappears on its own that we put focus somewhere. That might already be handled by just receiving a message, but double check.

devadula-nandan: i just cross referenced `chatGPT`, `Gemini`, to see what they were doing. and in both cases the input field stays focused once the user message is sent. which is exactly what we already have.
is the suggested behavior flagged by any A11y audit?
i just want to re-confirm if we want to have this behavior before i make any changes.

PR #668, opened 2025-11-10
feat: processing message indicator

### LINKED ISSUE #633: [Bug]: RESTART_CONVERSATION doesn't reset the message loading counter
### Browser

_No response_

### Package version

v1.0.0

_No response_

### Description

`RESTART_CONVERSATION` should have the effect of clearing the chat state including any displayed message loading counters, currently it does not and it persists after restart

### Reproduction / example

N/A

### Steps to reproduce

- Update the counter via `instance.updateIsMessageLoadingCounter('increase')`
- Restart the converstation
- Observer that the message loading indicator is still displayed

### Suggested severity

None

### Application / product

_No response_

### Code of Conduct

ethanwinters:
adds an optional message to go with the processing element

Closes #633 

Adds the following set of features:

- Adds optional string to pass into the loading indicator to show next to the three dots
- Exposed the current loading indicator states in getState
- Adds a new "reset" option to pass into the loading indicator to set the count back to 0
- Conditionally clears out loading indicators on conversation restart.
- Add a new option to the demo site to add a loading indicator with a message

### DIFF
--- a/packages/ai-chat/src/types/instance/ChatInstance.ts
+++ b/packages/ai-chat/src/types/instance/ChatInstance.ts
@@ -67,7 +67,7 @@ export type PublicChatState = Readonly<
     /**
      * Optional string to display next to the loading indicator.
      */
-    isMessageLoadingMessage?: string;
+    isMessageLoadingText?: string;
 
     /**
      * Counter that indicates if the chat is hydrating and a full screen loading state should be displayed.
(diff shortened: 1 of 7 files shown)


packages/ai-chat/src/types/instance/ChatInstance.ts
```diff
@@ -47,13 +47,32 @@ export interface ChatInstance extends EventHandlers, ChatActions {
...
+    /**
+     * Counter that indicates if a message is loading and a loading indicator should be displayed.
+     * If "0" then we do not show loading indicator.
+     */
+    isMessageLoadingCounter: number;
+
+    /**
+     * Optional string to display next to the loading indicator.
+     */
+    isMessageLoadingMessage?: string;
```
annawen1: Finding the `isMessageLoadingMessage` name a bit confusing, could we rename to something like `isMessageLoadingText` or `loadingIndicatorText`?

PR #673, opened 2025-11-12
feat(storybook): add rtl switcher

### LINKED ISSUE #562: [Feature Request]: add rtl toggle in storybook toolbar
### The problem

we need rtl control in storybook toolbar to test our components in rtl layout

### The solution

https://github.com/carbon-design-system/ibm-products/issues/6748#issuecomment-2639416337

### Examples

_No response_

### Application / product

_No response_

None

### Business priority

None

_No response_

### Additional contacts

_No response_

### Available extra resources

_No response_

### Code of Conduct

sangeethababu9223:
Closes #562

add rtl toggle in storybook toolbar

#### Changelog

**New**

-  Added `dir` option to `packages/ai-chat-components/.storybook/preview.js`
- Enabled `lang` switching by adding `document.documentElement.lang = locale;` to preview file.

#### Testing / Reviewing

- Text direction switcher is available in the storybook toolbar now. And  the direction can be switched with this.
- The language switch will be refelcted based on the chosen language from the switcher.

### DIFF
--- a/.github/workflows/release-base.yml
+++ b/.github/workflows/release-base.yml
@@ -89,7 +89,10 @@ jobs:
         if: |
           github.event.inputs.type == 'full release' &&
           github.event.inputs.semver-type == 'minor'
+        # pass in an empty commit so lerna will graduate the version to stable
+        # even though there are no new changes from last rc
         run: |
+          git commit --allow-empty -m "chore(release): graduate prerelease to stable"
           npx lerna publish minor --conventional-graduate ${{ env.PUBLISH }} ${{ env.FORCE }}
 
       - name: Publish first minor RC (ie. v1.x.0-rc.0)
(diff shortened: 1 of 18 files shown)


annawen1 (changes requested):
Looks good, can we also apply the same to the React wrapper storybook? Also do we need the language option? I'm currently not seeing anything change when I'm switching between the two options.

sangeethababu9223:
> Looks good, can we also apply the same to the React wrapper storybook? Also do we need the language option? I'm currently not seeing anything change when I'm switching between the two options.

Hey @annawen1 ,
Thank you for pointing out the missing text direction in React storybook. I've updated that now.
For the language switcher, it was already there in storybook toolbar but the switching wasn't happening. I only enabled the switching. 
Should we just remove the switcher from storybook toolbar altogether?

PR #684, opened 2025-11-14
feat(preview card): new response type to trigger workspace

### LINKED ISSUE #639: Make new response type for workspace preview card.
We need to create a new "workspace preview" response type using the preview card that @devadula-nandan already created. When clicked, it should throw an event that folks will use to open the workspace panel.

We should probably setup and integration branch for workspace stuff since we don't want to release a response type that doesn't do anything.

**Acceptance criteria**

A new preview workspace response type that fires a new open workspace event on click. 

**Types**

Note, all these types generate the docs so write great ts comments on them.

We need to add the enum: https://github.com/carbon-design-system/carbon-ai-chat/blob/main/packages/ai-chat/src/types/messaging/Messages.ts#L218

Make it part of generic item: https://github.com/carbon-design-system/carbon-ai-chat/blob/main/packages/ai-chat/src/types/messaging/Messages.ts#L646

Define the actual data for the response type: https://github.com/carbon-design-system/carbon-ai-chat/blob/main/packages/ai-chat/src/types/messaging/Messages.ts#L762

We sure be sure to have something generic like additional_data to allow folks to pass in whatever random data they need to open a workspace: https://github.com/carbon-design-system/carbon-ai-chat/blob/main/packages/ai-chat/src/types/messaging/Messages.ts#L801C3-L803C5

Of note, just using the response type isn't going to be able to open up every possible action in the toolbar, we only will really support the "open" action with the OOTB response type. For them to do fancier stuff, they will need to use a user_defined response and compose things themselves. 

**Registration and Display**

That happens in: https://github.com/carbon-design-system/carbon-ai-chat/blob/main/packages/ai-chat/src/chat/components-legacy/MessageTypeComponent.tsx

Import all the types and the component and add it to the switch here: https://github.com/carbon-design-system/carbon-ai-chat/blob/main/packages/ai-chat/src/chat/components-legacy/MessageTypeComponent.tsx#L241

**Throwing an event on click**

Add the new event to the enum: https://github.com/carbon-design-system/carbon-ai-chat/blob/main/packages/ai-chat/src/types/events/eventBusTypes.ts#L31

Define the event: https://github.com/carbon-design-system/carbon-ai-chat/blob/main/packages/ai-chat/src/types/events/eventBusTypes.ts#L333

You fire the event with `serviceManager.fire` and you can search for the various implementations.

**Preview card updates**

We should make use of the new toolbar components inside the preview card.

devadula-nandan:
Closes #639 

rounded border radius is fixed in #744

adds a new response type for preview card with `title` `subtitle` and `additional data`

#### Changelog

**New**

- response type preview card

#### Testing / Reviewing

interact with the button and notice the events firing in console. note that this needs to be further attached to opening and closing of workspace content.

### DIFF
--- a/packages/ai-chat/src/types/events/eventBusTypes.ts
+++ b/packages/ai-chat/src/types/events/eventBusTypes.ts
@@ -680,19 +680,10 @@ export interface BusEventWorkspacePreOpen extends BusEvent {
     message: GenericItem;
 
     /**
-     * The full message (response or request) that contains the message item.
+     * The full message response that contains the message item.
      */
-    fullMessage: Message;
-
-    /**
-     * The slot name for users of the web components cds-aichat-container or cds-aichat-custom-element.
-     */
-    slot?: string;
+    fullMessage: MessageResponse;
   };
-  /**
-   * The source of the message being sent.
-   */
-  source: MessageSendSource;
 }
 
 /**
@@ -707,19 +698,10 @@ export interface BusEventWorkspaceOpen extends BusEvent {
     message: GenericItem;
 
     /**
-     * The full message (response or request) that contains the message item.
+     * The full message response that contains the message item.
      */
-    fullMessage: Message;
-
-    /**
-     * The slot name for users of the web components cds-aichat-container or cds-aichat-custom-element.
-     */
-    slot?: string;
+    fullMessage: MessageResponse;
   };
-  /**
-   * The source of the message being sent.
-   */
-  source: MessageSendSource;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/ai-chat/src/types/events/eventBusTypes.ts
```diff
@@ -673,7 +673,22 @@ export interface BusEventCustomPanelClose extends BusEvent {
...
+  data: {
+    /**
+     * The individual message item that is being displayed in this custom response.
+     */
+    message: GenericItem;
+
+    /**
+     * The full message (response or request) that contains the message item.
+     */
+    fullMessage: Message;
```
ethanwinters: I think this would always be a MessageResponse vs a Message

packages/ai-chat/src/types/events/eventBusTypes.ts
```diff
@@ -697,7 +727,22 @@ export interface BusEventWorkspaceOpen extends BusEvent {
...
```
ethanwinters: ditto!

packages/ai-chat/src/serverEntry.ts
```diff
@@ -74,6 +74,10 @@ export {
   BusEventCustomPanelOpen,
   BusEventCustomPanelPreClose,
   BusEventCustomPanelPreOpen,
+  BusEventWorkspaceClose,
```
ethanwinters: Interesting on the close events. My first thought was "if they are in charge of closing the workspace panel, what is this even for?" but I guess there are times when we might close it for them. Like if someone restarts the conversation.

ethanwinters (approved):
looks great, just a couple of comments

Issue #697, opened 2025-11-18
[Bug]: Table default pagination is incorrect


nelsonchen90:
### Browser

Chrome

### Package version

1.2.1

### React version (if applicable)

17.0.2

### Description

Rendering a table with markdown format. The default pagination is incorrectly set. i.e the pagination component shown it's set to 10 items per page but it's only showing 5 items (even though there are more)

<img width="774" height="595" alt="Image" src="https://github.com/user-attachments/assets/6acd85bd-1cde-437f-8f47-092dbd7e7a12" />

### Reproduction / example

https://chat.carbondesignsystem.com/tag/latest/demo/index.html

### Steps to reproduce

It's in the demo site.

In the demo, select `table` from the dropdown
Notice the table is rendered with 5 rows but the paging component shows 10
<img width="774" height="595" alt="Image" src="https://github.com/user-attachments/assets/6acd85bd-1cde-437f-8f47-092dbd7e7a12" />

And if user change it around it will render correctly:
<img width="680" height="669" alt="Image" src="https://github.com/user-attachments/assets/b7ac6764-b310-4996-a6b7-b8d8dbf9dea8" />

### Suggested severity

Severity 3 = User can complete task, and/or has a workaround within the user experience of a given component.

### Application / product

watsonx.data intelligence

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/carbon-ai-chat/issues) for duplicate problems

PR #700, opened 2025-11-20
feat: release ai-chat-components beta

### LINKED ISSUE #343: [Epic]: Move first pass of components into @carbon/ai-chat-components
### Value proposition

This is an epic for the highest value components to release as standalone components.

### Pass 1.1.
Component: cds-aichat-markdown-text, cds-aichat-markdown-table, cds-aichat-markdown-code
Reason: These components are required for reasoning steps component

### No current priority

Component: header
Reason: Folks want to pass in their own header (or no header) into the chat component. Having the header component with all the options will allow them to compose it how they want.

Component: Prompt line
Reason: Folks want to pass in their own prompt line into the chat component. Having the prompt line component with all the options will allow them to compose it how they want.

Component: audio/video
Reason: We bring in the react-player component for audio/video and we want to remove that dependency. Most of the logic we are using in react-player is to detect what service the audio/video is using and then embed with their iframe technique. None of that requires React and we can lift code from react-player and be sure to include their MIT license information in the file.

Component: tile
Reason: We have specific style overrides for the tile component. We need to provide a way for folks to match that. We may provide a wrapper for tile and clickable tile or just instructions on how to override the tile to fit our styles (with any needed style exports)

### Expected outcomes

Unblock: https://github.com/carbon-design-system/carbon-ai-chat/issues/266 https://github.com/carbon-design-system/carbon-ai-chat/issues/314 and https://github.com/carbon-design-system/carbon-ai-chat/issues/342 and allow folks to extend the chat with custom responses and have them match

### Details

_No response_

### Preview

_No response_

### Code of conduct

- [x] I agree to follow this project's [code of conduct](https://github.com/carbon-design-system/carbon/blob/main/.github/CODE_OF_CONDUCT.md)

ethanwinters:
Closes #343

This integration branch has had all individual commits reviewed.


annawen1 (approved):

annawen1:
the lazy loading feature for table seems to have fixed the issue with the visible table rows on render: https://github.com/carbon-design-system/carbon-ai-chat/issues/697

PR #709, opened 2025-11-24
feat: improve mobile scrolling

ethanwinters:
The previous CSS only solution led to some strange results in Chrome for iOS including not being able to see the input field, and the ability the break everything across browsers by "overscrolling".

This PR introduces a new hook to manage freezing body scroll styles, and set the visual view port size based on JS measurements instead of CSS properties which could be unreliable on their own (especially in Chrome for iOS).

It also includes a change to generate isBrowser via function instead of a constant as it seemed like something were getting treeshaken away incorrectly. Finally, we had and extra `-` in our phone style name.

Testing instructions:

- Make sure you can send/receive a message in the mobile device and the text input is always visible when the keyboard is open.
- Try to scroll up and down aggressively to try to break out of the chat panel

Do this in iOS Safari, iOS Chrome and Android.

### DIFF
--- a/examples/react/custom-element/src/App.tsx
+++ b/examples/react/custom-element/src/App.tsx
@@ -19,6 +19,9 @@ const config: PublicConfig = {
   },
   layout: {
     showFrame: false,
+    customProperties: {
+      "messages-max-width": `max(60vw, 672px)`,
+    },
   },
   openChatByDefault: true,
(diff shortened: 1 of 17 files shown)


sangeethababu9223:
I couldn't find any input field issues in chrome in Android. But I noticed some issues in firefox with the visibility of input field. And this is fixed with the PR.

But no visible scrolling issues could be found in chrome in Android.

Issue #777, opened 2025-12-16
[Bug]: float variant lost its gradient border since v 1.3.0


devadula-nandan:
### Browser

Chrome

### Package version

latest

### Description

[v 1..3.0](https://chat.carbondesignsystem.com/version/v1.3.0/demo/index.html?settings=%7B%22framework%22%3A%22react%22%2C%22layout%22%3A%22float%22%2C%22writeableElements%22%3A%22false%22%2C%22direction%22%3A%22default%22%7D&config=%7B%22aiEnabled%22%3Atrue%2C%22messaging%22%3A%7B%7D%2C%22header%22%3A%7B%22isOn%22%3Afalse%7D%2C%22layout%22%3A%7B%22showFrame%22%3Afalse%7D%2C%22launcher%22%3A%7B%22isOn%22%3Atrue%7D%2C%22openChatByDefault%22%3Atrue%7D) missing ai border
[v 1.2.1](https://chat.carbondesignsystem.com/version/v1.2.1/demo/index.html?settings=%7B%22framework%22%3A%22react%22%2C%22layout%22%3A%22float%22%2C%22writeableElements%22%3A%22false%22%2C%22direction%22%3A%22default%22%7D&config=%7B%22aiEnabled%22%3Atrue%2C%22messaging%22%3A%7B%7D%2C%22header%22%3A%7B%7D%2C%22layout%22%3A%7B%22showFrame%22%3Afalse%7D%2C%22launcher%22%3A%7B%22isOn%22%3Atrue%7D%2C%22openChatByDefault%22%3Atrue%7D) ai border is present

| v 1.3.0 | v 1.2.1 |
|----------|----------|
| <img width="424" height="690" alt="Image" src="https://github.com/user-attachments/assets/7626f2c5-cc26-42ba-a86a-b3fdbc74fdba" />  | <img width="424" height="690" alt="Image" src="https://github.com/user-attachments/assets/8e9c0390-f5ba-40f1-9796-45b36f8d376a" />  |

### Reproduction / example

seen in demo

### Steps to reproduce

seen in demo

### Suggested severity

None

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/carbon-ai-chat/issues) for duplicate problems

PR #784, opened 2025-12-17
fix: feedback, reasoning steps, shell border

### LINKED ISSUE #782: Collapsed items in reasoning trace still allow keyboard focus

ethanwinters:
Closes #782 #777 

Three fixes:

1. The border on AI mode disappeared
2. The REACT version of thumbs up/down up wasn't defining the onClick handler correctly and was always coming back as "negative feedbacl
3. Applying aria-hidden or inert to a parent element of slotted content wasn't marking the slots as aria-hidden or inert so all the hidden reasoning step content was getting invisible keyboard focus when collapsed

#### Testing / Reviewing

1. Check if blue border is on chat when in AI mode, and that it is still gray with AI mode off
2. Make sure that hitting "thumbs" up actually selects thumbs up in demo site and both storybooks
3. From demo site, enter response for reasoning steps. When reasoning steps area is closed, you should be able to tab right passed it. When its open, you should be able to tab past all content collapsed in individual steps

### DIFF
--- a/packages/ai-chat-components/src/components/reasoning-steps/src/cds-aichat-reasoning-step.scss
+++ b/packages/ai-chat-components/src/components/reasoning-steps/src/cds-aichat-reasoning-step.scss
@@ -25,6 +25,10 @@
   display: none;
 }
 
+.#{$prefix}__panel--hidden {
+  display: none;
+}
+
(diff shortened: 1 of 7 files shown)


annawen1 (approved):
Tested, and LGTM

PR #792, opened 2025-12-19
fix: overlay panel border radius regressions

devadula-nandan:
Closes #

ever since the custom overlay panels refactor between the versions. there has been regressions on border radius.

https://github.com/user-attachments/assets/7e918e4b-bd3a-45da-86fc-e2d90a261918

this pr fixes border radius on buttons and overlay panels to pixel perfection

after

https://github.com/user-attachments/assets/7bef42e3-5fe0-469c-bf4f-f0afa25fefbe

#### Changelog

**Changed**

- some styles which adds better scope to header in overlay panel with header in chat. 
- scopes css property of border radius to 0 only on header that's in overlay panel.

#### Testing / Reviewing

make sure everything that must have rounded border radius inside overlay panel should have it.


annawen1 (approved):
LGTM!

Issue #731, opened 2025-12-02
Prompt line development


ethanwinters:
The current chat input (`Input.tsx` / `ContentEditableInput.tsx`) is a React component tightly coupled to Redux. Every feature addition — file uploads, structured data, keyboard shortcuts — has deepened this coupling, making the input difficult to extend or test in isolation.

A visual redesign of the input field (the "prompt line") is underway, requiring new components that don't carry this legacy baggage. Building those components as Lit web components aligns with the broader Carbon migration toward `@carbon/ai-chat-components`, where the chat shell, header, and toolbar already live.

At the same time, consumers need inline token extensibility — `@mentions`, `/commands`, and custom chips — in the chat input. The `Selection API` and `Range` constraints around shadow DOM make this a natural forcing function to get the input architecture right before layering on extensibility APIs.

Designs: https://www.figma.com/design/HynmDtJaDMh0IkGdvgWTZP/Carbon-AI---AI-Chat-Dev-Handoff---Phase-2?node-id=723-49372&p=f&m=dev

PR #800, opened 2026-01-06
fix: add accessible name to Input SVGs

tom-youd:
Closes # 

We noticed that there is a violation - https://www.ibm.com/able/requirements/requirements/#1_1_1 (from moving from 0.5.0) to 1.x.x with the send button SVG. This PR addresses that issue

#### Changelog

**New**

- Added `aria-label` and `role` attributes to the send button within the input field

**Changed**

- n/a

**Removed**

- n/a

#### Testing / Reviewing

Using the IBM equal access tool, you should not see the below error
Screenshot 2026-01-06 at 13 48 09

### DIFF
--- a/.github/workflows/create-release-tag-and-pr.yml
+++ b/.github/workflows/create-release-tag-and-pr.yml
@@ -27,12 +27,6 @@ on:
         description: "Create PR?"
         type: boolean
         default: true
-      update-versions-file:
-        # set to true to update versions.js file
-        required: true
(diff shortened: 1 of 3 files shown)


tom-youd:
@ethanwinters not sure if you want me to create a new issue or whether or not it falls under the [input ally issue](https://github.com/carbon-design-system/carbon-ai-chat/issues/731) you have already raised?

PR #804, opened 2026-01-08
feat(workspace): Integrate workspace in ai-chat main

devadula-nandan:
Closes #

workspace feature make available for preview/testing

#### Changelog

**New**

- builds workspace feature

#### Testing / Reviewing

manually make sure nothing breaks.
make sure it meets the expected behaviors documented from the epic and its sub issues.

### DIFF
--- a/packages/ai-chat/src/types/events/eventBusTypes.ts
+++ b/packages/ai-chat/src/types/events/eventBusTypes.ts
@@ -664,6 +664,7 @@ export interface BusEventCustomPanelClose extends BusEvent {
 
 /**
  * @category Events
+ * @experimental
  */
 export interface BusEventWorkspacePreOpen extends BusEvent {
   type: BusEventType.WORKSPACE_PRE_OPEN;
@@ -682,6 +683,7 @@ export interface BusEventWorkspacePreOpen extends BusEvent {
 
 /**
  * @category Events
+ * @experimental
  */
 export interface BusEventWorkspaceOpen extends BusEvent {
   type: BusEventType.WORKSPACE_OPEN;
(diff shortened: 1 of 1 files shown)


packages/ai-chat-components/src/components/workspace-shell/src/workspace-shell.scss
```diff
@@ -21,23 +21,29 @@ $css--plex: true !default;
 :host(#{$prefix}-workspace-shell) {
   @include type.type-style("body-01");
 
-  display: block;
+  display: flex;
+  /* stylelint-disable declaration-no-important */
+  box-sizing: border-box !important;
```
ethanwinters: Hmm... don't love that! why is that needed?

packages/ai-chat/src/types/events/eventBusTypes.ts
```diff
@@ -642,6 +662,78 @@ export interface BusEventCustomPanelClose extends BusEvent {
   type: BusEventType.CUSTOM_PANEL_CLOSE;
 }
 
+/**
+ * @category Events
```
ethanwinters: Lets mark all these experimental just in case!

packages/ai-chat-components/src/components/workspace-shell/src/workspace-shell.scss
devadula-nandan: https://github.com/user-attachments/assets/d3cbf14e-a030-4ff2-89b5-b34b21b1860a

`!important` was not necessary in the Web Components example.
However, when switching it to React, the `box-sizing` styles were supposed to be inherited, but that doesn’t work as expected and instead gets computed to `content-box`, which is the browser default.

I haven’t looked into it much and ended up using `!important`, since it’s only applied on the component host.
Maybe we can avoid it and clean this up if we replace `Modal` with `OverlayPanel` and make sure there are no visual issues without `!important` on box sizing.

Issue #1029, opened 2026-03-03
[Bug]: Disclaimer accept button disabled on first load


tom-youd:
### Browser

Chrome, Safari

### Package version

1.7.0

### React version (if applicable)

18

### Description

When the AI chat component is initially rendered as closed and you have the disclaimer turned on, when the user clicks on the launcher to open the chat the disclaimer accept button is disabled. The only way to work around it is to hard reload the page.
When the chat is opened by default this doesn't seem to be an issue.

I'm fairly confident it's to do with the scroll event (which looks like it was refactored in the 1.7.0 release)

### Reproduction / example

n/a

### Steps to reproduce

- Render the chat with a disclaimer enabled
- Ensure the chat is closed by default 
- Open the chat 
- Disclaimer button is disabled
- Reload the page - disclaimer button is enabled

### Suggested severity

None

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/carbon-ai-chat/issues) for duplicate problems

PR #1048, opened 2026-03-04
fix(bug): disclaimer not getting accepted when initially closed

tom-youd:
Closes https://github.com/carbon-design-system/carbon-ai-chat/issues/1029

Fix the disclaimer accept button enabling when the main window is closed initially and when the disclaimer HTML is a small paragraph.

For some reason (when the disclaimer HTML is small) the `isScrollable` property from the event (when closed) was `true`, but when initially open it's false

#### Changelog

**New**

- n/a

**Changed**

- Adjust the `useEffect` to fire the `onBodyScroll` event when the mainWindow is toggled on.

**Removed**

- n/a

#### Testing / Reviewing

Will post videos of before and after


tom-youd:
@ethanwinters @davidmenendez any chance I can get a review for this and squeeze this into the 1.8.0 release as it's a stop ship for CP4I.

PR #1098, opened 2026-03-18
feat: extend header width

### LINKED ISSUE #1094: dev work for header size changes

ethanwinters:
Closes #1094 

Add configurable header width constraint option

#### Changelog

**New**

- Added `hasContentMaxWidth` configuration option to `HeaderConfig` to control whether the header should be constrained to the messages max width or span full width
- Added header width switcher control in demo app to test the new configuration option
- Added smooth transition animation for header width changes (respects prefers-reduced-motion)

**Changed**

- Header now defaults to full width (`hasContentMaxWidth: false`) instead of being constrained to message width
- Header bottom writeable element now respects the `hasContentMaxWidth` configuration
- Updated header container styling to conditionally apply max-width constraint based on configuration
- Simplified demo layout switcher by removing the "Full w/hasContentMaxWidth: false" option (now controlled via header settings)

**Removed**

- Removed `fullscreen-no-gutter` layout option from demo (functionality replaced by header configuration)
- Removed hardcoded header width constraint behavior

#### Testing / Reviewing

Using the demo site:

1. **Test default behavior (full width header)**
   - Open the demo site
   - Select "Fullscreen" layout from the layout switcher
   - Verify the header spans the full width of the chat container
   - Verify messages remain constrained to their max-width

2. **Test constrained width header**
   - In the demo header settings panel, find "Has content max width" dropdown
   - Change from "Default" (or "False") to "True"
   - Verify the header now constrains to match the message width
   - Verify smooth transition animation when toggling (if motion is not reduced)

3. **Test header bottom writeable element**
   - Turn on writeable elements
   - Toggle "Has content max width" between True and False
   - Verify the writeable element content follows the same width constraint as the header

4. **Test in different layouts**
   - Test the header width configuration in Sidebar and Float layouts
   - Verify the behavior is consistent across all layout modes

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@carbon/ai-chat-examples-demo",
   "private": true,
-  "version": "1.8.0",
+  "version": "1.9.0-rc.0",
   "type": "module",
   "description": "A demo and testing framework using @carbon/ai-chat as a React component.",
   "scripts": {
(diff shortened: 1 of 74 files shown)


davidmenendez (approved):
LGTM

Issue #1102, opened 2026-03-18
[Feature Request]: Pass data attribute thru for top level component/element


nelsonchen90:
### The problem

Carbon autotrack provided a new way to track events based on regions https://pages.github.ibm.com/carbon/ibm-products/developing/instrumentation/autotrack/#autotrack-regions To use it, consumer can define `data-` attributes on html elements. 
However, currently the top level component (i.e `ChatCustomElement` doesn't allow that so consumer will have to use a wrapper element (i.e div) just to have this in the DOM tree.

### The solution

The top level components should allow at least the `data-` attribute to be passed thru on the top level element it will render.

### Is this a blocker to using AI Chat

No

### Business priority

None

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)

PR #1172, opened 2026-04-03
feat: lazy load examples

### LINKED ISSUE #727: Chat shell component examples
Create an example showing someone loading the initial shell in their build and then lazy loading in the rest of the chat as needed.

ethanwinters:
Closes #727 #1102 

Add custom element lazy-load and float-lazy-load examples for React and web-components using chat-shell component as fallback while lazy loading. Added a "custom-element-as-float" example with new exported float classes for folks that are allowing people to swap between modalities, something that is annoying when you have ChatContainer and ChatCustomElement both on the same page... now you can just use a custom element and make it look like float.

Beyond the 6 new examples, the key changes worth reviewing are:

- **`ChatCustomElement` DOM props passthrough** — `...domProps` spread enables consumers to drive float animations via CSS classes applied from outside; this is the core API the new examples rely on (`packages/ai-chat/src/react/ChatCustomElement.tsx`)
- **New public SCSS/CSS globals** — `_chat-float-layout.scss` and `_chat-launcher-layout.scss` define the float and launcher positioning/animation contract; these are now compiled and shipped as `dist/css/chat-float-layout.css` and `dist/css/chat-launcher-layout.css` (`packages/ai-chat/src/globals/scss/`, `packages/ai-chat/tasks/rollup.aichat.js`)
- **`readCarbonChatSession()` utility** — new public function for reading persisted session state from `sessionStorage` (`packages/ai-chat/src/globals/utils/readCarbonChatSession.ts`)
- **`slot-observer` connect fix** — `updateSlotStates(true)` on `connectedCallback` ensures slot state is reported immediately on attach, which is required for lazy-load scenarios (`packages/ai-chat-components/src/components/chat-shell/src/slot-observer.ts`)
- **`ChatCustomElement` ref timing fix** — callback ref replaced with `useRef` + `useLayoutEffect` to correctly gate rendering on element readiness (`packages/ai-chat/src/react/ChatCustomElement.tsx`)

#### Changelog

**New**

- `examples/react/custom-element-lazy-load` — React example that lazy-loads `ChatCustomElement` via `React.lazy()` with a `ChatShell` overlay displayed during bundle loading and initialization
- `examples/react/custom-element-as-float` — React example demonstrating float widget mode with a custom Carbon `ChatButton` launcher and CSS-class-driven open/close phase animations
- `examples/react/custom-element-as-float-lazy-load` — React example combining lazy loading and float mode; defers the `ChatCustomElement` bundle until first launcher click and uses `readCarbonChatSession()` to automatically reopen the chat if it was open in the previous session
- `examples/web-components/custom-element-lazy-load` — Lit/Web Components equivalent of the lazy-load example
- `examples/web-components/custom-element-as-float` — Lit/Web Components equivalent of the float example
- `examples/web-components/custom-element-as-float-lazy-load` — Lit/Web Components equivalent of the float-lazy-load example
- New public globals exported from `@carbon/ai-chat`:
  - **SCSS sources** — `_chat-float-layout.scss`, `_chat-launcher-layout.scss`, `_chat-theme.scss` available at `@carbon/ai-chat/dist/scss/` for consumers who compile their own CSS
  - **Compiled CSS** — `chat-float-layout.css`, `chat-launcher-layout.css` available at `@carbon/ai-chat/dist/css/` for direct import (used by the new float examples)
  - **`readCarbonChatSession(namespace?)`** utility function — reads persisted `viewState` from `sessionStorage`; returns `null` if there is no prior session or if the stored version does not match the current package version
  - **`PublicHistoryPanelState`** type — was previously missing from the public exports

**Changed**

- `ChatCustomElement` now accepts arbitrary DOM props — `...domProps` is spread onto the container `div` and `ChatContainer`, enabling consumers to attach CSS classes, `data-*` attributes, and event handlers for external state management (required for CSS-driven float animations)
- `ChatCustomElement` ref handling (bug fix) — switched from a callback ref to `useRef` + `useLayoutEffect` to ensure element readiness is signaled at the correct point in the React lifecycle
- `slot-observer` now fires on initial connection (bug fix) — `updateSlotStates(true)` is called on `connectedCallback` so consumers receive a slot-state notification immediately on attach; previously this was skipped, which broke lazy-load scenarios where the observer connected after slots were already populated
- Launcher tooltip position in RTL mode (bug fix) — tooltip direction is now computed once and memoized; previously the direction was evaluated inline at render time, which produced the wrong result because `cds-button` elements enforce `direction: ltr` via CSS
- Test helpers simplified in `examples/react/vite` and `examples/react/jest-happydom` — `renderChatContainer()` and `waitForChatElement()` now locate the chat widget by searching shadow DOM for its `data-testid` rather than waiting on specific element upgrades or shadow root presence; added `deepQuerySelector()` helper for recursive shadow DOM traversal; removed implementation-detail assertions in favor of behavior-level checks
- Webpack configs updated across all examples — examples now watch local `@carbon/ai-chat` package changes (`managedPaths: []`, `watchOptions.ignored` pattern) to support monorepo development
- `developer-handbook.md` and example READMEs updated to document the new examples and public CSS/SCSS exports

**Removed**

- `packages/ai-chat/docs/Migration-1.0.0.md`
- `packages/ai-chat/docs/migration-0.5.1.md`

#### Testing / Reviewing

See `examples/react/README.md` and `examples/web-components/README.md` for setup and run instructions for all examples.

**custom-element-lazy-load** (React and web-components)

- Run the example and verify that a loading overlay appears while the bundle loads (there is a simulated delay), then disappears and the chat renders correctly
- Note: these examples have no launcher button — the chat opens automatically on load. If the chat is in a closed state from a prior session the overlay will appear but the chat will not open. Clear `sessionStorage` and reload to reset. This edge case is left unguarded intentionally to keep the example simple

**custom-element-as-float** (React and web-components)

- Verify the launcher button is visible on load
- Click the launcher and confirm the chat opens with an entrance animation
- Close the chat and confirm the exit animation plays and the launcher reappears

**custom-element-as-float-lazy-load** (React and web-components)

- Click the launcher for the first time and verify the loading overlay appears while the bundle downloads, then disappears once the chat is ready
- Open the chat, refresh the page, and confirm the chat reopens automatically (open state is persisted via `readCarbonChatSession()`)
- Close the chat, refresh the page, and confirm the chat stays closed on reload


docs/developer-handbook.md
```diff
@@ -70,67 +64,88 @@ Afterwards, you should be good to go!
...
-of Sass.
+## Directory Structure
 
-Using these `mixins.scss` entrypoints allows you as an application developer to
-control when these side-effects are applied in your project.
+```
+carbon-ai-chat/
+├── packages/
+│   ├── ai-chat/              # Core React chat package (@carbon/ai-chat)
+│   └── ai-chat-components/   # Lit web components package (@carbon/ai-chat-components)
```
warrenmblood: These package descriptions could be improved. This make it seem like `ai-chat-components` is the WC version of `ai-chat`

examples/react/basic/webpack.config.js
```diff
@@ -90,6 +90,12 @@ export default () => {
     },
     plugins: createPlugins(shouldAnalyze),
     devtool: "source-map",
+    snapshot: {
+      managedPaths: [], // don't treat node_modules as immutable
+    },
+    watchOptions: {
+      ignored: /node_modules\/(?!@carbon\/ai-chat)/, // watch only our packages
```
warrenmblood: Would be a good idea to make it clear that we want to track both @carbon/ai-chat and @carbon/ai-chat-components as it's not totally clear from that regex and comment

packages/ai-chat/docs/Migration-1.0.0.md
warrenmblood, line 1: Let's remove this file in a separate PR after we check the number of users still on an older version than 1.0.0

packages/ai-chat/src/chat/components-legacy/responseTypes/util/ClickableImage.scss
warrenmblood, line 1: There's a regression where the icon in the lower right corner of the clickable image isn't visible:

New:

Image

Exisiting:
Image

packages/ai-chat/src/globals/scss/_chat-launcher-layout.scss
warrenmblood, line 1: There's a styling issue being introduced where the launcher button's width is compressed. I'm seeing this across the demo and examples:

Image

warrenmblood (changes requested):

warrenmblood (approved):
LGTM!

PR #1213, opened 2026-04-10
feat: better memoization

ethanwinters:
Reduce unnecessary re-renders across `@carbon/ai-chat` by replacing whole-state Redux selectors with granular per-slice selectors and adding targeted `React.memo` wrappers to key components. Previously, any Redux dispatch (even to an unrelated state slice) would re-render `AppShell` and `MessagesComponent` along with their entire subtrees. The demo app's 2-second `setInterval` made this especially visible: the whole component tree was re-rendering every tick.

#### Changelog

**Changed**

- `AppShell` now uses 15 individual `useSelector` calls (one per state slice) instead of `useSelector((state) => state)`, so it only re-renders when a slice it actually reads changes. Also wrapped in `React.memo` to prevent re-renders caused by parent (`ChatAppEntry`) prop changes that `AppShell` doesn't consume (e.g. `renderUserDefinedResponse`).
- `MessagesComponent`'s state injector now selects only the 5 `AppState` fields the class reads (`config`, `allMessagesByID`, `humanAgentState`, `persistedToBrowserStorage`, `assistantInputState`) with `shallowEqual` comparison, instead of spreading the entire state as props.
- `useInputCallbacks` no longer depends on `appState`. `onSendInput` reads fresh state from the store at call time (which is also more correct — avoids a stale closure). `showUploadButtonDisabled` takes `humanAgentFileUploadInProgress` as a direct boolean prop.
- `AppShellPanels` and `AppShellWriteableElements` wrapped in `React.memo` to prevent re-renders when parent re-renders but their props are unchanged.
- `MessageTypeComponent`'s inline `renderMessageComponent` callback (duplicated 3x in JSX) extracted into a single `useCallback` with empty deps, making the existing `React.memo` on `BodyMessageComponents`, `CardItemComponent`, and `GridItemComponent` actually effective.
- Demo: `renderCustomMessageFooter` wrapped in `useCallback`; `setInterval` cleanup added; `stateText` added to `allWriteableElements` dependency array so writeable elements properly receive parent state updates; extracted `historyIsMobile` variable to satisfy lint rules.

#### Testing / Reviewing

All verification should be done via the demo site (both React and Web Component modes). Use React DevTools Profiler with "Highlight updates when components render" enabled to confirm re-render boundaries.

- [ ] **Writeable elements update with parent state**: With writeable elements enabled in demo settings, confirm the green writeable element boxes display a changing timestamp value (the `parentStateText` prop from the 2-second interval). This verifies the `allWriteableElements` dependency fix.
- [ ] **User-defined responses update with parent state**: Send a message that triggers a `user_defined` response type (e.g. the "green" type). Confirm the response displays an updating `parentStateText` timestamp and its own local timestamp.
- [ ] Similarly, double check the basic react and web component examples user_defined responses are fine
- [ ] **No cascading re-renders on interval tick**: With React DevTools Profiler recording, observe that the 2-second interval does NOT cause `AppShell`, `MessagesComponent`, `Header`, `Input`, or `AppShellPanels` to highlight/re-render. Only the portal containers for user-defined responses and writeable elements should update.
- [ ] **Messages render correctly**: Send text messages, receive responses with various types (text, buttons, cards, carousels, images). Verify all render correctly and streaming works.
- [ ] **Panels work**: Open/close the workspace panel, history panel, iframe panel, and custom panel. Confirm animations and content render correctly.
- [ ] **Home screen**: Toggle between the home screen and conversation view. Confirm starters work and the transition is smooth.
- [ ] **Custom footer slots**: Trigger a response with a custom footer. Confirm the footer content renders in the correct slot.
- [ ] **Web component demo**: Switch to the web component demo tab. Confirm writeable elements, user-defined responses, and custom footers all function identically to the React demo.

### DIFF
--- a/demo/src/react/DemoApp.tsx
+++ b/demo/src/react/DemoApp.tsx
@@ -64,7 +64,8 @@ function DemoApp({ config, settings, onChatInstanceReady }: AppProps) {
   const isSidebarLayout = settings.layout === "sidebar";
 
   useEffect(() => {
-    setInterval(() => setStateText(Date.now().toString()), 2000);
+    const id = setInterval(() => setStateText(Date.now().toString()), 2000);
+    return () => clearInterval(id);
   }, []);
   /**
    * Handler for user_defined response types. You can just have a switch statement here and return the right component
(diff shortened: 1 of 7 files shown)


davidmenendez (approved):
LGTM

PR #1242, opened 2026-04-15
fix: sometimes selector calls didn't work

ethanwinters:
Replaces the custom `useSelector` implementation with `useSyncExternalStoreWithSelector` from `use-sync-external-store/with-selector` — the same utility `react-redux` uses internally. This fixes a concurrent-rendering bug reported by a consumer where `showLauncher` remained truthy after the chat was closed, causing the launcher to stay rendered.

The previous implementation had two concurrent-safety issues:

1. `getSnapshot` was not pure — it mutated `lastSelectedRef.current` during render. React may call `getSnapshot` multiple times per render for tearing detection, and mutation during that call could freeze the cached reference on a stale value.
2. `getSnapshot` identity changed across renders (via `useCallback` with a `selector` dep), which can cause `useSyncExternalStore` to treat each render's snapshot as independent, opening the door to stale reads under transitions / StrictMode.

Delegating to `useSyncExternalStoreWithSelector` eliminates the class of bug entirely while preserving the public `useSelector` signature, so no call sites need to change.

#### Changelog

**Changed**

- `useSelector` hook now delegates to `useSyncExternalStoreWithSelector` from `use-sync-external-store/with-selector`, making store reads safe under concurrent rendering.

**Removed**

- Custom `computeSelected` / `lastSelectedRef` caching logic inside `useSelector` (no longer needed — the upstream utility handles equality-based reference stabilization).
- Unused `UNINITIALIZED` sentinel.

#### Testing / Reviewing

We are going to have to publish a new rc and have the consumer test

### DIFF
--- a/packages/ai-chat/src/chat/hooks/useSelector.tsx
+++ b/packages/ai-chat/src/chat/hooks/useSelector.tsx
@@ -8,23 +8,16 @@
  */
 
 /**
- * Basic selector using `useSyncExternalStore`.
- * Pure snapshot returns previous ref when values are equal. Caches last value...comparator defaults to `Object.is`.
+ * Basic selector backed by `useSyncExternalStoreWithSelector` — the same
+ * utility `react-redux` uses internally. This delegates selector identity,
+ * equality-check, and snapshot-purity handling.
  *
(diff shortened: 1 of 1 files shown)


warrenmblood (approved):
LGTM!

PR #1249, opened 2026-04-16
fix(history): overflow menu visible for active chat

### LINKED ISSUE #1158: [Bug]: Chat history - mobile overflow menu on chat items can be difficult to open on tap
### Browser

_No response_

### Package version

v1.10.0

_No response_

### Description

When opening the chat history on mobile, there is an overflow menu for each chat history item (pin, rename, delete). However the overflow menu doesn't appear until the history item is hovered on or focused on. This may be be difficult to do on mobile as there's not hover

Solution would be to expose the overflow menu on mobile.

After, speaking with the accessibility team, suggestion is to expose the overflow menu for the selected / active chat item in both mobile and desktop views. This way user knows the overflow menu exists before they hover / focus on chat items.

### Reproduction / example

N/A

### Steps to reproduce

N/A

### Suggested severity

None

### Application / product

_No response_

### Code of Conduct

annawen1:
Closes #1158 

This PR makes the overflow menu which currently is not visible until user hovers or focuses on the chat item, visible for the active chat item. This lets the user know that there is an overflow menu available. 

Screenshot 2026-04-16 at 10 56 22 AM

#### Changelog

**New**

- add style for selected chat item

#### Testing / Reviewing

check demo deploy preview and enable chat history. Confirm that the overflow menu is visible for the active chat item.

### DIFF
--- a/packages/ai-chat-components/src/components/chat-history/src/chat-history.scss
+++ b/packages/ai-chat-components/src/components/chat-history/src/chat-history.scss
@@ -279,6 +279,10 @@
   visibility: hidden;
 }
 
+:host(#{$prefix}-history-panel-item[selected]) #{$cds-prefix}-overflow-menu {
+  visibility: visible;
+}
+
 :host(#{$prefix}-history-panel-item:hover) #{$cds-prefix}-overflow-menu,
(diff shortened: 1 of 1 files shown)


warrenmblood (approved):
LGTM!

PR #1269, opened 2026-04-20
fix: set focus flow when deleting / unpinning / pinning history chat item

### LINKED ISSUE #1187: [A11y]: Chat history - focus when chat item is deleted / pinned / unpinned
### Package version

latest

### Description

When deleting, pinning, or unpinning a chat item in chat history, the focus gets lost and user has to tab through from the top all over again. 

Instead it should be:
After delete → Move focus to the next available chat item (or previous if it was the last item)
After pin/unpin → Maintain focus on the same item (updated state)

### Reproduction / example

### Steps to reproduce

N/A

### Suggested severity

None

### Code of conduct

annawen1:
Closes #1187

Currently when performing the delete / pin / unpin actions on the chat items, the focus disappears and user has to tab through all the chat history components again from the top. This PR sets the focus states for those actions.

Expected focus behaviors:
After delete → Move focus to the next available chat item (or previous if it was the last item)
After pin/unpin → Maintain focus on the same item (updated state)

https://github.com/user-attachments/assets/28409e51-d4f9-4e9a-9056-c2c87216d4d4

For the delete action, `_getFocusDetailForDeletedItem()` is added to determine the next focusable item and is added to the event details when the "history-delete-confirm" custom event is dispatched. Then in `history-shell` we have event listeners for the "history-delete-confirm" custom event which then handles the focus management to the next item. Also handles the case where the active / selected chat is deleted, it will move the active state to the next chat item that is focused.

For the pin / unpin actions, we provide a utility `focusElementAfterRepaint()` for users to implement on their end. Since these actions are customizable, I didn't want to build in that logic into our chat history components.

```
_handlePinToTop = () => {
    .......
    
    focusElementAfterRepaint(
        this.renderRoot,
        `cds-aichat-history-panel-item#${CSS.escape(itemId)}`,
      );
}
```

#### Changelog

**New**

- `focusElementAfterRepaint` focus utility
- Added documentation for the delete panel and focus utility usage in the storybook `*.mdx` files
- chat-history.test.ts file

**Changed**

- Added the delete and pin / unpin focus changes to the examples

#### Testing / Reviewing

Check the demo deploy preview and storybooks (both React and Web Components):

- turn on chat history
- tab to options for a chat item and delete item - confirm that focus goes to the next chat item
- tab to options for the last chat item and delete item - confirm that focus goes to the previous chat item
- tab to options for a chat item and select pin / unpin - confirm that focus stays with the same chat item

### DIFF
--- a/demo/src/react/HistoryWriteableElementExample.tsx
+++ b/demo/src/react/HistoryWriteableElementExample.tsx
@@ -29,9 +29,10 @@ import {
   resultItemSection,
 } from "../fixtures/history/chatHistoryData";
 import { customLoadHistory } from "../fixtures/history/customLoadHistory";
+import { focusElementAfterRepaint } from "@carbon/ai-chat-components/es/globals/utils/focus-utils";
 
 import { PinFilled, Search, Time } from "@carbon/icons-react";
-import React, { useState, useCallback } from "react";
+import React, { useState, useCallback, useRef } from "react";
 import { ChatInstance, PanelType } from "@carbon/ai-chat";
 
 interface HistoryExampleProps {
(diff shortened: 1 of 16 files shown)


davidmenendez (approved):
doesn't look like this was an easy bug to fix. looks great and working as intended!

PR #1322, opened 2026-05-01
fix: announce chat history updates

### LINKED ISSUE #1321: [Chat history]: search results not being announced when changed
Image

When the search results label is updated it's not being read by screen readers to let the user know that the results have updated. The element that contains that text just needs to have `aria-live` added to it.

davidmenendez:
Closes #1321 

adds `aria-live` to the search results slot in the chat history component.

#### Changelog

**New**

- adds `aria-live` to the search results slot in the chat history component

#### Testing / Reviewing

1. enable VO or your screenreader of choice
2. open the chat history component in storybook or in the demo app
3. enter in a search
4. verify that results are being announced

### DIFF
--- a/packages/ai-chat-components/src/components/chat-history/src/history-content.ts
+++ b/packages/ai-chat-components/src/components/chat-history/src/history-content.ts
@@ -44,17 +44,19 @@ class CDSAIChatHistoryContent extends LitElement {
 
   render() {
     return html`
-      ${this._hasResultsCount
-        ? html`<span class="${prefix}--history-content__results-count">
-            <slot
+      <div class="${prefix}--history-content__container" aria-live="polite">
+        ${this._hasResultsCount
+          ? html`<span class="${prefix}--history-content__results-count">
+              <slot
+                name="results-count"
+                @slotchange=${this._handleResultsCountSlotChange}
+              ></slot>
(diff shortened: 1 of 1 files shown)


packages/ai-chat-components/src/components/chat-history/src/history-content.ts
```diff
@@ -44,17 +44,19 @@ class CDSAIChatHistoryContent extends LitElement {
...
               @slotchange=${this._handleResultsCountSlotChange}
-            ></slot>
-          </span>`
-        : html`<slot
-            name="results-count"
-            @slotchange=${this._handleResultsCountSlotChange}
-          ></slot>`}
+            ></slot>`}
+      </div>
       <slot></slot>
```
annawen1, line 60: Should the wrapper div should wrap around the `<slot></slot>` here as well since the search results appear in the slot here?

davidmenendez: i tried that initially, but i decided against it because if we do it it starts to read **all** of the results and it gets cluttered very quickly 😂 

i think it's best to just stick with the label since it's direct and concisely announces that the dom has changed

annawen1 (approved):

PR #1330, opened 2026-05-04
chore(chat-history): use attributes for results count and label instead of slot

### LINKED ISSUE #1321: [Chat history]: search results not being announced when changed

annawen1:
Closes #1321

Currently for the chat history component, when typing into the search input field, the screen reader will read just the results count number (ie. 0) instead of "Results: 0". This can be confusing to the user. Instead of having the results label and number come through the slot, have these as attributes in the `history-content` component instead - the screen reader will then pick up on the 'Results' as well as the results count number.

#### Changelog

**Changed**

- in `history-content.ts`, set up two new properties (`results-label` and `results-count`)
- update demo and examples to reflect the property change instead of passing the information through the slots.

#### Testing / Reviewing

- Go to demo deploy preview and enable chat history
- enable screen reader
- input string in search field
- screen reader should read out "results" with number

### DIFF
--- a/demo/src/react/HistoryWriteableElementExample.tsx
+++ b/demo/src/react/HistoryWriteableElementExample.tsx
@@ -378,10 +378,14 @@ function HistoryWriteableElementExample({
         onNewChatClick={handleNewChat}
         onSearchInput={handleSearchInput}
       />
-      <HistoryContent>
-        {(showSearchResults || noSearchResults) && (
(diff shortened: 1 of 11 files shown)


davidmenendez (approved):
awesome!

PR #1335, opened 2026-05-05
fix: checkbox accessibility violations

### LINKED ISSUE #1334: [A11y]: Checkbox label violation
Placeholder for an issue that needs to be addressed in the core web component checkbox

Image

This violation shows up in numerous places in our end, so it needs to be addressed

Image

**Update**

After some brief, initial investigation: it looks like the issue stems from the usage of `slot` with the checkbox component. using `label-text` instead of the default slot resolves this issue. I'm not sure if this is a general issue with slots or is a false positive from the accessibility checker, but for the time being i'll update our usage to switch from the slot to the prop to alleviate the violations.

davidmenendez:
Closes #1334 

alleviates checkbox accessibility violations

#### Changelog

**New**

- Added a story for the markdown component that demonstrates it's usage with carbon checkboxes

**Changed**

- switched from using the body slot in `cds-checkbox` to using the `label-text` prop

#### Testing / Reviewing

`npm run test` and accessibility checker

### DIFF
--- a/demo/src/framework/demo-chat-feedback-switcher.ts
+++ b/demo/src/framework/demo-chat-feedback-switcher.ts
@@ -1,5 +1,5 @@
 /*
- *  Copyright IBM Corp. 2025
+ *  Copyright IBM Corp. 2025, 2026
  *
  *  This source code is licensed under the Apache-2.0 license found in the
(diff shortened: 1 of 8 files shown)


annawen1 (approved):
LGTM!

PR #1336, opened 2026-05-05
fix: adjust auto-scroll to show previous messages

### LINKED ISSUE #1290: [Feature Request]: New Message Auto Scroll - Change position slightly
### The problem

Currently when you enter a new message it scrolls all the way to the top and in some cases might appear that its a new chat.  This could confuse the user that its a new session.

### The solution

Maybe dont scroll ALL the way to the top but rather display partial of the previous message.
Also make the scroll more obvious because right now i think its just an instant move up and not really a scroll.

Image

### Examples

_No response_

### Application / product

_No response_

None

### Business priority

None

_No response_

### Additional contacts

_No response_

### Available extra resources

_No response_

### Code of Conduct

kavyanekkalapu:
Closes #1290

#### Changelog

**Changed**

- Modified `AUTO_SCROLL_EXTRA` constant from `20` to `-60` in `messagesAutoScrollController.ts`
- This keeps ~60px of previous conversation visible when scrolling to new messages

#### Testing / Reviewing

- Manually tested by sending messages and verifying scroll behavior
- Previous messages remain partially visible above new messages

#### After screenshot
Screenshot 2026-05-05 at 4 10 32 PM

#### Before screenshot
Screenshot 2026-05-05 at 4 42 49 PM

### DIFF
--- a/packages/ai-chat/src/chat/utils/messagesAutoScrollController.ts
+++ b/packages/ai-chat/src/chat/utils/messagesAutoScrollController.ts
@@ -1,5 +1,5 @@
 /*
- *  Copyright IBM Corp. 2025
+ *  Copyright IBM Corp. 2025, 2026
  *
  *  This source code is licensed under the Apache-2.0 license found in the
  *  LICENSE file in the root directory of this source tree.
@@ -72,7 +72,7 @@ function applySpacerDeficit(spacerElem: HTMLElement, deficit: number): void {
  * Note: A negative value will leave more of the previous message visible above the target message.
  * For example, -60 will show approximately 60px of the previous conversation above the new message.
  */
-const AUTO_SCROLL_EXTRA = -60; 
+const AUTO_SCROLL_EXTRA = -60;
 
 /**
  * The visible portion (in pixels) to show at the bottom of a tall message when auto-scrolling.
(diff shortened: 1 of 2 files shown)


kavyanekkalapu:
I have read the DCO document and I hereby sign the DCO.

kavyanekkalapu:
recheck

annawen1:
@kavyanekkalapu if you run `npm run format:write` at the root of the project, that should fix the format check

kavyanekkalapu:
> @kavyanekkalapu if you run `npm run format:write` at the root of the project, that should fix the format check

Fixed it, Thank you!

davidmenendez (approved):
LGTM

@annawen1 i know you worked on this, so i'll defer to you if this doesn't seem correct visually or isn't to spec 👍

annawen1:
I was thinking we would also need to make adjustments for the use case where the request message is too long. We have [logic where if the message height is larger than 1/4 of the chat height](https://github.com/carbon-design-system/carbon-ai-chat/blob/main/packages/ai-chat/src/chat/utils/messagesAutoScrollController.ts#L246-L270), we adjust the scroll position to show the bottom portion of the request message. 

However it looks like there was a regression since I don't see that happening in our current demo site, so we can merge this in and [open another issue](https://github.com/carbon-design-system/carbon-ai-chat/issues/1359) for the "request message is too long" scenario! 

Thank you @kavyanekkalapu for contributing!

annawen1 (approved):

PR #1340, opened 2026-05-06
feat(chat-history):  add `show-actions` option  to display actions for all chat items

### LINKED ISSUE #1339: [chat history]: add option to always show actions for all history panel items
Currently, the overflow menu or actions only appear on render for the selected / active chat item. To access the  overflow menus for other chat items, user will have to focus or hover on the chat items. 

Image

There are potential cases where there is no active chat and on first render, especially on mobile it may be difficult for users to discover the overflow menu.

Solution is to add a `show-actions` attribute to `history-panel` that will make all the actions visible for all history panel items.

annawen1:
Closes #1339

This PR adds a new `showActions` property to the chat history panel components that allows actions menus to be persistently visible on history items without requiring hover or selection.

Also fixes an issue with the recently merged [chore(chat-history): use attributes for results count and label instead of slot](https://github.com/carbon-design-system/carbon-ai-chat/pull/1330) PR where the "Results: 0" shows in Web Components even when user has not input anything in the search field.

#### Changelog

**New**

- Added `showActions` property to `HistoryPanel` component to control visibility of action menus across all child items
- Added `showActions` property to `HistoryPanelItem` component to control visibility of action menu on individual items
- Added MutationObserver in `HistoryPanelItem` to automatically inherit `showActions` from parent panel
- Added CSS rule to make overflow menu visible when `show-actions` attribute is present
- Added `showActions` control to Storybook stories for both React and web components

**Changed**

- Updated `resultsCount` property type in `HistoryContent` from `Number` to `String | Number` to handle empty string values
- Improved `resultsCount` display logic to properly handle undefined, null, and empty string values

#### Testing / Reviewing

- [ ] Verify that the `showActions` toggle in Storybook controls the visibility of action menus on history items
- [ ] Verify that action menus still appear on hover/selection when `showActions` is disabled (default behavior)
- [ ] Verify that the results count displays correctly (only visible when user types in search input)

### DIFF
--- a/packages/ai-chat-components/src/components/chat-history/__stories__/chat-history-react.stories.jsx
+++ b/packages/ai-chat-components/src/components/chat-history/__stories__/chat-history-react.stories.jsx
@@ -77,11 +77,16 @@ export const Default = {
       control: "boolean",
       description: "renders close chat history action in header.",
     },
+    showActions: {
+      control: "boolean",
+      description: "Show actions on all history panel items.",
+    },
   },
   args: {
     HeaderTitle: "Chats",
     searchOff: false,
     showCloseAction: true,
+    showActions: false,
   },
   render: (args) => {
     const historyShellRef = useRef(null);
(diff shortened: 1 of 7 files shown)


warrenmblood (approved):
LGTM!

PR #1362, opened 2026-05-08
fix(markdown): merge the split tokens for markdown table cells

### LINKED ISSUE #1361: [Bug]: markdown table renders anchor HTML element incorrectly
### Browser

_No response_

### Package version

v1.12.0

_No response_

### Description

When passing in an anchor element into the table markdown

ie.
```
| Lorem        | Ipsum      | Odor    | Amet      |
|--------------|------------|---------|-----------|
| <a href="www.google.com">Carbon<a> | adipiscing | elit    | Venenatis |
| 0            | 1          | 2       | 3         |
| bibendum     | enim       | blandit | quis      |
```

the markdown component doesn't render the `a` element correctly.

Image

Image

The element is parsed incorrectly shown in the above screenshot

### Reproduction / example

https://stackblitz.com/edit/github-fjgf8ab3?file=src%2FcustomSendMessage.ts

### Steps to reproduce

Go to stackblitz: https://stackblitz.com/edit/github-fjgf8ab3?file=src%2FcustomSendMessage.ts

and see that the anchor element does not render with the default link styles.

### Suggested severity

Severity 4 = Unrelated to a user task, has a workaround or does not need a workaround.

### Application / product

_No response_

### Code of Conduct

annawen1:
Closes #1361

Currently when using the anchor element inside a markdown table, it doesn't render the link correctly as the raw inline HTML gets split. The link text gets render outside of the anchor tags (ie. `<a href="www.ibm.com"></a>Carbon`)

Image

In `markdown-render.ts`, use `combineConsecutiveHtmlInline` which is used for regular markdown text already for the `renderCellTokens()` to merge the tokens together when rendering the markdown table cells so the link text stays inside the anchor tags. 

AFTER:
Screenshot 2026-05-08 at 4 35 48 PM

#### Changelog

**New**

- add test case for markdown table cell rendering anchor element

**Changed**

- use `combineConsecutiveHtmlInline` in `renderCellTokens()`
- updated markdown table in demo app `constants.ts` file to render anchor element

#### Testing / Reviewing

type `table` in the demo deploy preview, see that the second row of the table (Silicon) renders correctly as an anchor link

### DIFF
--- a/demo/src/customSendMessage/constants.ts
+++ b/demo/src/customSendMessage/constants.ts
@@ -1,5 +1,5 @@
 /*
- *  Copyright IBM Corp. 2025
+ *  Copyright IBM Corp. 2025, 2026
  *
  *  This source code is licensed under the Apache-2.0 license found in the
  *  LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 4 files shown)


davidmenendez (approved):
LGTM!

PR #1363, opened 2026-05-08
fix: minor visual bugs on cards stories and toolbar

devadula-nandan:
Closes N/A

Story updates

Fixes AILabel not rendering in react stories, previously WC ai-label is re-exported with Lit-React wrapper as AILabel. which used to accept `slot="decorator"`. now that the re-exports are removed. the React AILabel is wrapped in div with `slot="decorator"`

Fixes duplicate classname adding additional border in preview card stories

This PR also fixes toolbar styles for rounded border radius on corners. which should also be fixed in the app.

#### Changelog

**Changed**

- added data-rounded="top-right" to the right element of toolbar
- story fixes after all the re-exports of lit-react wrappers are removed

#### Testing / Reviewing

all card stories and toolbar stories should be looking good, with all operational controls.
the toolbar component border radius changes must be reflecting in the app.

### DIFF
--- a/packages/ai-chat-components/src/components/card/__stories__/preview-card-react.stories.jsx
+++ b/packages/ai-chat-components/src/components/card/__stories__/preview-card-react.stories.jsx
@@ -1,3 +1,12 @@
+/**
+ * @license
+ *
+ * Copyright IBM Corp. 2025, 2026
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
 /* eslint-disable */
 import React, { useState, useEffect } from "react";
 import { ICON_INDICATOR_KIND } from "@carbon/web-components/es/components/icon-indicator/defs.js";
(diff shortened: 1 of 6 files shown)


packages/ai-chat-components/src/components/card/__stories__/preview-card-react.stories.jsx
```diff
@@ -67,14 +79,16 @@ export const Small = {
...
-          slot="decorator"
-        >
-          {aiContent}
-        </AILabel>
+        <div slot="decorator">
+          <AILabel
+            size="mini"
+            autoalign
+            alignment="bottom-right"
+            slot="decorator"
```
annawen1: now that we're wrapping the `AILabel` in the div, do we still need to assign the slot attribute to `AILabel` here?

annawen1 (commented):
LGTM! I think we just need to remove that missed instance of the slot attribute for the `AILabel`

devadula-nandan:
> LGTM! I think we just need to remove that missed instance of the slot attribute for the `AILabel`

done 👍🏻

annawen1 (approved):

PR #1377, opened 2026-05-13
chore: add CSP playwright tests and remove CSP from demo

ethanwinters:
Drops the meta-tag CSP from the demo and replaces it with a Playwright-only strict CSP that the suite enforces at test time. `ibm-common.js` (and the Tealium / TrustArc cascade it triggers) needs a CSP allowlist broad enough that maintaining it inline against a moving third-party target wasn't worth it. To keep CSP regressions in our own code from slipping by, Playwright now route-rewrites the demo's document response to inject a strict policy, registers a `securitypolicyviolation` listener before the first byte renders, and the smoke / disclaimer / homescreen / overflow-menu specs assert no violations were recorded. This forced several already-known fixes (inline-style removals, stylesheet-adoption ordering, idempotent stylesheet adoption) and a couple of unrelated cleanups noted below.

Particularly complex changes worth a careful read:

- [`demo/tests/utils.ts`](demo/tests/utils.ts) — adds `installTestCsp` / `installCspGuard` / `expectNoCspViolations` / `getCspViolations`. `installTestCsp` uses `page.route("**/*", …)` and must `route.fallback()` for non-document requests so the existing `ibm-common.js` abort still fires; the `<head\b[^>]*>` regex is what makes the meta-tag injection survive attribute additions on `<head>`; the `cspInstalledPages` WeakSet prevents duplicate route handlers when both `beforeEach` and `prepareDemoPage` install the CSP.
- [`packages/ai-chat/src/chat/hooks/useStyleInjection.tsx`](packages/ai-chat/src/chat/hooks/useStyleInjection.tsx) — now preserves any other sheets already adopted on the root (e.g. the shared dynamic-css-var-sheet used by avatars / iframes / grid cells). Without this, the second run of the effect drops those sheets and components fall back to a `<style>` element that strict CSP blocks.
- [`packages/ai-chat-components/src/components/shared/dynamic-css-var-sheet.ts`](packages/ai-chat-components/src/components/shared/dynamic-css-var-sheet.ts) — `adoptOnRoot` now short-circuits if the root is already adopted, before calling `ensureSheet()`. Previously the call ordering could spuriously create a fallback `<style>` element on a root that already had the adopted `CSSStyleSheet`, which then trips style-src `'self'`.
- [`packages/ai-chat/docs/typedoc/demoteValidationWarningsPlugin.js`](packages/ai-chat/docs/typedoc/demoteValidationWarningsPlugin.js) — paired with `treatValidationWarningsAsErrors: true` in `typedoc.json`. The plugin only demotes warnings emitted during the **conversion** phase, not the **validate** phase; the docstring explains the constraint so future patterns added here don't silently fail to demote.

#### Major changes

**New**

- Test-time strict CSP and violation guard for the Playwright suite ([`demo/tests/utils.ts`](demo/tests/utils.ts)): `installTestCsp`, `installCspGuard`, `getCspViolations`, `expectNoCspViolations`, and a `CspViolation` shape.
- `expectNoCspViolations(page)` assertion wired into the `afterEach` of `smoke.spec.ts`, `disclaimer.spec.ts`, `homescreen.spec.ts`, and `overflow-menu-arrow-navigation.spec.ts`. Runs before `destroyChatSession` so the assertion fires against the test's own interactions, not the cleanup tear-down.

**Changed**

- Demo runs without a production CSP. The meta tag in [`demo/index.html`](demo/index.html) and the `Content-Security-Policy` header on the webpack dev-server in [`demo/webpack.config.js`](demo/webpack.config.js) are both removed.
- Inline `style="…"` removed from demo content where it tripped style-src under the test CSP:
  - [`demo/src/customSendMessage/doCarousel.ts`](demo/src/customSendMessage/doCarousel.ts) — inline color + star-rating styles
  - [`demo/src/framework/demo-chat-instance-switcher.ts`](demo/src/framework/demo-chat-instance-switcher.ts) — deprecated-API warnings now use `<strong>` instead of inline red
  - [`demo/src/framework/demo-header.ts`](demo/src/framework/demo-header.ts) — visually-hidden announcement uses a class instead of inline absolute positioning
  - [`demo/src/framework/demo-homescreen-switcher.ts`](demo/src/framework/demo-homescreen-switcher.ts) — disclaimer `<p>` elements rely on UA margins instead of inline `padding-bottom`
- [`packages/ai-chat/src/chat/components-legacy/input/ContentEditableInput.tsx`](packages/ai-chat/src/chat/components-legacy/input/ContentEditableInput.tsx) — auto-resize stops setting `editor.style.overflowY` directly and routes through `applyDynamicStyles` / `clearDynamicStyles` from `cspStyleUtils`, with a cleanup function returned from the layout effect.
- [`packages/ai-chat/src/chat/hooks/useStyleInjection.tsx`](packages/ai-chat/src/chat/hooks/useStyleInjection.tsx) — preserves other adopted stylesheets on the root when (re-)adopting the application + CSS-variable sheets, so re-running the effect doesn't drop the shared dynamic-css-var-sheet.
- [`packages/ai-chat-components/src/components/shared/dynamic-css-var-sheet.ts`](packages/ai-chat-components/src/components/shared/dynamic-css-var-sheet.ts) — `adoptOnRoot` is now idempotent at the entry, before `ensureSheet()`, so a second call on the same root can't create a fallback `<style>` element on top of an already-adopted `CSSStyleSheet`.

#### Minor changes

**New**

- `packages/ai-chat/docs/typedoc/demoteValidationWarningsPlugin.js` and `treatValidationWarningsAsErrors: true` in [`packages/ai-chat/typedoc.json`](packages/ai-chat/typedoc.json). Validation warnings (e.g. referenced-but-unexported types) now fail the docs build, with a small allowlist for known-acceptable messages.
- `SystemMessageItem` is now exported from both [`packages/ai-chat/src/aiChatEntry.tsx`](packages/ai-chat/src/aiChatEntry.tsx) and [`packages/ai-chat/src/serverEntry.ts`](packages/ai-chat/src/serverEntry.ts) (previously only `SystemMessageVariant` was exported, which made the referenced item type unresolvable in docs).

**Changed**

- `retries` in [`demo/playwright.config.ts`](demo/playwright.config.ts) is now `process.env.CI ? 1 : 0` — CI gets one auto-retry to absorb timing races; local devs see flakes immediately.

#### Testing / Reviewing

Visual smoke in a browser to confirm the inline-style removals didn't visibly regress anything:
   ```bash
   npm run aiChat:start
   ```
   - Open the demo at `http://localhost:3001/`.
   - In the sidebar, switch to **Carousel** under custom send-message responses — confirm the carousel cards render with their star ratings and descriptions readable (no inline colors anymore, but legibility should be unchanged).
   - Switch the **homescreen** dropdown to `disclaimer` and `disclaimer-with-default` — confirm the disclaimer paragraphs still have visible vertical spacing between blocks (UA `<p>` margins now do the work).
   - In the **Chat instance switcher**, open any of the four "deprecated panel" options — confirm the "⚠️ This API is deprecated" warning is still visually distinguishable from body copy (now bold rather than red).
   - Open the chat, send a multi-line message that crosses the auto-resize threshold — confirm the textarea switches to scrolling instead of growing unbounded.


annawen1:
1. With the `disclaimer-with-default` option I don't see the bottom padding for the paragraphs coming through 
Screenshot 2026-05-13 at 11 54 02 AM

2. The textarea is grows to a certain fixed height, but scroll doesn't appear for me
Screenshot 2026-05-13 at 11 59 43 AM

annawen1 (approved):

Issue #1380, opened 2026-05-13
Add onViewChange / onViewPreChange to ChatContainer and cds-aichat-container


ethanwinters:
## Summary

The `onViewChange` and `onViewPreChange` props exist on the **custom-element** variants (`ChatCustomElement` / `cds-aichat-custom-element`) but are missing from the **plain container** variants (`ChatContainer` / `cds-aichat-container`).

This is asymmetrical with no design reason — the custom-element versions need them to drive the wrapping element's visibility default, but container consumers also have legitimate reasons to react to view changes.

## Why it matters

Today, a `ChatContainer` / `cds-aichat-container` consumer who wants to react to open/close has to:

```tsx
<ChatContainer
  onBeforeRender={(instance) => {
    instance.on({ type: BusEventType.VIEW_CHANGE, handler: ... });
    instance.on({ type: BusEventType.VIEW_PRE_CHANGE, handler: ... });
  }}
  ...
/>
```

That's awkward compared to the custom-element variant, which exposes the hooks directly. Legitimate container-side use cases include:

- Analytics — fire when the chat opens/closes.
- Parent-app state mirror — sync a "chat open" boolean into the rest of the page.
- Side panels / chrome — toggle unrelated UI when the chat opens.
- Modal/focus traps in the host app.

## Suggested fix

Add `onViewChange?` and `onViewPreChange?` as top-level props on:

- `ChatContainerProps` (`packages/ai-chat/src/types/component/ChatContainer.ts`)
- `CdsAiChatContainerAttributes` + the `cds-aichat-container` Lit element (`packages/ai-chat/src/web-components/cds-aichat-container/index.ts`)

Wire them up inside each component's existing `onBeforeRenderOverride` by calling `instance.on({ type: VIEW_PRE_CHANGE | VIEW_CHANGE, handler })` only when the prop is provided.

The custom-element variants should remain unchanged — they already declare and consume these props themselves for the wrapping-element visibility default, and they don't forward them to the inner container in their render templates. So there's no double-firing risk.

## Out of scope

This issue is intentionally scoped to the missing props. The custom-element variants' existing behavior (default visibility toggle when `onViewChange` isn't supplied) should not change.

PR #1420, opened 2026-05-20
feat: add onViewChange, onViewPreChange, renderCustomMessageFooter

### LINKED ISSUE #1379: [Feature Request]: Add renderCustomMessageFooter to web component containers
### The problem

`renderCustomMessageFooter` is exposed as a render-callback prop on the React `<ChatContainer>` and `<ChatCustomElement>` ([packages/ai-chat/src/react/ChatContainer.tsx](https://github.com/carbon-design-system/carbon-ai-chat/blob/main/packages/ai-chat/src/react/ChatContainer.tsx), [ChatCustomElement.tsx](https://github.com/carbon-design-system/carbon-ai-chat/blob/main/packages/ai-chat/src/react/ChatCustomElement.tsx)), where it's wired into `CustomFooterPortalsContainer` so consumers get a one-line API for rendering into per-message footer slots.

The web component counterparts (`<cds-aichat-container>` and `<cds-aichat-custom-element>`) do not expose an equivalent property. They listen to `BusEventType.CUSTOM_FOOTER_SLOT` only far enough to mint pass-through `<div slot="..."><slot name="..."></slot></div>` stubs, leaving consumers to subscribe to the bus event themselves and project their own slotted children (as documented in `packages/ai-chat/docs/WebComponent.md`).

This is asymmetric with `renderUserDefinedResponse`, which **does** have a WC equivalent on both `cds-aichat-container` and `cds-aichat-custom-element` and handles the bus subscription internally. The same convenience should be available for custom message footers.

### The solution

Add a `renderCustomMessageFooter` property to both `<cds-aichat-container>` and `<cds-aichat-custom-element>`, modeled after the existing `renderUserDefinedResponse` implementation:

- Subscribe to `BusEventType.CUSTOM_FOOTER_SLOT` internally when the callback is set.
- Track accumulated per-slot state and pass it (plus the `ChatInstance`) to the consumer's `renderCustomMessageFooter` function.
- Render the returned Lit `TemplateResult` into the matching slot, so consumers don't have to wire bus handlers or manage slot DOM themselves.
- Keep the existing slot-passthrough behavior as the fallback when no callback is provided, for backward compatibility.

Then update `packages/ai-chat/docs/WebComponent.md` to:
- Document the new `renderCustomMessageFooter` property as the recommended approach.
- Keep the manual `CUSTOM_FOOTER_SLOT` bus-handler example as an advanced alternative.

### Examples

Approximate target shape (mirrors `renderUserDefinedResponse`):

```ts
import "@carbon/ai-chat/cds-aichat-container";
import { html } from "lit";

const renderCustomMessageFooter = (state, instance) => html`
  <my-footer
    .slotName=${state.slotName}
    .message=${state.message}
    .instance=${instance}
  ></my-footer>
`;

html`<cds-aichat-container
  .config=${config}
  .renderCustomMessageFooter=${renderCustomMessageFooter}
></cds-aichat-container>`;
```

### Application / product

Carbon AI Chat — web component build (`@carbon/ai-chat/cds-aichat-container`, `@carbon/ai-chat/cds-aichat-custom-element`).

No

### Business priority

Medium Priority = upcoming release but is not pressing

_No response_

### Additional contacts

_No response_

### Available extra resources

_No response_

### Code of Conduct

ethanwinters:
Closes #1379 #1380 

Adds opt-in view-change observation hooks (`onViewChange` / `onViewPreChange`) to the float `ChatContainer` and `cds-aichat-container`, and a managed `renderCustomMessageFooter` callback to the web components so custom message footers can be rendered with the same library-managed contract already used for user-defined responses.

#### Changelog

**New**

- `onViewChange` and `onViewPreChange` props on the React `ChatContainer` and matching properties on the `cds-aichat-container` web component. These are pure observation hooks — useful for analytics or mirroring the chat's open/closed state into your own UI. Unlike `cds-aichat-custom-element`, the float container has no wrapping element to size, so there is no default visibility behavior; a handler is only subscribed when the consumer provides the prop.
- `renderCustomMessageFooter` callback property on `cds-aichat-container` and `cds-aichat-custom-element`. When provided, the library handles all `CUSTOM_FOOTER_SLOT` event listening, per-slot state tracking, and element lifecycle (including `RESTART_CONVERSATION` cleanup) — the web component analogue of the React `RenderCustomMessageFooter`.
- New exported types `RenderCustomMessageFooterState` (accumulated per-slot state) and `WCRenderCustomMessageFooter` (the callback signature), surfaced from both `aiChatEntry` and `serverEntry`.
- Documentation: a `renderCustomMessageFooter` section in `WebComponent.md` (with the prior manual-event approach kept as a "Legacy" subsection), plus `onViewChange` / `onViewPreChange` notes in `React.md` and `WebComponent.md`.

**Changed**

- `cds-aichat-custom-element` now delegates custom-footer rendering to the inner `cds-aichat-container` when `renderCustomMessageFooter` is set, and skips its own `CUSTOM_FOOTER_SLOT` subscription / manual slot wrappers in that case.
- The internal `CustomFooterSlotState` interface in `CustomFooterPortalsContainer.tsx` is now a type alias of the public `RenderCustomMessageFooterState`, giving the shape a single source of truth.
- The `RESTART_CONVERSATION` subscription on `cds-aichat-container` is registered once and clears whichever callback paths are active (user-defined and/or custom-footer), instead of being tied to `renderUserDefinedResponse` alone — avoiding a double-fire when both callbacks are provided.
- Copyright year headers on the two web-component `index.ts` files bumped to `2025, 2026`.

**Removed**

- Nothing user-facing. The legacy event + manual-slot path for custom footers is fully preserved as the default when `renderCustomMessageFooter` is omitted.

#### Testing / Reviewing

- Check the rendered `WebComponent.md` / `React.md` docs and the example snippets compile against the exported types.

### DIFF
--- a/packages/ai-chat/docs/React.md
+++ b/packages/ai-chat/docs/React.md
@@ -63,6 +63,8 @@ This component requires a `className` prop that defines the size and positioning
 
 If you don't want these behaviors, you can also listen for {@link BusEventType.VIEW_PRE_CHANGE} and {@link BusEventType.VIEW_CHANGE} events directly. These events fire in sequence (PRE_CHANGE -> view state update -> CHANGE), and both are awaited, making async handlers ideal for animations. See the event type documentation for complete details on timing and usage. Just be aware that the {@link ChatCustomElementProps.onViewChange} default behavior will still run if you don't replace that function with your own.
 
+{@link ChatContainerProps.onViewChange} and {@link ChatContainerProps.onViewPreChange} are also available on {@link ChatContainer} as opt-in observation hooks — useful for analytics or mirroring the chat's open state into your own UI. The float container has no wrapping element to size, so there is no default visibility behavior on the container: the callbacks simply fire when provided.
+
 See {@link ChatCustomElementProps} for an explanation of the various accepted props.
 
 ```javascript
--- a/packages/ai-chat/docs/WebComponent.md
+++ b/packages/ai-chat/docs/WebComponent.md
@@ -78,6 +78,8 @@ If you don't want these behaviors, then provide your own `onViewChange` prop to 
 
 Just be aware that the `onViewChange` default behavior will still run if you don't replace that function with your own.
 
+`onViewChange` and `onViewPreChange` are also available on `cds-aichat-container` as opt-in observation hooks — useful for analytics or mirroring the chat's open state into your own UI. Because the float container has no wrapping element to size, there is no default visibility behavior on the container: the callbacks simply fire when provided.
+
 See {@link CdsAiChatCustomElementAttributes} for an explanation of the various accepted properties and attributes.
 
 ```typescript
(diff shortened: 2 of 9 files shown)


davidmenendez (approved):
LGTM!

PR #1473, opened 2026-05-29
fix: messages scroll behavior

### LINKED ISSUE #1376: [Bug]: AI Chatbot scroll propagates to page when reaching end of message list
### Browser

Chrome

### Package version

v1.11.0

v.19.2.4

### Description

When scrolling toward the bottom of the AI Chatbot conversation using the mouse wheel, once the end of the message list is reached, the page behind the chatbot starts scrolling as well.

Scrolling should remain contained within the AI Chatbot component. If the user wants to scroll the underlying page, they should move the cursor outside the chatbot area (or close the chatbot).

I found that adding the following style resolves the issue:

```
.cds-aichat--messages__wrapper {
  overscroll-behavior: contain;
}
```

After applying this, scroll chaining to the main page no longer occurs.

### Reproduction / example

### Steps to reproduce

1. Go to the demo site.
2. From the "Choose Chat Component Option” section, set the layout type to “Sidebar”.
3. Open the Developer Console and locate the main `html` element.
4. Set its height to `8000px` to simulate a page with enough content to scroll.
5. Move your mouse over the widget and start scrolling.

You will see that once the chatbot reaches the end of its scrollable area (if there is any), the page behind it starts scrolling as well.

This can be prevented by applying the CSS property mentioned above to the widget’s scroll container.

You can try it as well in the DevTools: search for the class `cds-aichat--messages__wrapper`.
Then apply:

`overscroll-behavior: contain;`

You will see that the background page no longer scrolls, and scrolling is contained within the widget. The page will only scroll again when you move your mouse outside the widget area.

This behavior should be the default, or alternatively controlled via a property exposed by Carbon.

### Suggested severity

Severity 3 = User can complete task, and/or has a workaround within the user experience of a given component.

### Application / product

_No response_

### Code of Conduct

davidmenendez:
Closes #1376 

fixes scroll behavior in the messages widget

#### Changelog

**Changed**

- `packages/ai-chat/src/chat/components-legacy/MessagesComponent.scss` adds `overscroll-behavior: contain;`

#### Testing / Reviewing

`npm run test` and verified locally after following the same reproduction steps mentioned in the issue

### DIFF
--- a/packages/ai-chat/src/chat/components-legacy/MessagesComponent.scss
+++ b/packages/ai-chat/src/chat/components-legacy/MessagesComponent.scss
@@ -1,5 +1,5 @@
 /* 
- *  Copyright IBM Corp. 2025
+ *  Copyright IBM Corp. 2025, 2026
  *  
  *  This source code is licensed under the Apache-2.0 license found in the
  *  LICENSE file in the root directory of this source tree.
@@ -32,6 +32,7 @@
   flex: 1;
   block-size: 100%;
   inline-size: 100%;
+  overscroll-behavior: contain;
 }
 
 .cds-aichat--messages__wrapper:focus {


warrenmblood (approved):
LGTM!

PR #1517, opened 2026-06-02
fix: human agent banner focus on render

### LINKED ISSUE #1476: [A11Y]: human agent disconnect not convenient to access via keyboard
In a use case where the user has a long chat history, it becomes unreasonably difficult to access the `disconnect` button that appears when connected to a human agent. Watch the video below

https://github.com/user-attachments/assets/42d53850-854d-495d-b7fd-0fe65d601689

As you can see, in order to reach the `disconnect` button you have to tab through very single actionable element between the main text input and the button at the top of the chat. There needs to be an easier way to reach this button. A quick solution may just be to move it from the top to the bottom so that you only have to tab once from the main text input to reach it. Another option, which i think is actually the more accessible option, is to assign focus to the disconnect button. Since this is dynamic content that rearranges the page I think this is the preferable option. ultimately I think doing both things would be better for the flow because even if we reassign focus to disconnect the user would still have to tab through the entire chat history to reach the main text input.

I will tag design on this for their input.

davidmenendez:
Closes #1476 

automatically focuses the human agent banner action on render and the confirm modal that appears when you click the disconnect button in the banner.

#### Changelog

**Changed**

- `packages/ai-chat/src/chat/components-legacy/MessagesComponent.tsx` automatically focuses the human agent banner action on render
- `packages/ai-chat/src/chat/components/modals/ConfirmModal.tsx` automatically focuses the secondary button in the modal

#### Testing / Reviewing

run the demo and enter / select `human agent` => click `connect to agent` => verify that the `disconnect` button in the human agent banner is focused => click disconnect => verify that the modal is focused


annawen1:
Running into an issue when I am connected to human agent. Focus keeps going to the "Disconnect" when I try typing in the input field - I'm not able to type in more than one character before focus jumps to the disconnect

https://github.com/user-attachments/assets/1b73566b-1bbf-4c33-a35e-a027b46a0fb4

davidmenendez:
> Running into an issue when I am connected to human agent. Focus keeps going to the "Disconnect" when I try typing in the input field - I'm not able to type in more than one character before focus jumps to the disconnect
> 
>  Screen.Recording.2026-06-02.at.3.07.53.PM.mov

woah! interesting. i didn't encounter this originally. i'll test this more stringently 🔬

davidmenendez:
fixt the bug and also added the logic necessary to focus the confirmation modal that's triggered from the disconnect button

annawen1 (approved):
works now!

PR #1547, opened 2026-06-04
fix(feedback): pass selected category value via data-content attr

### LINKED ISSUE #1546: [Bug]: Feedback categories returning empty
### Browser

_No response_

### Package version

v1.13.0

_No response_

### Description

When categories are selected from the feedback component and submitted, the selected categories array comes back empty.

### Reproduction / example

https://chat.carbondesignsystem.com/components/storybook/tag/latest/index.html?path=/story/preview-feedback--with-categories

### Steps to reproduce

Go to latest storybook: https://chat.carbondesignsystem.com/components/storybook/tag/latest/index.html?path=/story/preview-feedback--with-categories

Select categories from the feedback options and hit "submit"
An alert should appear with the selected categories, but it is reporting none.

https://github.com/user-attachments/assets/09790e51-2c1b-4718-9b38-9915a845230d

### Suggested severity

Severity 2 = User cannot complete task, and/or no workaround within the user experience of a given component.

### Application / product

_No response_

### Code of Conduct

annawen1:
Closes #1546

Regression where selected categories from the feedback component are not being returned. See video.

https://github.com/user-attachments/assets/09790e51-2c1b-4718-9b38-9915a845230d

The `_handleCategoryClick` method retrieves the category value via `button?.getAttribute("data-content")`, but this attribute is not set on the rendered tags, causing it to return `null` and preventing categories from being added to the selection.

Added `data-content="${value}"` attribute to each `<cds-selectable-tag>` element so the click handler can properly identify which category was selected.

#### Changelog

**New**

- Added test case to cover this issue

**Changed**

- Add `data-content="${value}"` to selectable-tag so the value can be retrieved

#### Testing / Reviewing

Go to WC and React storybook deploy previews
Go to Feedback story with categories
Select a few categories and hit submit
You should see the selected categories listed in the alert

### DIFF
--- a/packages/ai-chat-components/src/components/feedback/__tests__/feedback.test.ts
+++ b/packages/ai-chat-components/src/components/feedback/__tests__/feedback.test.ts
@@ -46,4 +46,51 @@ describe("feedback", function () {
     );
     await expect(el).dom.to.equalSnapshot();
   });
+
+  it("should include selected categories in submit event", async () => {
+    const categories = ["Inaccurate", "Unhelpful", "Not relevant"];
+    let submittedCategories: string[] | undefined;
+
+    const el = await fixture<Feedback>(
+      html`<cds-aichat-feedback
+        is-open
+        show-text-area
+        .categories=${categories}
+        @feedback-submit=${(event: CustomEvent) => {
+          submittedCategories = event.detail.selectedCategories;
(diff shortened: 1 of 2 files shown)


warrenmblood (approved):
LGTM

PR #1579, opened 2026-06-11
fix: processing failed with no loop value set

ethanwinters:
Fixes the `Processing` component's `quickLoad` variant, which was effectively dead: there was no CSS implementing `.quick-load`, and the Storybook bindings used `?quickLoad` (the property's attribute is `quick-load`), so the prop never took effect. This also clarifies the loop/no-loop class logic so the non-looping variant reliably renders when `loop` is omitted, and opts the chat typing indicator into `quickLoad` so the dots appear immediately instead of waiting out the ~1s entry delay.

- [`packages/ai-chat-components/src/components/processing/src/processing.scss`](packages/ai-chat-components/src/components/processing/src/processing.scss) — adds the new `.quick-load` keyframe-delay rules (and the RTL mirror); review the per-dot `animation-delay` timing values, which shift the whole sequence ~1s earlier while only overriding the delay so it composes with both `.linear` and `.linear--no-loop`.
- [`packages/ai-chat-components/src/components/processing/src/processing.ts`](packages/ai-chat-components/src/components/processing/src/processing.ts) — reworks the `classMap` so `loop`/`!loop` drive `linear`/`linear--no-loop` and `quick-load` is applied independently of the loop variant.

#### Changelog

**New**

- `.quick-load` animation rules (LTR + RTL) so `quickLoad` finally does something — it removes the ~1s entry delay and now composes with both the looping and non-looping variants.

**Changed**

- [`MessagesTypingIndicator`](packages/ai-chat/src/chat/components-legacy/MessagesTypingIndicator.tsx) now passes `quickLoad` alongside `loop`, so the chat typing indicator shows instantly.
- `classMap` in [`processing.ts`](packages/ai-chat-components/src/components/processing/src/processing.ts) reworked: `loop` drives `linear`/`linear--no-loop`, and `quick-load` is applied on its own rather than as a mutually exclusive variant.
- Storybook stories now bind `?quick-load` (was `?quickLoad`, which never set the attribute) and the `QuickLoad` stories set `loop` too, to demo composition; MDX docs reworded and the React stories file gained the standard license header.
- Tests updated to cover the no-loop fallback when `loop` is unset and `quickLoad` composing with both the looping and non-looping variants.

**Removed**

- None.

#### Testing / Reviewing

This is the chat typing indicator, so it's reachable in the demo, but the variants are easiest to review in Storybook plus the unit suite.

- **Storybook (Lit/React)** — Open the **Processing** stories. `QuickLoad` (now `loop` + `quickLoad`) should animate in immediately with no entry delay; `LinearLoop` and `LinearNoLoop` should be unchanged.
- **Demo** — with the watcher running (`npm run aiChat:start`), send a message in the demo chat; the typing indicator dots should appear instantly rather than after the ~1s delay.

### DIFF
--- a/packages/ai-chat-components/src/components/processing/__stories__/processing-react.mdx
+++ b/packages/ai-chat-components/src/components/processing/__stories__/processing-react.mdx
@@ -1,12 +1,16 @@
-import { ArgTypes, Markdown, Meta } from '@storybook/addon-docs/blocks';
+import { ArgTypes, Markdown, Meta } from "@storybook/addon-docs/blocks";
 import * as ProcessingReactStories from "./processing-react.stories.jsx";
 
-<Meta of={ProcessingReactStories} parameters={{ previewTabs: { "storybook/docs/panel": { title: "Overview" } } }} />
+<Meta
+  of={ProcessingReactStories}
+  parameters={{
+    previewTabs: { "storybook/docs/panel": { title: "Overview" } },
+  }}
+/>
 
 # Processing
 
 [Source code](https://github.com/carbon-design-system/carbon-ai-chat/tree/main/packages/ai-chat-components/src/components/processing) &nbsp;|&nbsp; [Feedback](https://github.com/carbon-design-system/carbon-ai-chat/issues)
-
 
 ## Table of Contents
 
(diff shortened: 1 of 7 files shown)


davidmenendez (approved):
LGTM!

PR #1589, opened 2026-06-15
feat: add slots for actions and expanded prompt bar

### LINKED ISSUE #1350: Add actions menu start and end slots

ethanwinters:
Closes #1350 

Adds two host-writeable slots **inside** the input composer — `PROMPT_LINE_ACTIONS_END` (after the action buttons, before the prompt line) and `PROMPT_LINE_SEND_BUTTON_START` (after the prompt line, directly before the send button) — so hosts can inject content right next to the prompt line, not just before/after the whole input (which `BEFORE_INPUT_ELEMENT` / `AFTER_INPUT_ELEMENT` already cover). They reuse the input shell's existing `message-actions` / `send-control` slots, and they're gated on content: an unused slot collapses out of the flex flow (no layout shift), while a populated one stays a real `data-floating-menu-container` box so Carbon tooltips/popovers slotted into it position correctly. This PR does nothing to update the styles to the new design, just adding the slots and the expanded functionality. 

Files with particularly complex changes worth a closer look:

- [`packages/ai-chat/src/chat/components/input/PromptLineWriteableSlot.tsx`](packages/ai-chat/src/chat/components/input/PromptLineWriteableSlot.tsx) — the occupancy detector. It renders the real box only while the flattened slotted content is non-empty, watching **both** `slotchange` **and** a `MutationObserver` on the assigned subtree. The two signals are needed because the three delivery paths fill the slot differently: web-component `slot=` children and the imperative `instance.writeableElements` API change slot assignment (caught by `slotchange`), while the React `renderWriteableElements` portal mutates the assigned host `<div>`'s children asynchronously (caught by the observer). Review the re-observe-on-`slotchange` logic in the effect.
- [`packages/ai-chat-components/src/components/input/src/input-shell.ts`](packages/ai-chat-components/src/components/input/src/input-shell.ts) — excludes the marked passthrough (`data-prompt-line-slot`) from the `--has-message-actions` occupancy check, so an empty/always-present prompt-line slot doesn't shrink the input's start padding.

#### Major changes

**New**

- Two new writeable elements (slots) on the public `@carbon/ai-chat` API: [`WriteableElementName.PROMPT_LINE_ACTIONS_END`](packages/ai-chat/src/types/instance/WriteableElements.ts) (`"promptLineActionsEnd"`) and `WriteableElementName.PROMPT_LINE_SEND_BUTTON_START` (`"promptLineSendButtonStart"`). They're available through the same mechanisms as existing writeable elements — `renderWriteableElements` (React), `slot="..."` (web component), and `instance.writeableElements[...]` (imperative) — and render inside the input composer: actions-end after the action buttons and before the prompt line, send-button-start after the prompt line and directly before the send button.

#### Minor changes

**New**

- [`PromptLineWriteableSlot`](packages/ai-chat/src/chat/components/input/PromptLineWriteableSlot.tsx) — internal component that projects each slot into the input shell and occupancy-gates it (real `data-floating-menu-container` box only when populated; `hidden` when empty so it adds no flex gap before the prompt line or send button).
- Demo coverage for both slots in the React ([`DemoApp.tsx`](demo/src/react/DemoApp.tsx)) and web-component ([`demo-app.ts`](demo/src/web-components/demo-app.ts) enumerates writeable elements automatically) demo apps, rendered compact via the `writeable-element-example` components when **"Show writeable elements"** is enabled.
- Tests: a contract test for `PromptLineWriteableSlot` ([`promptLineWriteableSlot_spec.tsx`](packages/ai-chat/tests/components/spec/promptLineWriteableSlot_spec.tsx)) and an occupancy-exclusion case in the [`input-shell` test](packages/ai-chat-components/src/components/input/src/__tests__/input-shell.test.ts).

**Changed**

- `input-shell` now ignores the prompt-line passthrough in its `--has-message-actions` padding check and adds a `gap` to the send-control container so a populated send-button-start slot is spaced from the send button.
- Docs: [`AGENTS.md`](AGENTS.md) notes to confirm `aiChat:start` isn't already running before kicking off a `build` gate (the watcher owns builds).

#### Testing / Reviewing

This is fully reachable from the demo site (no new config toggle — it rides the existing **Writeable element visibility** switcher).

1. Run the demo (`npm run start --workspace=@carbon/ai-chat-examples-demo`, or use the `aiChat:start` dev server).
2. In the sidebar, open **Page Settings → Writeable element visibility** and select **"Show writeable elements."**
3. In the input row at the bottom, confirm two green example chips: **`promptLineActionsEnd`** sits at the start of the input (after the action area, before the prompt line), and **`promptLineSendButtonStart`** sits at the end (after the prompt line, directly before the send button).
4. Set the same dropdown back to **"Hide writeable elements"** and confirm the prompt-line placeholder and the send button sit flush — **no extra gap or shift** (the unused slots collapse out of the flex flow).
5. Repeat steps 2–4 with the **Framework** switcher set to **Web component** — its slot-forwarding path differs from React, so verify both behave identically (chips in the right spots when shown, no gap when hidden).
6. _(Optional, for the `data-floating-menu-container` guarantee)_ slot a Carbon component that opens a tooltip/popover into one of the slots and confirm it anchors/positions against the slot box.

Unit coverage: `npm run test --workspace=@carbon/ai-chat` (includes `promptLineWriteableSlot_spec.tsx`) and `npm run test --workspace=@carbon/ai-chat-components` (includes the `input-shell` occupancy case).


packages/ai-chat-components/src/components/input/src/input-shell.scss
```diff
@@ -87,20 +87,15 @@
 }
 
 // -----------------------------------------------------------
-// Field container (text area + char count) — shares a row with the send
-// control. `flex: 1` lets it consume all remaining space next to send.
+// Text + actions row — message actions sit beside the editor and share the
+// row with the send control. `flex: 1 0` lets it consume all remaining space
+// next to send. In the expanded variant this collapses to `display: contents`
+// so the editor and actions reflow onto their own rows (see the expanded
+// block at the end of this file).
```
warrenmblood, line 94: I think we want to consider how these elements should be positioned when in float or sidebar layout and they are forced to wrap. In the demo the positioning of the send button and promptLineSendButtonStart is different depending on whether the prompt line is expanded. Is this intentional?

Not expanded:

Image

Expanded:

Image

packages/ai-chat-components/src/components/input/src/input-shell.ts
```diff
@@ -62,15 +73,13 @@ class InputShellElement extends LitElement {
...
-                name="message-actions"
-                @slotchange=${this._handleMessageActionsSlotChange}
-              ></slot>
-              <div class="${prefix}--input-text-area">
-                <slot name="editor"></slot>
-              </div>
+          <div class="${prefix}--input-text-and-actions">
+            <slot
+              name="message-actions"
+              @slotchange=${this._handleMessageActionsSlotChange}
```
warrenmblood, line 93: In the WC demo specifically in float mode, after showing and then hiding writeable elements the class `cds-aichat--input-container--has-message-actions` isn't going away and the smaller padding persists:

Image

In react this class isn't getting added at all with "Show writeable elements"

warrenmblood (changes requested):

packages/ai-chat-components/src/components/input/src/input-shell.scss
ethanwinters: thanks totally missed the other layouts.

warrenmblood (approved):
LGTM Thanks!

PR #1595, opened 2026-06-15
fix(scroll): wait for markdown to render before triggering scroll logic

### LINKED ISSUE #1359: [Bug]: auto scroll positioning for request messages that are too long
### Browser

_No response_

### Package version

v1.12.0

_No response_

### Description

When entering a message that is too long (more than 1/4 the chat height specifically), the auto scroll logic should adjust so that only the bottom 100px of the request message shows. This is so user is able to see the response message without having to scroll down.

We have the [logic in place](https://github.com/carbon-design-system/carbon-ai-chat/blob/main/packages/ai-chat/src/chat/utils/messagesAutoScrollController.ts#L246-L270) but it looks like there was a regression and is no longer working.

### Reproduction / example

### Steps to reproduce

- Go to the demo site: https://chat.carbondesignsystem.com/tag/latest/demo/index.html
- Input a super long message and hit send.
- Notice the scroll sits at the top of the request message 

Image

### Suggested severity

None

### Application / product

_No response_

### Code of Conduct

annawen1:
Closes #1359

We have logic where when there is a tall request message (more than 1/4 of the chat height), it should scroll and only reveal the bottom 100px of that tall request message. There was a regression with this logic.

The request message height was being measured before it was rendered and so none of the messages were considered tall. The messages are rendered with `cds-aichat-markdown`, we have to wait for the markdown component to fully render before we trigger the measure and scroll logic.

WITH FIX:
https://github.com/user-attachments/assets/53a4015a-f8bf-4b49-9b21-a1665710e453

#### Changelog

**Changed**

- Add `waitForMessageComponentLayout()` taps into the markdown component, waits for the Lit `updateComplete` promise to resolve, and then waits one browser frame to apply layout so we are sure the message is rendered before we allow for the scroll logic to trigger

#### Testing / Reviewing

- Go to demo deploy preview and insert an extra long message in the input field, hit enter
- only the bottom 100px of the message should show
- enter a one line message and make sure it renders as it does currently with short messages (with extra spacing above the top of the request message)

### DIFF
--- a/packages/ai-chat/src/chat/components-legacy/MessagesComponent.tsx
+++ b/packages/ai-chat/src/chat/components-legacy/MessagesComponent.tsx
@@ -511,19 +511,13 @@ class MessagesComponent extends PureComponent<MessagesProps, MessagesState> {
       return;
     }
 
-    const markdownElements = Array.from(
-      targetElem.querySelectorAll(`${prefix}-markdown`),
-    );
+    const markdownElement = targetElem.querySelector(`${prefix}-markdown`);
 
-    if (!markdownElements.length) {
+    if (!markdownElement) {
       return;
     }
 
-    await Promise.all(
-      markdownElements.map((element) =>
-        element.updateComplete?.catch((): undefined => undefined),
-      ),
-    );
+    await markdownElement.updateComplete?.catch((): undefined => undefined);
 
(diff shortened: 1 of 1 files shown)


warrenmblood (approved):
LGTM

PR #1598, opened 2026-06-15
fix: stop unneeded re-renders

ethanwinters:
Contributes to #1650 

The chat was re-rendering far more than it needed to. Small things — typing a character, switching the theme, or the surrounding app just re-rendering for its own reasons — could cause the **entire** chat to re-render. This PR fixes that so each part of the chat re-renders only when data it actually uses changes. It touches the four components apps use to add the chat to a page, plus the shared store and config plumbing underneath them.

## The four entry points

"Entry points" are the four components an app uses to drop the chat onto a page:

- **React:** `ChatContainer` and `ChatCustomElement`
- **Web components:** `cds-aichat-container` and `cds-aichat-custom-element`

Each one used to list every config field by hand and pass the config down in several separate props (`strings`, `serviceDesk`, `serviceDeskFactory`, `markdown`, and the rest). Those hand-written lists slowly fell out of sync with each other and with the real config type. Worse, the React container rebuilt its `config` object from a hand-written list of dependencies — miss one field and the object got a fresh reference too often (extra re-renders) or held a stale one (wrong UI).

Now all four are driven by **one shared list of fields** (`FLATTENED_PUBLIC_CONFIG_FIELDS`, in the new `flattenedPublicConfig.ts`), and the build **fails** if someone adds a config field without adding it to that list. Everything is packed into a single `config` object, and that object keeps the **same reference** when nothing changed — so re-renders downstream don't fire for no reason. This single-config-channel change is the foundation everything below builds on.

<details>
<summary>Files changed</summary>

- `packages/ai-chat/src/web-components/shared/flattenedPublicConfig.ts` (new shared field list + `resolveFlattenedConfig` + the build-time check)
- `packages/ai-chat/src/react/ChatContainer.tsx` (memo dependencies come from the shared list; DOM props split off using it)
- `packages/ai-chat/src/react/ChatCustomElement.tsx`
- `packages/ai-chat/src/web-components/cds-aichat-container/cds-aichat-internal.tsx`
- `packages/ai-chat/src/web-components/cds-aichat-container/index.ts`
- `packages/ai-chat/src/web-components/cds-aichat-custom-element/index.ts`
- `packages/ai-chat/src/chat/ChatAppEntry.tsx` (now receives a single `config` prop; the separate side props are gone)
</details>

## The optimizations

Each one below is a separate kind of fix. The collapsed "Files changed" lists let you review them one at a time.

### 1. Replace two Contexts with narrow selectors

Two pieces of data — the **language pack** (all the chat's UI text) and the **markdown settings** — used to be shared through React Context. Because any Context change re-renders *every* reader, editing a single string re-rendered everything that touched the language pack. We removed both Contexts. Components now read only the exact strings they need from the store with `useSelector`. For example, the markdown component used to grab the whole language pack; now it reads only the ~10 strings it actually displays (with `shallowEqual`), so changing an unrelated string no longer re-renders it.

<details>
<summary>Files changed</summary>

- Deleted: `contexts/LanguagePackContext.tsx`, `providers/LanguagePackProvider.tsx`, `hooks/useLanguagePack.tsx`, `contexts/MarkdownConfigContext.ts`, `types/utilities/HasLanguagePack.d.ts`
- `packages/ai-chat/src/chat/components/util/MarkdownWithDefaults.tsx` (reads just the strings it needs, with `shallowEqual`; markdown via a selector)
- `packages/ai-chat/src/chat/store/selectors.ts` (`selectLanguagePack`)
- `packages/ai-chat/src/chat/ChatAppEntry.tsx` (removed the two Context providers)
</details>

### 2. Read single fields instead of whole objects

Some components subscribed to a big object (like the entire config) and so re-rendered whenever **anything** inside it changed. We narrowed them to read just the field they use. For example, `AppShell` now reads the theme, layout, header, and so on each through its own selector — so switching the theme re-renders only the theme-related parts, not the message list. Where a selector has to build a small object, we pair it with `shallowEqual` so that typing a character (which changes the input text, not these fields) no longer re-renders the file-upload button or the stop button.

<details>
<summary>Files changed</summary>

- `packages/ai-chat/src/chat/AppShell.tsx` (one selector per config field instead of reading the whole config)
- `packages/ai-chat/src/chat/hooks/useShouldSanitizeHTML.tsx`
- `packages/ai-chat/src/chat/hooks/useInputCallbacks.tsx`
- `packages/ai-chat/src/chat/store/selectors.ts` (`selectInputUploadAndStreamingFields`)
</details>

### 3. Keep object references stable when config changes

When config changes while the chat is running, the chat rebuilds its whole internal config object from scratch. Rebuilding from scratch gives **every** part a brand-new reference — which would undo the narrow selectors from #2, because they'd all see "a new object" and re-render. We added a step that, right after the rebuild, **reuses the previous reference for every part whose data didn't actually change.** So toggling one field (say `input.isDisabled`) hands a new reference only to the one part that changed; everything else stays `===`-equal and skips re-rendering. We also stopped deep-copying the entire store when only config changed, and we skip the update entirely when nothing changed at all.

<details>
<summary>Files changed</summary>

- `packages/ai-chat/src/chat/store/doCreateStore.ts` (`reconcileAppConfigReferences`, `reconcileObjectReferences`)
- `packages/ai-chat/src/chat/store/reducers.ts` (shallow-copy instead of deep-copy when only config changed)
- `packages/ai-chat/src/chat/utils/dynamicConfigUpdates.ts` (reuse references before saving; skip a no-change update)
</details>

### 4. Move frequently-changing data out of the config object

The language pack and markdown settings used to live **inside** the config object. That meant changing a single string counted as "config changed" and rippled outward. We moved them to their own top-level spots in the store, so they can change without touching config — and config changes don't touch them.

<details>
<summary>Files changed</summary>

- `packages/ai-chat/src/types/state/AppState.ts` (new `languagePack` and `markdownConfig` slices)
- `packages/ai-chat/src/types/state/AppConfig.d.ts` (removed `derived.languagePack`)
- `packages/ai-chat/src/types/config/MarkdownConfig.ts` (new type for the markdown slice)
- `packages/ai-chat/src/chat/store/doCreateStore.ts` (`buildLanguagePack`, slice setup)
</details>

### 5. Compute input flags when read, instead of copying them into the store

Whether the input is visible / disabled / read-only used to be **copied** from config into the store at startup, and re-copied on every config change. We removed the copying. The store now holds only a runtime **override** (empty — `null` — unless something like the human-agent flow sets it), and small selectors compute the real value from config at the moment it's read. That removes a whole category of store updates whose only job was to copy config, and there's no longer a copy that can drift out of sync.

<details>
<summary>Files changed</summary>

- `packages/ai-chat/src/chat/store/reducerUtils.ts` (`DEFAULT_INPUT_STATE`: overrides start as `null`)
- `packages/ai-chat/src/types/state/AppState.ts` (`InputState` now stores overrides, not copies)
- `packages/ai-chat/src/chat/store/selectors.ts` (`selectInputIsReadonly`, `selectInputIsDisabled`, `selectInputFieldVisible`)
- `packages/ai-chat/src/chat/store/doCreateStore.ts` (startup no longer copies config into the input state)
- `packages/ai-chat/src/chat/utils/dynamicConfigUpdates.ts` (removed the block that re-copied these on every change)
</details>

### 6. Protect the big memoized component from unstable props

`AppShell` is the large, expensive part of the chat, wrapped in `React.memo` so it can skip re-rendering when its props are unchanged. The problem: it was handed the `renderWriteableElements` object, which apps passing live data rebuild every render — a new reference each time. That broke the `React.memo` check and re-rendered the whole shell constantly. Now `AppShell` gets only a small, stable string listing **which** slots have content (which rarely changes), so the memo holds. The actual live content still flows through — but only to small sibling components that are *meant* to update.

<details>
<summary>Files changed</summary>

- `packages/ai-chat/src/chat/ChatAppEntry.tsx` (computes the stable slot-list string; routes live render props to the small siblings)
- `packages/ai-chat/src/chat/AppShell.tsx`
- `packages/ai-chat/src/chat/AppShellWriteableElements.tsx`
</details>

## Supporting changes

- **Tell Lit to compare config by content, not by reference.** Lit (the web-component layer) decides "did this prop change?" with `===` by default. Our config getter builds a fresh object each render, so `===` was always "changed" and re-rendered the React app every time. We told Lit to compare config by its **contents** instead, so an identical config no longer triggers a re-render. *(`cds-aichat-internal.tsx`)*
- **One place rebuilds the text/date formatter.** Several spots used to rebuild the localization formatter, and they could disagree and leave stale text. Now a single store listener rebuilds it whenever the strings or locale change — one source of truth. *(`store/subscriptions.ts`, `utils/intlUtils.ts`, `services/loadServices.ts`)*
- **A debug warning for unstable props.** When `debug` is on, the chat prints a one-time warning if an app passes an object prop that's a new reference every render but has the same contents — a hint to wrap it in `useMemo`. Docs explain which props to keep stable. *(`ChatAppEntry.tsx`, `types/AGENTS.md`, `types/config/PublicConfig.ts`)*
- **Removed an old "what changed" helper.** Deleted a helper that pre-computed a bag of "what changed" booleans; the update function now just compares the previous and next config directly, which is simpler and shorter. *(deleted `utils/configChangeDetection.ts` and its tests; `utils/dynamicConfigUpdates.ts`)*

## Tests

New tests lock this behavior in so it can't quietly regress: tests that count how many times components render, that check object references are reused, that cover the new input-flag selectors, the human-agent factory behavior, the localization listener, that all four entry points behave identically, and that the shared field list stays complete. They live under `tests/store/spec/`, `tests/config/spec/`, and `tests/web-components/spec/`.

## Manual testing strategy

This was a big refactor, so please exercise the demo before approving. Start it with `npm run aiChat:start` (this builds the library and serves the demo; if a teammate already has it running, just open the URL it printed). Open your browser DevTools (install React Profiler if you have not before). The two goals are: **show the re-render wins are real** (React Profiler) and **show nothing broke** (functional checklist).

### The three tools you'll use

Almost everything below is done with one of these. Read this section once, then the checklists are quick.

**1. The demo sidebar (left side of the page).** A set of collapsible sections with dropdowns. The ones this PR cares about:

- **Choose Chat Component → Framework** — React vs web component.
- **Choose Chat Component → Layout Type** — `float` uses the *container* component; `sidebar` and `fullscreen` use the *custom element* component. Framework × Layout Type together pick which of the four entry points you're testing. (Changing either reloads the page — that's expected.)
- **Page Settings → Page Theme** — the surrounding page's theme. **Page Settings → Text Direction** — left-to-right vs right-to-left.
- **Chat Configuration → Carbon Theme** — the chat's theme, or "Inherit from parent" to follow the page.
- **Chat Configuration → Input** — dropdowns for "Input field visibility," "Input field state" (enabled/disabled), and file uploads.
- **Chat Configuration → Homescreen & disclaimer**, **Launcher** (has the unread-indicator dropdown), **Header**, **Layout**, **History**, **Feedback**, **Keyboard Shortcuts**, **Messaging**.
- **Chat instance methods** (this section appears once the chat has loaded) — buttons for focus, restart, loading counters, open/close, and panels.

**2. `window.setChatConfig({ ... })` — typed into the DevTools Console.** This applies any config field live, including ones with **no** sidebar control: `isReadonly`, `strings`, `locale`, `namespace`, `markdown`, and `debug`. Two things to know:

- It **shallow-merges**, so pass the whole nested object you want — e.g. `window.setChatConfig({ input: { isDisabled: true } })`, not just the leaf value.
- The first call switches the demo into "setChatConfig mode," which **replaces the sidebar** with a single "Leave setChatConfig Mode" button. Click that button (it refreshes the page) to get the normal sidebar back. So do your sidebar checks first, then your Console checks.

**3. `window.chatInstance.<method>(...)` — typed into the Console.** The chat instance is exposed globally. Use it for instance methods that have no sidebar button, e.g. `window.chatInstance.updateInputFieldVisibility(false)`, `window.chatInstance.updateInputIsDisabled(true)`, `window.chatInstance.updateAssistantUnreadIndicatorVisibility(true)`.

**Producing chat messages.** Open the chat first (click the launcher button in `float` layout; it's already open in `sidebar`/`fullscreen`). Then type one of these **exact** phrases into the input to get that kind of reply: `text`, `text with feedback`, `text (stream)`, `grid`, `table`, `card`, `carousel`, `code`, `iframe`, `image`, `option list`, `user_defined`, `human agent`. Any other text returns a generic welcome reply.

### Setup for profiling

- Install the **React Developer Tools** browser extension. It adds **Components** and **Profiler** tabs to DevTools.
- In **Components** → gear icon → turn on **"Highlight updates when components re-render."** Now any component that re-renders briefly flashes an outline on the page — a re-render storm becomes impossible to miss.
- In **Profiler** → gear → turn on **"Record why each component rendered."** Press record, do one action, press stop, and it lists exactly what re-rendered and why.
- The demo updates its own state every ~2 seconds on purpose, to imitate a busy host app. Leave it **on** for the first test below; add **`?disableParentStateTimer`** to the URL (e.g. `http://localhost:PORT/?disableParentStateTimer`) to turn it off for a quiet baseline when profiling a single action.

### Part 1 — Show the re-render wins (React Profiler)

1. [ ] **The headline test.** Open the demo with the timer **on** and "Highlight updates" turned on. Watch for ~10 seconds. **Pass:** only the small area showing the changing host value flashes every ~2s. **Fail:** the header, message list, or whole chat flashes too. (Protects #6.)
2. [ ] **Typing.** Add `?disableParentStateTimer`, open the chat, start a Profiler recording, type a few characters, stop. **Pass:** only the input area re-rendered. **Fail:** the message list or the shell re-rendered. (Protects #2.)
3. [ ] **Theme switch.** Record while changing **Chat Configuration → Carbon Theme**. **Pass:** only theme-related parts re-render. (Protects #2 and #3.)
4. [ ] **One config field.** Record while changing **Input → "Input field state"** between Default / Enable / Disable. Use "why did this render" to confirm the header, launcher, and theme did **not** re-render. (Protects #3.)
5. [ ] **A string change.** In the Console run `window.setChatConfig({ strings: { input_placeholder: "Type here..." } })`. **Pass:** the input placeholder updates and the message list does not re-render. (Protects #1 and #4.) Afterward click "Leave setChatConfig Mode" to restore the sidebar.

### Part 2 — Show nothing broke (functional checklist)

Do each both at startup and again live (change it with a sidebar dropdown or `window.setChatConfig` while the chat is open — the live path is the one this PR reworked most).

1. [ ] **Input visible / disabled / read-only (highest risk — this logic was rewritten).**
   - Sidebar **Input → "Input field visibility"** → Hide, then Show: the input hides and comes back.
   - Sidebar **Input → "Input field state"** → Disable, then Enable: the input and send button disable and re-enable.
   - Console `window.setChatConfig({ isReadonly: true })`: the input becomes read-only (then "Leave setChatConfig Mode").
   - Default (the "Default" dropdown option) → input is visible, enabled, and editable.
   - Runtime override still wins: with the input shown, run `window.chatInstance.updateInputFieldVisibility(false)` → it hides; `window.chatInstance.updateInputIsDisabled(true)` → it disables.
   - Human-agent path: type `human agent`, click **Connect**; the input behaves correctly during the agent chat and returns to normal when it ends.
2. [ ] **Text & locale.** `window.setChatConfig({ strings: { input_placeholder: "Hola" } })` → placeholder changes; reset with `window.setChatConfig({ strings: {} })` → back to default. `window.setChatConfig({ locale: "fr" })` → message timestamps switch to French formatting. Use **Page Settings → Text Direction** to try RTL.
3. [ ] **Markdown.** Send `text` and `code`, then change markdown settings with `window.setChatConfig({ markdown: { ... } })` and confirm the rendered markdown updates. (If you'd rather exercise a custom renderer, use the optional temp-edit at the bottom.)
4. [ ] **Human agent / service desk.** Type `human agent`, click **Connect**, and exchange a message through the mock desk. **Key regression:** with the timer **on** (host re-rendering every 2s), the connection must **stay alive** — it must not disconnect or restart. End the chat and confirm it returns to the assistant cleanly.
5. [ ] **Disclaimer.** Sidebar **Homescreen & disclaimer → "Disclaimer Only,"** then reload: you must accept the disclaimer before chatting. Accept it and reload again: it stays accepted. Set the dropdown back to "None."
6. [ ] **Home screen.** Sidebar **Homescreen & disclaimer** → pick a homescreen option → the home screen shows on a fresh conversation; turn it off → it's gone.
7. [ ] **Namespace.** `window.setChatConfig({ namespace: "alt" })`, send a message, reload — the "alt" namespace keeps its own separate history.
8. [ ] **Theme inherit mode.** Sidebar **Carbon Theme → "Inherit from parent,"** then change **Page Theme** → the chat follows the page. Switch Carbon Theme to an explicit value (e.g. "Dark (g90)") → it stops following the page.
9. [ ] **Launcher unread dot.** Sidebar **Launcher** → set the unread-indicator dropdown on; close and reopen the chat window per its rules; confirm an unrelated change (e.g. theme) doesn't bring a dismissed dot back.
10. [ ] **Writeable elements / custom responses.** Sidebar **Page Settings → Writeable Elements** on; type `user_defined`; with the timer on, the host value inside still updates without re-rendering the rest of the chat.
11. [ ] **The rest of the switchers.** Header, Layout, Launcher, Keyboard Shortcuts, Feedback, History — change each and confirm it applies live and at startup.
12. [ ] **General smoke test.** Type each of `text`, `text with feedback`, `grid`, `table`, `card`, `carousel`, `code`, `iframe`, `image`, `option list`, `user_defined`; try `text (stream)` with the stop button; enable **Input → File uploads** and attach a file. No console errors (other than intentional `debug` warnings).

### Part 3 — Check all four entry points

- [ ] Repeat at least items 1, 2, and 4 above on each of the four entry points. Use **Choose Chat Component → Framework** (React / web component) and **Layout Type** (`float` = container component, `sidebar` = custom element component) — that's four combinations. Config now flows through one shared list, so a bug here looks like a setting that works on one combination but not another.


annawen1:
https://github.com/user-attachments/assets/ccdaec69-6553-4c09-86e3-b3e65d38952b

> Input flags don't churn the tree — open the Input config switcher and toggle isDisabled / isVisible. Only the input region should update; the message list, header, and launcher should not re-render. The input should reflect the new state immediately (now derived from config via the new selectors rather than mirrored into state).

It looks like other parts of the chat still re-render when disabling / enabling the input field

ethanwinters:
For 1.17.0, want us to have to use this in the branch for a couple weeks.

ethanwinters:
@davidmenendez i just need to double check that this work actually fixes the #1650 issue now that we have full reproduction steps. When I verify, will take out of draft.

davidmenendez (approved):
LGTM!

PR #1652, opened 2026-06-25
fix: a11y announcements for file upload

### LINKED ISSUE #1609: Screen reader announcements for file upload
### Problem
The file-upload / attachment flow doesn't announce its state changes to screen readers. The **assistant upload path announces nothing**; the **human-agent path announces only success** (`fileSharing_ariaAnnounceSuccess`, used solely in `packages/ai-chat/src/chat/services/haa/HumanAgentServiceImpl.ts`) — it has the same gaps otherwise. Two announcement strings already exist in the language pack but have **zero code usages** (`fileSharing_uploadFailed`, `fileSharing_fileTooLarge`), suggesting the work was started but never wired up.

### Current coverage
| Transition | Announced? |
|---|---|
| File added | ❌ none |
| File removed | ❌ none |
| Upload starts | ❌ no announcement (a visual "Uploading file" label exists via `fileSharing_statusUploading`, but it's an `aria-label`/assistive text on the spinner, not a live-region status message) |
| Upload succeeds — assistant | ❌ none |
| Upload succeeds — human agent | ✅ `fileSharing_ariaAnnounceSuccess` |
| Upload fails | ❌ none. Assistant: `fileSharing_uploadFailed` defined but unused. Human-agent: the `setFileUploadStatus` error branch adds a **visible inline error message** to the conversation but still never announces it. |
| Validation: file too large | ❌ none (`fileSharing_fileTooLarge` defined but unused) |
| Validation: unsupported type / too many files | ❌ none (browser-level / silently disabled button) |

> The added / removed / upload-start / failure / validation rows apply to **both** the assistant and human-agent paths — they share the input flow, and the human-agent path only layers the success announcement on top. So this is **not** "assistant catches up to a complete human-agent path"; both need the full set.

### Requirements
- Announce upload state changes via the existing live regions: file added / removed, upload start, success, failure — **polite**; validation/upload **errors** (too large, unsupported type, max files) — **assertive** (`role="alert"`), since the user must act.
- Wire up the already-defined strings (`fileSharing_uploadFailed`, `fileSharing_fileTooLarge`, `fileSharing_statusUploading`, `fileSharing_ariaAnnounceSuccess`) and add any missing keys (file added/removed).
- Cover **both** the assistant and human-agent paths. The human-agent path currently announces only success; it still lacks added / removed / start / **failure** / validation. Its failure branch (`setFileUploadStatus` with `isError`) even adds a visible inline error message to the conversation without announcing it — fix that specifically.

### Where / how
- Reuse the React announcer: `store.dispatch(actions.announceMessage({ messageID, messageValues }))` → `AriaAnnouncerProvider` (3 polite live regions). Already used for streaming/reasoning starts and the human-agent upload success.
- Assistant-path hook points in `packages/ai-chat/src/chat/services/ChatActionsImpl.ts`: the pending-upload add (file added / upload start), the remove-pending-upload path (file removed), and the upload status-update success and error branches.
- Human-agent hook points in `packages/ai-chat/src/chat/services/haa/HumanAgentServiceImpl.ts`: `setFileUploadStatus` (add the failure announcement alongside the existing success one), and the file-selection / send paths for added / start.
- Validation announcements belong wherever size/type/maxFiles checks live — or add pre-flight validation if none exists (the unused strings imply it's missing).

### Notes
- Confirm the upload UI is covered in both assistant and human-agent modes.

ethanwinters:
Closes #1609

Make the file-upload flow announce to screen readers, and validate a selection client-side before upload.

Orientation (this PR targets `feat/prompt-line`, but reviewing it needs no knowledge of that branch): users attach files to a message on two surfaces — the **input area** (chips above the composer, owned by the shared `file-uploads` web component) and the **human-agent** path (success/failure decided after send, in the app's service layer). An upload runs staged → uploading → settles (success/failure), and can be removed before send. None of those transitions were announced; invalid selections (wrong type / too large / too many) failed silently or only via the native picker, which drag-drop and multi-select bypass.

The shared `AriaAnnouncerManager` gains an assertive channel; each surface announces its own transitions (component owns the UI → component announces; state lives in the app → `@carbon/ai-chat` announces — see [references/accessibility.md](references/accessibility.md)).

Complex changes worth a closer look:

- [`aria-announcer-manager.ts`](packages/ai-chat-components/src/globals/utils/aria-announcer-manager.ts) — splits the single region/queue into independent `polite` + `assertive` channels; each keeps its own rotation, pending queue, and 250 ms coalescing timer so an assertive message never merges with a polite one.
- [`file-uploads.ts`](packages/ai-chat-components/src/components/file-uploads/src/file-uploads.ts) — diffs `uploads` across frames in `updated()`. Success is inferred from the `uploading → not-uploading` transition (the list never carries a positive "complete"). Removals announce from the event handler, not the diff (the diff also fires when uploads clear on send). Adds/uploads coalesce into one counted announcement per frame. Element stays mounted when empty so the last file's removal still announces.
- [`AriaAnnouncerComponent.tsx`](packages/ai-chat/src/chat/components/aria/AriaAnnouncerComponent.tsx) — adds two `aria-live="assertive"` regions plus a separate microtask queue so assertive strings never coalesce with the polite node queue.
- [`fileUploadValidation.ts`](packages/ai-chat/src/chat/utils/fileUploadValidation.ts) — pure type/size/count gate returning accepted files + ready-to-announce rejections; catches drag-drop and multi-select that native `accept` can't.

#### Changelog

**New**

- `assertive` flag on `AnnounceMessage` / `useAriaAnnouncer()` — routes blocking errors to a dedicated `aria-live="assertive"` region.
- `maxFileSizeBytes` / `maxFiles` on the human-agent `FileUploadCapabilities` — over-limit files rejected before upload with an assertive reason. Mirrors the existing `UploadConfig` fields on the assistant path.
- Lifecycle announcements: added, uploading, success, failure, removed — added/uploading counts coalesce per frame, locale-pluralized via intl.
- Client-side selection validation (type / size / count) before the upload pipeline, covering drag-drop and multi-select.
- Language strings: `fileSharing_ariaAnnounceFilesAdded`, `fileSharing_ariaAnnounceFilesUploading`, `fileSharing_ariaAnnounceFileRemoved`, `fileSharing_unsupportedType`, `fileSharing_tooManyFiles`.

**Changed**

- `AriaAnnouncerManager.connect(politeRegions, assertiveRegions?)` and `announce(text, "polite" | "assertive")` — assertive falls back to polite when no assertive regions are connected.
- `file-uploads` stays mounted (hidden live regions only) when empty and reflects a `has-uploads` attribute; the input slot takes layout/spacing only when populated.
- File selection resets the input value early, so re-selecting the same file re-fires `change` even when the prior selection was fully rejected.
- Human-agent upload button also disables once `maxFiles` is reached.

**Removed**

- Unused `MENU_SIZE` re-export from [`Menu.tsx`](packages/ai-chat/src/chat/components/carbon/Menu.tsx).

#### Testing / Reviewing

Input-area chips (assistant path) — run the `input-file-upload` example:

- React: `npm run start --workspace=@carbon/ai-chat-examples-react-input-file-upload`
- Web components: `npm run start --workspace=@carbon/ai-chat-examples-web-components-input-file-upload`

With a screen reader: attach file(s) → "File added." / "N files added."; send → uploading then success; remove a chip → "File removed." For rejections, add `allowedFileUploadTypes`, `upload.maxFiles`, and/or `upload.maxFileSizeBytes` to the example's `upload` config, then select a wrong-type / over-count / too-large file → assertive rejection, file not staged.

Human-agent post-send (app-owned) — send `human agent` to connect to the mock service desk, then attach a file. Mock allows only `image/*,.txt`; an unsupported file announces assertively, and resolve/failure announces success/failure.

Unit:

- `npm run test:web-components --workspace=@carbon/ai-chat-components` — covers `file-uploads` and `aria-announcer-manager`.
- From `packages/ai-chat/`: `npx jest tests/utils/spec/fileUploadValidation_spec.ts`.


packages/ai-chat-components/src/components/file-uploads/src/file-uploads.ts
```diff
@@ -42,11 +63,147 @@ class FileUploadsElement extends LitElement {
...
+  @property({ type: String, attribute: "file-removed-label" })
+  fileRemovedLabel = "File removed.";
+
+  /** Announced when a file finishes uploading successfully. */
+  @property({ type: String, attribute: "upload-success-label" })
+  uploadSuccessLabel = "The file was uploaded successfully.";
+
+  /** Announced when a file fails to upload. */
+  @property({ type: String, attribute: "upload-failure-label" })
+  uploadFailureLabel = "There was an error uploading the file.";
```
davidmenendez, line 80: minor, but these strings should be moved into the storybook story and we should leave them empty by default at the component label

davidmenendez (changes requested):

davidmenendez:
everything else looked good. i'm going to make a note that we should revisit this though. i'm not familiar with the nuances of this experience from a screenreader point of view and i'd like to get some more information for future enhancements.

for example; when i set the limit to 2 but uploaded 10 files I would have expected VO to say something like `2 out of 10 uploaded` and not just `file uploaded, file uploaded...` not something i think needs to be handled right away, but i think it's worth investigating best practices here 👍

packages/ai-chat-components/src/components/file-uploads/src/file-uploads.ts
ethanwinters: oh jeez that isn't what we are doing anywhere else. Maybe this is a re-work in 2.0?

PR #1665, opened 2026-06-29
fix: lost max height of input in non-tiptap mode

ethanwinters:
Restore the prompt line's max-height + scrollbar in lite (non-Tiptap) mode.

#1604 moved the 157px height cap off the shared editor slot and re-added it only to the rich Tiptap content node. The prompt line boots in textarea ("lite") mode and upgrades to rich on demand, so in its default state the field grew unbounded — no cap, no scrollbar.

- [`prompt-line-controller.ts`](packages/ai-chat-components/src/components/input/src/prompt-line-controller.ts) — field becomes its own scroll container (`overflow-y: auto` + `max-block-size`) so the browser tracks the caret natively; grow wrapper bounds the box and clips the oversized mirror.
- New [`prompt-line-constants.ts`](packages/ai-chat-components/src/components/input/src/prompt-line-constants.ts) — single source for the 157px cap, shared by both surfaces so they can't drift (the mismatch that caused this).

#### Changelog

**Changed**

- Lite-mode prompt line caps at 157px and shows a vertical scrollbar past it, matching rich mode.
- Rich PM content reads the shared height constant instead of a hardcoded literal.

#### Testing / Reviewing

- `npm run test:web-components --workspace=@carbon/ai-chat-components` — new `getComputedStyle` assertions cover the cap in both textarea and rich mode.
- Manual: open the input, type ~15 lines **before** triggering rich mode; field stops growing at 157px, scrolls, caret stays in view. Trigger rich mode and confirm the cap is identical.

### DIFF
--- a/packages/ai-chat-components/src/components/input/src/__tests__/prompt-line.test.ts
+++ b/packages/ai-chat-components/src/components/input/src/__tests__/prompt-line.test.ts
@@ -11,6 +11,7 @@ import { expect, fixture, html, oneEvent } from "@open-wc/testing";
 import { Extension } from "@tiptap/core";
 
 import "../prompt-line.js";
+import { PROMPT_LINE_MAX_BLOCK_SIZE } from "../prompt-line-constants.js";
 import type PromptLineElement from "../prompt-line.js";
 
 async function makePromptLine(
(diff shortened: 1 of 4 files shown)


warrenmblood (approved):
LGTM. Thanks!

PR #1675, opened 2026-06-30
fix: table rounded corners bug

### LINKED ISSUE #1519: [Bug]: Table inside a Card does not respect the card's rounded corners
### Browser

Chrome

### Package version

v1.4.0 (as shown in the published Storybook)

### Description

When a `Table` is rendered inside a `Card`, the table's bottom corners do not follow the card's `border-radius`. The card has rounded corners, but the inner table content (the last table row, and the pagination footer when present) renders with square corners that visibly poke past the card's rounded corners. The table content appears not to be clipped to the card's rounded boundary.

This is visible in both the **Default** and **With Pagination** Table stories:

- Default: https://chat.carbondesignsystem.com/components/storybook/react/tag/latest/index.html?path=/story/components-table--default
- With Pagination: https://chat.carbondesignsystem.com/components/storybook/react/tag/latest/index.html?path=/story/components-table--with-pagination

**Default story** — the last row (`Alex Kim`) has square bottom-left and bottom-right corners that extend past the card's rounded corners.

**With Pagination story** — the pagination footer (`Items per page` / item count / page controls) has square bottom corners that extend past the card's rounded corners.

**Expected:** the bottom corners of the table / pagination footer should be clipped to (or share) the card's `border-radius`, so the table sits cleanly within the rounded card.

**Actual:** square corners on the table/footer protrude beyond the card's rounded corners.

Likely cause: the card needs `overflow: hidden` (or the inner table container needs matching bottom `border-radius`) so the last row / pagination footer is clipped to the rounded boundary.

### Reproduction / example

https://chat.carbondesignsystem.com/components/storybook/react/tag/latest/index.html?path=/story/components-table--default

### Steps to reproduce

1. Open the Table **Default** story (link above).
2. Look at the bottom-left and bottom-right corners of the table where it meets the card edge.
3. Note the table's last row has square corners that extend past the card's rounded corners.
4. Open the Table **With Pagination** story and observe the same issue on the pagination footer's bottom corners.

davidmenendez:
Closes #1519 

the table component wasn't properly rounding corners when it should have been. it wasn't just enough to add `overflow: hidden` because the mixin wasn't working in the lit version. looks like the selector just needed to be updated. i went initially with just putting it on the host element, but the container will still work if refactored outside of the host selector like this;

```
:host { ... }
.cds-ai-chat-table-container {
  overflow: hidden;
  @include rounded-self;
}
```

#### Changelog

**Changed**

- `packages/ai-chat-components/src/components/table/src/table.scss` fixed css to properly round table corners

#### Testing / Reviewing

run the storybook and review the corners in the table story

### DIFF
--- a/packages/ai-chat-components/src/components/table/src/table.scss
+++ b/packages/ai-chat-components/src/components/table/src/table.scss
@@ -1,5 +1,5 @@
 /*
- *  Copyright IBM Corp. 2025
+ *  Copyright IBM Corp. 2025, 2026
  *
  *  This source code is licensed under the Apache-2.0 license found in the
  *  LICENSE file in the root directory of this source tree.
@@ -19,9 +19,9 @@
 // Apply rounded corners to the container when data-rounded is set
 :host([data-rounded]),
 :host([data-rounded=""]) {
-  .cds-ai-chat-table-container {
-    @include rounded-self;
-  }
+  overflow: hidden;
+
+  @include rounded-self;
 }
 
 // Carbon link styles for links in table cells


warrenmblood (approved):
LGTM!

Issue #1681, opened 2026-07-01
Implement service teardown and correct the disposal guidance


ethanwinters:
### Background

Nothing tears a chat down. When a host removes the chat, its store subscriptions keep running. The theme watcher keeps watching. In-flight requests keep going, and the human-agent service keeps its connection. Every removed chat leaves that behind, so a host that mounts and removes the chat repeatedly piles up a set per cycle.

Two contributor guides already say otherwise. Both tell contributors to dispose services in `ChatInstanceImpl.destroy()` and a matching `unloadServices()`. Neither exists. There is no `unloadServices` in the source at all, and the instance offers only `destroySession()`, which resets saved state rather than disposing anything. So the guides describe an API nobody built, and warn about the exact leak they leave open.

Parent epic: #1728. Two records decide how this resolves, and they rule out the obvious fix. See the contract below.

### Goal

A disposed chat leaves nothing running, and the contributor guides describe teardown as it really works.

### Acceptance criteria

- [ ] Unloading a manager leaves none of the store subscriptions from `loadServices.ts` live.
- [ ] Unloading a manager leaves no event-bus handler registered.
- [ ] Unloading a manager stops the theme watcher and its `MutationObserver`.
- [ ] Unloading a manager releases the human-agent service and its connection.
- [ ] Unloading a manager with a pending `customSendMessage` leaves no live abort controller.
- [ ] Teardown is exhaustive: the spec boots a manager, records every subscription and timer it registers, unloads it, and asserts none survives.
- [ ] Calling teardown twice on one manager does not throw and has no second effect.
- [ ] `packages/ai-chat/AGENTS.md` and `packages/ai-chat/references/services.md` describe the teardown that exists. Neither names a public `instance.destroy()`.
- [ ] No `*.md` under `packages/ai-chat` names `unloadServices()` or `ChatInstanceImpl.destroy()` unless that symbol exists in `src`.
- [ ] The gate for `packages/ai-chat` passes, per the repository's definition-of-done commands.

Teardown behavior is proved by `tests/services/spec/unloadServices_spec.ts`.

### Public API / contract

**No public surface changes.** The public `instance.destroy()` the guides describe is not what gets built. That was the obvious resolution, and two records reject it for separate reasons.

[...]

Issue #1669, opened 2026-06-29
[Feature Request]: Provide an official public API to get current chat history items


rtruong:
### The problem

It is possible to set the history items into the chat using `insertHistory()`, but there is no way to get the current history items.

From @ethanwinters:
> we have had a couple informal conversations for "getMessages" and/or "getMessageById" or including the current message list in "getState" but no concrete issue filed. If you file one, its something we can add in the next couple of releases.

_(source: https://apptio.slack.com/archives/C07RV4MPJNL/p1782755865464579?thread_ts=1782755040.648179&cid=C07RV4MPJNL)_

### The solution

Provide an official public API to get current chat history items

### Is this a blocker to using AI Chat

No

### Business priority

Low Priority = release date is not dependent on fix or not upcoming

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)

---

## Implementation notes (epic #1728)

Setting out the contract before implementation starts, because this is new public surface.

### Goal

A host can read the conversation and be told when it changes, without reaching into anything internal.

### Acceptance criteria

- [ ] `instance.state.messages.get()` returns the conversation's turns, oldest first. The array and its messages are frozen.
- [ ] The array includes a partly streamed response while it is still arriving.
- [ ] `instance.state.status.get()` reports `ready`, `submitted`, `streaming`, or `error`, per the rules below. One spec per value.
- [ ] `instance.state.error.get()` returns the current blocking error, or `null` when there is none.
- [ ] Whenever `status` reads `error`, `error` is not null, and whenever `status` reads anything else, `error` is null.
- [ ] Each store notifies only when its own value changes. A spec asserts that a status change alone does not wake a `messages` subscriber.
- [ ] `get()` returns the same reference until that store next notifies.
- [ ] Nothing here reaches browser storage. A spec drives a full conversation and asserts neither store puts message text in session or local storage.
- [ ] Right after boot, before any message, `instance.state.messages.get()` is empty rather than undefined — the read #1733 points hosts at instead of an adopted flag.
- [ ] `npm run build --workspace=@carbon/ai-chat` exits 0. These are new public types, so TypeDoc link validation runs.
- [ ] The gate for `packages/ai-chat` passes, per the repository's definition-of-done commands.

[...]

Issue #1650, opened 2026-06-25
[Bug]: InputConfig.isDisabled breaks the change on  React 19


shidunce:
### Package version

1.17.0

### React version (if applicable)

19

### Description

I  tried  to use  your new way for disabling/enabling input.
I am on  1.7.0 version now.
What I got is totally broken  chat.
Right after attempt to disable  input the Chat aborts request with "Conversation restarted" reason and write in chat "Request cancelled".
Besides  that after the  attempt the  Chat clears everything from sessionStorage (which I use on  my own bad),  so it breaks my functionality.
After  attempt  to  enable the input  is  kind of activated - I can enter  the  text, but can not send it.
Hide window button is broken.
Blue bottom boundary of chat disappears too.

https://github.com/user-attachments/assets/4b5bc405-2c83-426a-bb39-443cb0a64121

### Reproduction / example

-

### Steps to reproduce

-

### Suggested severity

Severity 4 = Unrelated to a user task, has a workaround or does not need a workaround.

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/carbon-ai-chat/issues) for duplicate problems

ethanwinters:
Related to the persist/SDK decomposition epic #1728 — specifically **A4 #1681** (`destroy()` / `unloadServices()` teardown and the `disposed` guards). Those guards stop a disposed instance from continuing to dispatch, reconnect, or clear session storage mid-flight, which is the disposal side of the symptom reported here. Linking for visibility — this bug is tracked as *related*, not closed, by that work; the React 19 `InputConfig.isDisabled` behavior is verified separately.

PR #1697, opened 2026-07-08
feat: persist api

### LINKED ISSUE #1693: SDK foundations (1.x): reuse-instance groundwork, slot-state stores, ChatSDK facade, and boundary enforcement
## Background

`@carbon/ai-chat` 2.0 will want the option to package into a framework-agnostic core, a thin React wrapper, and a standalone headless SDK export. Today the service/lifecycle layer (`ServiceManager`, the boot sequence, slot-state wiring for user-defined responses and custom footers) is entangled with React — setters typed `React.Dispatch<...>`, repointed on every mount — even though the underlying store and services are already framework-agnostic plain TS.

This issue tracks stage 1 of that separation: internal-only, non-breaking restructuring so the non-view scaffolding no longer depends on React, laying the groundwork for the eventual core/wrapper/SDK split without changing any public API.

- [ ] **Reuse-instance groundwork** — `src/chat/services/reuseInstanceRegistry.ts`: a ref-counted registry (namespace → `ServiceManager`) with grace-timer disposal, so a remounted host can adopt its existing instance instead of tearing down and rebuilding. Paired with real teardown: `unloadServices()` (disposes store subscriptions, event-bus listeners, the human-agent service, etc.) and a public `instance.destroy()` that routes to it. This implements the disposal API that #1681 found documented in `AGENTS.md`/`references/services.md` but missing from the code.
- [ ] **Framework-agnostic slot-state stores** — `src/chat/sdk/valueStore.ts` (a minimal `useSyncExternalStore`-compatible value store, zero imports) and `src/chat/sdk/slotStates.ts` (`attachSlotStateTracking`, reducing the `USER_DEFINED_RESPONSE`/`CHUNK_USER_DEFINED_RESPONSE`/`CUSTOM_FOOTER_SLOT`/`RESTART_CONVERSATION` events into per-slot state). Replaces the old pattern of repointing `React.Dispatch` setters on every mount — a remounting subscriber now gets the current value for free.
- [ ] **React-free shared utils** — `domUtils.ts`/`miscUtils.ts` trimmed of React-only types, `carbonIcon.ts` (genuinely React) relocated to a new `utils-react/` directory, `src/chat/shared/viewHandles.ts` extracted for view-handle interfaces. The old `chatBoot.ts` (321 lines) was retired, its logic folded into the new `ChatSDK.ts`.
- [ ] **Internal `ChatSDK` facade** — `src/chat/sdk/ChatSDK.ts`: `acquireChatSDK()` (create-or-adopt against the reuse registry) and a `ChatSDK` class (`instance`, `slotStates`, `attach()`, `release()`, `destroy()`) as the lifecycle surface `ChatAppEntry` now consumes, in place of the old ad-hoc boot/cleanup code.
- [ ] **SDK boundary enforcement** — `src/chat/sdk/index.ts` (a barrel exporting the above; not yet wired into `package.json#exports` or the rollup build — that's 2.0 packaging work), an eslint `no-restricted-imports` fence added to root `package.json`'s `eslintConfig.overrides`, scoped to `packages/ai-chat/src/chat/{services,store,events,instance,schema,shared,sdk}/**`, banning `react`/`react-dom`/`lit`/`@lit/react`/view-directory imports, and `tests/sdk/spec/sdkBoundary_spec.ts` — a transitive import-graph walker from the barrel that fails on any runtime react/lit import or view-directory import reachable from the SDK core, with an explicit allowlist for the handful of type-only React references in public types. Mechanically proves "the core is react/lit-free."
- [x] **WC container adopts the slot-state stores** — `packages/ai-chat/src/web-components/cds-aichat-container/index.ts` now subscribes directly to `slotStates.userDefinedBySlot`/`.customFooterBySlot` instead of maintaining its own duplicate event→slot reduction. `tests/web-components/spec/slotStateAdoption_spec.ts` covers streaming-chunk rendering, custom-footer rendering, clearing on restart, reseeding on a reuse remount (no new event needed), and unsubscribing on disconnect.

Public API surface (`src/aiChatEntry.tsx`, `src/serverEntry.ts`, `src/types/**`) is untouched by this work.

## Not in this issue

- Shipping the `/sdk` or `/react` subpath exports, or any `package.json#exports`/rollup change — that's 2.0 packaging work, planned separately.
- A "SDK boundary" note in `packages/ai-chat/references/architecture.md` — that file currently lives on the `#731` feature branch, not here; see the follow-up sub-issue.

Related: #1681 (the disposal-API doc drift this closes the gap on — left open for its own PR/issue to close explicitly).

ethanwinters:
Closes #1693 #1681 #1669

Contributes to #1650

We are going to re-work things in 2.0 to allow us to have the SDK to be *potentially* standalone exportable for folks that want to bring more of their own UI. We may never release a standalone SDK, but we wanted to at least make it React free. 

This is the first slice of the 2.0 SDK work: it extracts boot/teardown out of `chat/utils/chatBoot.ts` into a framework-agnostic `chat/sdk/` module (`ChatSDK`, `sdk/index.ts`), and builds three new capabilities on top of that boundary — an opt-in `featureFlags.reuseInstance` that keeps a live `ServiceManager` alive across a host unmount/remount (this initial bug that caused looking into this), a synchronous `getMessagesState`/`getMessage` API backed by a memoized value-store reduction of the app store, and an `onAttach` lifecycle hook that fires on every mount (cold boot or reuse re-attach) so both the React and web-component surfaces can hand out the same `ChatInstance` across a remount. `instance.destroy()` is also added as the hard-teardown counterpart to `destroySession`.

- [`packages/ai-chat/src/chat/sdk/ChatSDK.ts`](packages/ai-chat/src/chat/sdk/ChatSDK.ts) — new facade that subsumes the old `initServiceManagerAndInstance`; `acquireChatSDK` is create-or-adopt (checks the reuse registry first) and `ChatSDK.release()`/`destroy()` decide between a grace-window release and immediate eviction. Review the accidental-remount warning (`maybeWarnAccidentalReboot`) carefully — it's suppressed both by a StrictMode-double-boot time window and by `reuseInstance` being on, and those two conditions interact.
- [`packages/ai-chat/src/chat/sdk/messagesState.ts`](packages/ai-chat/src/chat/sdk/messagesState.ts) — per-turn memoized reduction from `AppState` to `PublicMessage[]`/status/error. The cache logic (`derivePublicMessage`) has a two-tier fast path (reference-equal, then deep-equal-but-rebound) specifically to preserve `===` identity for unrelated turns across a `CHANGE_STATE` deep-merge; worth a close read against `feedback_reducer_reference_stability`.
- [`packages/ai-chat/src/chat/services/reuseInstanceRegistry.ts`](packages/ai-chat/src/chat/services/reuseInstanceRegistry.ts) — framework-agnostic, ref-counted registry (a bare module `Map`, deliberately outside React/Lit) that holds a `ServiceManager` alive for `reuseInstanceGraceMs` after the last mount releases it.
- [`packages/ai-chat/src/chat/ChatAppEntry.tsx`](packages/ai-chat/src/chat/ChatAppEntry.tsx) — boot path now goes through `acquireChatSDK`/`adopted` instead of always cold-booting; `onBeforeRender`/`onAfterRender`/the initial view-change transition are all gated on `!adopted` (boot-once), while the new `onAttach` and the internal `onSlotStatesReady` handshake fire unconditionally on every attach. Slot-state portals moved from local `useState` to `useSyncExternalStore` over core-owned stores so a reuse re-attach shows already-accumulated content with no setter repointing.
- [`packages/ai-chat/src/web-components/cds-aichat-container/index.ts`](packages/ai-chat/src/web-components/cds-aichat-container/index.ts) — mirrors the same slot-state subscription change on the Lit side, replacing bespoke bus-event accumulation (`enhancedUserDefinedHandler` et al.) with a subscription to the shared `ChatSlotStates` stores — this is the proof the mechanism is genuinely framework-agnostic.

#### Changelog

**New**

- `featureFlags.reuseInstance` (+ `reuseInstanceGraceMs`): opt in to keep the live chat instance alive in-memory across a host unmount/remount (React StrictMode, a changing `key`, conditional rendering) instead of cold-booting and losing the conversation. Slated to become the default behavior in 2.0.
- `instance.messaging.getMessagesState()` / `getMessage(messageId)`: synchronous snapshot of the conversation's messages/status/error, stripped of internal bookkeeping fields, backed by `BusEventType.MESSAGES_STATE_CHANGE`.
- `onAttach(instance, { remount })`: new lifecycle hook (React prop and web-component attribute) that fires on every mount, letting a host distinguish first boot from a reuse re-attach.
- `instance.destroy()`: hard teardown that disposes the instance and all its services immediately, bypassing any reuse grace window.
- Dev-only warning when a namespace cold-boots twice in a row without `reuseInstance` enabled, flagging an accidental unmount/remount that discarded the conversation.

**Changed**

- Boot/teardown internals moved from `chat/utils/chatBoot.ts` into a new `chat/sdk/` module (`ChatSDK`, `sdk/index.ts`) that forms the root of the future `@carbon/ai-chat/sdk` import boundary (enforced by a new `sdkBoundary_spec.ts`).
- User-defined-response and custom-footer slot state moved off local component `useState`/bus-event accumulation onto framework-agnostic value stores owned by the `ServiceManager`, consumed identically by both the React tree (`useSyncExternalStore`) and the Lit web component.

#### Testing / Reviewing

Not reachable from the demo site (`reuseInstance`, `getMessagesState`, and `onAttach` have no demo query-param toggle or UI in this PR) — verify via the new unit suites instead:

```bash
npm run test --workspace=@carbon/ai-chat
```

Focus areas:
- `tests/config/spec/reuseInstanceAdoption_spec.tsx` — remount reuses the same `ServiceManager`/`ChatInstance` within the grace window; a remount past the window cold-boots.
- `tests/config/spec/rebootWarning_spec.tsx` — the accidental-remount dev warning fires/doesn't fire in the right conditions (StrictMode window, `reuseInstance` on/off).
- `tests/sdk/spec/chatSDK_spec.ts`, `tests/sdk/spec/sdkBoundary_spec.ts` — `ChatSDK` lifecycle (`acquire`/`attach`/`release`/`destroy`) and the framework-agnostic import boundary.
- `tests/sdk/spec/messagesState_spec.ts`, `tests/instance/spec/messaging/getMessagesState_spec.ts` — `getMessagesState`/`getMessage` correctness, including the reference-stability guarantees across unrelated state updates.
- `tests/web-components/spec/slotStateAdoption_spec.ts` — the Lit container subscribing to the shared slot-state stores instead of its own bus handlers.

Also run `npm run build --workspace=@carbon/ai-chat` since this touches boot sequencing on both the React and web-component surfaces.


ethanwinters:
This branch is getting replaced by the sdk-persistence-integration branch to avoid a giant single PR

PR #1703, opened 2026-07-08
fix: Plex override example

davidmenendez:
Adds an example that demonstrates how to override the Plex font in Carbon AI Chat.

The only change necessary to the codebase is in `AppShell.scss`

```
:host {
  font-family: type.font-family("sans");
}
```

Previously it would have been impossible to override the font for consumers because Plex was built into the application. By making a simple change we can make it much more flexible.

```
:host {
  font-family: inherit;
}
```

Now, using a [solution that Taylor came up with](https://github.com/carbon-design-system/carbon/pull/22119#pullrequestreview-4604270759), consumers can import their own fonts and have the components use that instead. Here's the snippet from the example stylesheet

```
@use "sass:string";

@use "@carbon/styles" with (
  // Turn off Carbon's Plex `@font-face` generation
  $css--font-face: false,

  // Change the root `font-family` definition(s)
  $font-families: (
      "sans": string.unquote(
          "'Permanent Marker', system-ui, -apple-system, BlinkMacSystemFont, '.SFNSText-Regular', sans-serif"
        ),
      "mono": string.unquote(
          "'Permanent Marker', system-ui, -apple-system, BlinkMacSystemFont, '.SFNSText-Regular', monospace"
        ),
      "serif": string.unquote(
          "'Permanent Marker', system-ui, -apple-system, BlinkMacSystemFont, '.SFNSText-Regular', serif"
        ),
      // Include any other Carbon font family keys your application compiles.
    )
);

@font-face {
  font-display: swap;
  font-family: "Permanent Marker";
  font-style: normal;
  font-weight: 400;
  src: url("./permanent-marker-v16-latin-regular.woff2") format("woff2");
}
```

In this example; any Plex fonts that are using sans, mono, or serif should be serving up the `Permanent Marker` font.

Before:

Screenshot 2026-07-08 at 3 37 46 PM

This before demonstrates how even with that font override that the declaration isn't reaching the application because it's got Plex baked into it.

After:

Screenshot 2026-07-08 at 3 38 03 PM

With `font-family: inherit` the new font will be able to reach the application. The only caveat is that consumers will need to [provide plex themselves similar to the way we are doing in the demo application](https://github.com/carbon-design-system/carbon-ai-chat/blob/main/demo/src/react/DemoApp.tsx#L11), but that should be a very minor thing since most of our consumers will most certainly already be importing and using Plex somewhere. This change will enable the most amount of flexibility without making any changes to the core build system.


examples/web-components/theme-plex-override/src/customSendMessage.ts
```diff
@@ -0,0 +1,289 @@
+/*
+ *  Copyright IBM Corp. 2025, 2026
```
ethanwinters, line 2: can you rename example to theme-plex-override?

And can you just have Bob make a react version of this, too?

ethanwinters (approved):
Approved with comments.

I'm going to have Bob do an audit of what `@carbon/web-components` force their own fonts so we can get an idea of what is still missing.

examples/web-components/theme-plex-override/src/customSendMessage.ts
davidmenendez: you bet!

Issue #1679, opened 2026-07-01
[Bug]: customSendMessage and customLoadHistory would not be called if event changed in onViewPreChange


nelsonchen90:
### Browser

Chrome

### Package version

v1.15.0

### React version (if applicable)

v18.3.1

### Description

We have requirement to force the chat to open/closed cross pages/services in the product.
We used to have an implementation to force `changeView` in an useEffect during chatInstance initialization. However that has been conflicting with the initial render/view change event from carbon.

So that take us to try to use the approach of using `onViewPreChange` and change the mainWindow/launcher in the viewState. 

However, it seems like once the `event.viewState` is changed (in this case, force the mainWindow to be open), neither `customSendMessage` or `customLoadHistory` is being called after the chat window appear to be open visually

### Reproduction / example

https://stackblitz.com/edit/github-gd1bdmnh?file=src%2FApp.tsx

### Steps to reproduce

- set up such a chat instance following the description
- refresh the page and force the mainwindow to open
- see neither the `customSendMessage` or `customLoadHistory` is called

(recording of a refresh of the page)

https://github.com/user-attachments/assets/f9eed701-b97d-4e7c-92aa-6377a9be1bd4

### Suggested severity

None

### Application / product

watsonx.data intelligence

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/carbon-ai-chat/issues) for duplicate problems

PR #1708, opened 2026-07-10
feat: add consumer-owned session state persistence

ethanwinters:
Contributes to #1679

## Summary

Adds `PublicConfig.persistedState` — an opt-in callback interface letting hosts own where session-state (views, disclaimer acceptance, human-agent connection state) is persisted. By default, `sessionStorage` remains unchanged; providing `initialState` and/or `onStateChange` hands that storage to the host, mirroring the pattern established by `customLoadHistory`. Includes completeness guard fix: `persistedState` is classified in the config fan-in spec as a reactive-backbone field (boot-time read + fresh-read-per-fire, no extra per-field side-effect block).

**Files with particularly complex changes**:
- [`packages/ai-chat/src/chat/store/persistenceUtils.ts`](packages/ai-chat/src/chat/store/persistenceUtils.ts) — translation layer between internal `PersistedState` and public `PersistableState`; review the dual-direction mapping (`fromPersistableState` / `toPersistableState`) to confirm the frame-internal bookkeeping (`version`, `wasLoadedFromBrowser`) is handled correctly.
- [`packages/ai-chat/src/chat/store/subscriptions.ts`](packages/ai-chat/src/chat/store/subscriptions.ts) — `copyToSessionStorage` now branches on whether the host owns persistence; ensure the subscription gate (reference equality on `persistedToBrowserStorage`) still fires the callback at the right moments.

#### Changelog

**New**

- `PublicConfig.persistedState` + `PersistedStateConfig` interface: opt-in host-owned session-state persistence (boot from `initialState`, receive changes via `onStateChange`). When neither field is set, `sessionStorage` behavior is unchanged.
- `PersistableState` type: the public contract for session state passed to/from `onStateChange`. Opaque blob — store verbatim to avoid regressions (dropping fields like `disclaimersAccepted` or `humanAgentState` regresses the UX on reload).
- `persistenceUtils.ts`: `fromPersistableState` / `toPersistableState` helpers ensuring the public and internal state shapes cannot drift.
- Public documentation at `packages/ai-chat/docs/StatePersistence.md` (linked from `CustomHistory.md` and `CustomServer.md`).
- Unit tests: `persistedState_spec.ts` verifies host-owned persistence boot/restore flow; `persistenceUtils_spec.ts` verifies the translation layer.

**Changed**

- `doCreateStore.ts`: boot logic now checks `config.persistedState` first; if a host has opted in (either field set), uses `initialState` instead of `sessionStorage.loadSession()`, and skips the `sessionStorage` write entirely.
- `copyToSessionStorage` subscription: now routes persistence to `onStateChange` callback (if host owns it) instead of `userSessionStorageService.persistSession()`.
- `ChatActionsImpl.destroySession()`: skips `userSessionStorageService.clearSession()` when host owns persistence (the state reset dispatched below flows to `onStateChange` like any other change).
- `FLATTENED_PUBLIC_CONFIG_FIELDS`: added `persistedState` (property-only, no attribute, carries the callback).
- `configFanInCompleteness_spec.ts`: classified `persistedState` as a `REACTIVE_BACKBONE_FIELD` (boot-time read of `initialState` + fresh-read-per-fire of `onStateChange`; no extra per-field side-effect block needed).

#### Testing / Reviewing

This feature is not reachable from the demo site (requires a host to wire up the callback interface). Verify via unit tests:

```bash
cd packages/ai-chat
npx jest tests/config/spec/persistedState_spec.ts tests/store/spec/persistenceUtils_spec.ts
```

Focus areas:
- `persistedState_spec.ts` — verifies boot from `initialState`, observes state changes via `onStateChange`, confirms `sessionStorage` is never touched when host owns persistence.
- `persistenceUtils_spec.ts` — verifies the round-trip translation between `PersistedState` (internal) and `PersistableState` (public), confirms `version` and `wasLoadedFromBrowser` bookkeeping is added/dropped correctly.
- `configFanInCompleteness_spec.ts` — verifies `persistedState` is classified (failing test would indicate an unclassified field).

Also run the completeness spec:
```bash
npx jest tests/config/spec/configFanInCompleteness_spec.ts
```

All existing session-restore tests should remain green (behavior unchanged when `persistedState` is not set).

### DIFF
--- a/packages/ai-chat/docs/CustomHistory.md
+++ b/packages/ai-chat/docs/CustomHistory.md
@@ -195,5 +195,6 @@ instance.on({
 - [chat-history-fullscreen example](https://github.com/carbon-design-system/carbon-ai-chat/tree/main/examples/web-components/chat-history-fullscreen) — a fullscreen web-component layout with a custom history panel.
 - [Slots](./WriteableElements.md) — render your own content into the history panel slot.
 - [Message format](./MessageFormat.md) — the request/response shapes history items wrap.
+- [Session state persistence](./StatePersistence.md) — own where the chat's session and UI state (views, disclaimer, human-agent connection) is stored, separately from conversation messages.
 - [Adding messages (legacy)](./AddMessageChunk.md) — getting live responses onscreen.
 - [Adding messages (experimental)](./UpsertMessage.md) — revising a message after it renders.
--- a/packages/ai-chat/docs/CustomServer.md
+++ b/packages/ai-chat/docs/CustomServer.md
@@ -6,6 +6,7 @@ children:
   - ./UpsertMessage.md
   - ./StructuredData.md
   - ./CustomHistory.md
+  - ./StatePersistence.md
 ---
 
 ## Overview
@@ -22,6 +23,7 @@ This page is the entry point. Depending on what you are building, continue to:
 - [Adding messages (experimental)](./UpsertMessage.md) — the preferred, experimental insert-or-update-by-ID flow.
 - [Structured data](./StructuredData.md) — send typed fields and uploaded files alongside the user's text.
 - [Conversation history](./CustomHistory.md) — loading and restoring previous conversations.
+- [Session state persistence](./StatePersistence.md) — own where the chat's session and UI state is stored.
 
 ## Connecting your server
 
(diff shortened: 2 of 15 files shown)


davidmenendez (approved):
looks great 👍

Issue #1732, opened 2026-07-13
Add the acquire and boot pipeline in src/sdk/, and enforce the SDK boundary


ethanwinters:
### Background

There is no object that owns a chat. Booting one is a side effect of a React component mounting. The mount effect builds the services, the store, and the instance, and returns no cleanup. So the old graph is abandoned rather than torn down. Nothing else can start a chat, and nothing can stop one.

That is what blocks a host on any other framework. It is also why the conversation layer keeps drifting back into the view. With no boot object to hold shared plumbing, plumbing gets attached to whatever component was nearby.

This issue creates the object and fences the boundary so it stays clean.

Parent epic: #1728. ADR-0025 decides the shape of the acquire and where lifecycle lives. ADR-0002 decides that no SDK surface is exported in 1.x, and that this boundary is enforced by a machine rather than by habit.

### Goal

Booting and tearing down a chat runs through one framework-free path that the shells call. A guard fails the build if view code reaches the conversation layer.

### Acceptance criteria

- [ ] `src/sdk/` exists, holds the acquire, and imports only from `src/shared/`.
- [ ] `acquireChatSDK(config)` boots the services, the store, and the instance with no container and no DOM. `tests/sdk/spec/chatSDK_spec.ts` acquires in a bare environment and sends a message.
- [ ] It resolves only once boot is done, so a caller never holds a half-built chat.
- [ ] What it resolves to is one flat object: the conversation surface plus `release()` and `destroy()`. There is no extra hop to reach the instance.
- [ ] `release()` and `destroy()` both route through the teardown from #1681.
- [ ] A second `release()` or `destroy()` on the same handle is a no-op.
- [ ] A service manager can be unloaded with no acquire involved. `chatSDK_spec.ts` boots one directly, unloads it, and asserts teardown ran — proving the manager holds no reference back to whatever booted it.
- [ ] A partial config inherits each default field by field. `tests/boot/spec/appBoot_spec.ts` covers the merge.
- [ ] The React shell and both web-component shells boot through this path.
- [ ] `chat/utils/chatBoot.ts` and `tests/utils/spec/chatBoot_spec.ts` are deleted.
- [ ] The shells reach the acquire through an internal entry point typed for the whole instance, so a shell callback still receives the full instance.
- [ ] The callback-render path in the web components reads the slot stores from #1741 instead of tracking slot state itself. `tests/web-components/spec/slotStateAdoption_spec.ts` covers it.
[...]

Issue #1728, opened 2026-07-13
[Epic]: Persist API — framework-agnostic SDK foundations, instance reuse, and a public conversation API


ethanwinters:
### Value proposition

A host that unmounts and remounts the chat loses the conversation. React StrictMode, a changing `key`, or a conditional render is enough, and the transcript does not come back on its own. When a human agent was connected, the fresh boot reads the stale connection flag and tells the user their conversation with that agent is over. A render the host never thought of as a lifecycle event ends a real conversation.

There is also no supported way to read the conversation back. A host can push history in and cannot get it out. So anything it wants to know about the transcript, it has to track separately and keep in step by hand.

Both come from the same place. Booting a chat is a side effect of a React component mounting. Nothing owns a conversation apart from a mounted component, so there is no object to keep alive across a remount, and none to read from.

This epic builds that object. Three things reach consumers in 1.x:

- **Conversations survive a remount.** Opt-in `featureFlags.reuseInstance` keeps the running chat in memory across an unmount and remount, instead of cold-booting and throwing it away.
- **Read the conversation.** `instance.state.messages`, `.status`, and `.error` expose the transcript, the messaging lifecycle, and the current error. Each is watched on its own. (closes #1669)
- **Stop a turn from code.** `instance.messaging.stop()` stops the active turn, and the built-in stop button is re-pointed at that same path.

Underneath, it moves the conversation layer into its own top-level folder, draws a framework-free boundary around it, and enforces that with a machine. It also gives boot and teardown one home. So the headless SDK becomes a packaging step later, not a rewrite.

### Expected outcomes

- [ ] No React, Lit, or view import reaches the conversation layer. The ESLint fence and the import-graph guard are green, and the guard is proved to bite.
- [ ] `featureFlags.reuseInstance` keeps a conversation alive across a host remount, in-session.
- [ ] `instance.state.messages`, `.status`, and `.error` read and watch the conversation. `get()` returns a stable frozen value, and `subscribe` fires only on change.
- [ ] `instance.messaging.stop()` stops the active turn, and the shipped stop button drives the same path.
- [ ] Boot and teardown run through one framework-free path. A torn-down chat leaves no live subscription, timer, or in-flight request.
- [ ] No SDK runtime surface is exported in 1.x — no entry point, no acquire, not even an experimental export.
- [ ] Nothing in this epic writes conversation text to browser storage.
- [ ] The conversation layer lives in `src/shared/` and imports nothing else from `src/`.

[...]

PR #1742, opened 2026-07-14
feat: add framework-agnostic slot-state value stores

### LINKED ISSUE #1741: Add the framework-agnostic slot-state value stores
### Background

The chat keeps some of its state nowhere in particular. A host can supply its own content for a response or a message footer. To work out which of those slots are live, the chat watches the event bus and builds the answer up inside the view. Every view that needs it builds its own copy, and the answer dies with the view. So a remount starts over, and a host with no view cannot reach it at all.

That is a symptom of a missing piece. The chat has no small, framework-free way to hold a value and let anything watch it change. So state ends up wherever something was already listening.

This issue adds that piece and moves the slot bookkeeping onto it. It lands inert. Nothing calls the tracking yet; #1732 wires it into boot.

Parent epic: #1728. The store's contract comes from ADR-0004, which makes per-field stores the way chat state is read.

### Goal

A framework-free value store exists and behaves exactly as ADR-0004 says. Slot bookkeeping for user-defined responses and custom footers lives in two of them, instead of being rebuilt per view. Three slot events and a conversation restart drive them.

### Acceptance criteria

- [ ] A store's `get` and `subscribe` are bound function properties, so they survive being passed as bare references.
- [ ] `subscribe` never calls its listener at subscribe time: zero calls before any write.
- [ ] `subscribe` returns its own unsubscribe function, and calling it stops delivery to that listener only.
- [ ] A listener receives the new value as its argument.
- [ ] Writing a value equal by `Object.is` notifies nobody.
- [ ] Writing a different reference notifies every listener exactly once.
- [ ] `get()` returns the same reference until the store next notifies, read across a no-op write.
- [ ] Firing a user-defined-response event makes that slot readable from the user-defined store.
- [ ] Firing a chunked user-defined-response event updates the same store for that slot.
- [ ] Firing a custom-footer-slot event makes that slot readable from the footer store.
- [ ] Restarting the conversation empties both stores.
- [ ] Attaching slot tracking twice to one service manager still yields one update per event.
- [ ] The store module imports nothing at all.
- [ ] The slot module imports no React, no Lit, and nothing that registers an element.
- [ ] The two public slot-state types move to their own module and are re-exported from their old home.
- [ ] `npm run build --workspace=@carbon/ai-chat` exits 0, proving no published type changed.
- [ ] Nothing calls slot tracking yet: the chat behaves the same, and no existing spec changes.
- [ ] The gate for `packages/ai-chat` passes, per the repository's definition-of-done commands.

The store criteria are proved by `tests/sdk/spec/valueStore_spec.ts`, the slot criteria by `tests/sdk/spec/slotStates_spec.ts`, which exercises tracking against a fake service manager with no real boot.

### Public API / contract

**Internal only.** The store type and the tracking function are not exported from the package. The two slot-state types move modules but keep their published shape and their old import path, so no consumer sees a change.

The store matches ADR-0004's contract now, rather than being adapted later, because that record makes this shape the public read model:

```ts
interface ChatStore<T> {
  get: () => T;
  subscribe: (listener: (value: T) => void) => () => void;
}
```

**Derivation.** A store notifies when its value changes by `Object.is`. `get()` returns that same reference until the next notification. Values are worked out once per change and cached, never rebuilt per read. Store objects are stable properties, not getters that mint a new store per access. Both halves are what let a host memoize above a store safely, so both are asserted rather than assumed.

**No-op case.** Writing an equal value does nothing. Attaching tracking to a manager that already has it does nothing.

Everything else is unchanged: no new events — the tracking only listens to events that already fire — no failure path, no new defaults, no announcement, no consumer action.

### Out of scope

- Calling the tracking, wiring it into boot, and subscribing the shells to it. That is #1732.
- Putting any store on the public instance surface. The conversation stores are #1669. The view stores and the `getState()` deprecation are #1829, under epic #1799.
- The conversation stores `messages`, `status`, and `error`. #1669 builds those on this primitive.

### Related

- Parent epic #1728.
- Blocks #1732, which consumes both modules, and #1669, which builds its three stores on this store.
- #1829 (epic #1799) puts stores of this shape on the public instance surface. Getting the contract right here is what lets that issue expose this store rather than adapt it.

ethanwinters:
Closes #1741

THIS PR DOESN'T ACTUALLY DO ANYTHING, ITS JUST SPLITTING UP WORK INTO SMALLER PRs, the next PR will wire it up as well as other related parts.

Part of the persist/SDK decomposition epic #1728 — the slot-state store module, split out of #1732 (ChatSDK facade) so it reviews on its own ahead of the ~2,200-line boot refactor. Lands **inert**: nothing calls `attachSlotStateTracking` yet; #1732 wires it into boot and the web-components.

- `sdk/valueStore.ts` — a framework-agnostic, `Object.is`-gated subscribable (`useSyncExternalStore`-compatible). Notifies only its own subscribers, only on a real change; a remounting subscriber reads the current value on first `get()`.
- `sdk/slotStates.ts` — `attachSlotStateTracking` reduces the `USER_DEFINED_RESPONSE` / `CHUNK_USER_DEFINED_RESPONSE` / `CUSTOM_FOOTER_SLOT` / `RESTART_CONVERSATION` bus events into two per-slot value stores hung off the `ServiceManager`; idempotent per manager.
- Relocates `RenderUserDefinedState` / `RenderCustomMessageFooterState` from `ChatContainer.ts` into `types/component/slotStates.ts`, re-exported from `ChatContainer` — public shape unchanged.
- `ServiceManager` gains `slotStates?`.

#### Changelog

**New**

- `sdk/valueStore.ts`, `sdk/slotStates.ts` — framework-agnostic slot-state value stores (not yet wired in).
- `ServiceManager.slotStates` field.

**Changed**

- `RenderUserDefinedState` / `RenderCustomMessageFooterState` relocated to `types/component/slotStates.ts`, re-exported from `ChatContainer` (no public API change).

#### Testing / Reviewing

- `cd packages/ai-chat && npx jest tests/sdk/spec/slotStates_spec.ts` — 9 passing (valueStore get/subscribe/`Object.is` no-op; per-slot reduction; idempotent re-attach — all against a fake manager, no boot).
- `npx tsc --noEmit` clean; `npm run build --workspace=@carbon/ai-chat` clean (TypeDoc validates the relocated public types; 0 errors).
- Inert module — nothing calls it yet, so no runtime behavior change.

---

_Also bundles the `shell-quote@^1.10.0` root override (CVE-2026-13311 DoS, pulled in transitively by `concurrently`) so this branch's security checks pass — the same override is going to `main` standalone in #1743._

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41271,9 +41271,9 @@
       }
     },
     "node_modules/shell-quote": {
-      "version": "1.8.4",
-      "resolved": "https://registry.npmjs.org/shell-quote/-/shell-quote-1.8.4.tgz",
-      "integrity": "sha512-VsC6n6vz1ihYYyZZwX7YZSF5l5x36ca17OC+a69h94YqB7X6XLwf+5MOgynYir2SLFUbl8gIYvBo8K8RoNQ6bQ==",
+      "version": "1.10.0",
+      "resolved": "https://registry.npmjs.org/shell-quote/-/shell-quote-1.10.0.tgz",
+      "integrity": "sha512-w1aiOKwKuRgtwAReIIj89puqg+I7GvX4IbLrvmhXbzQsj1+Zwi4VO3+fa6ZF91TWSjIxoEkKnMeHcLEODK5ZXA==",
       "license": "MIT",
       "engines": {
         "node": ">= 0.4"
--- a/package.json
+++ b/package.json
@@ -236,6 +236,7 @@
     "@carbon/ai-chat": "^0.0.0"
   },
   "overrides": {
-    "tar": "^7.5.10"
+    "tar": "^7.5.10",
+    "shell-quote": "^1.10.0"
   }
 }
--- a/packages/ai-chat/src/chat/sdk/slotStates.ts
+++ b/packages/ai-chat/src/chat/sdk/slotStates.ts
@@ -1,0 +1,135 @@
+/*
+ *  Copyright IBM Corp. 2025, 2026
+ *
+ *  This source code is licensed under the Apache-2.0 license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ *  @license
+ */
+
+import type { ServiceManager } from "../services/ServiceManager";
+import { createValueStore, ReadableValueStore, ValueStore } from "./valueStore";
+import {
+  BusEventChunkUserDefinedResponse,
+  BusEventCustomFooterSlot,
+  BusEventType,
+  BusEventUserDefinedResponse,
+} from "../../types/events/eventBusTypes";
+import {
+  RenderCustomMessageFooterState,
+  RenderUserDefinedState,
+} from "../../types/component/slotStates";
+
+/**
+ * Framework-agnostic slot-projection state for the two portal surfaces (user-defined responses and
+ * custom message footers), keyed by slot. The core owns these value stores and reduces the relevant
+ * bus events into them; the React (and, later, Lit) view layer subscribes for rendering. Because
+ * the stores hang off the {@link ServiceManager}, the accumulated state survives a host remount —
+ * a remounting subscriber reads the current value on first `get()`, no setter repointing required.
+ */
+export interface ChatSlotStates {
+  /** User-defined-response state by slot, reduced from the USER_DEFINED_RESPONSE event family. */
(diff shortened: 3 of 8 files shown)


davidmenendez (approved):
LGTM

PR #1748, opened 2026-07-15
docs: apply tone guide to Launcher.md

ethanwinters:
Closes #

Tone pass on `Launcher.md` (Launcher) per `references/tone.md`. Prose only — code blocks and page structure unchanged.

#### Changelog

**Changed**

- Reword `Launcher.md` for voice and reading level.

#### Testing / Reviewing

- Flesch-Kincaid grade: 8.6 (target 8–11). Measured with `npm run reading-level`, which ships in #1740.
- Confirm the reword kept every fact, default, and qualifier.

### DIFF
--- a/packages/ai-chat/docs/Launcher.md
+++ b/packages/ai-chat/docs/Launcher.md
@@ -6,11 +6,11 @@ title: Launcher
 
 The launcher is the corner button that welcomes users and opens the chat. Configure it through the {@link PublicConfig.launcher | launcher config}, supply your own button in its place, or leave it out of the layout entirely.
 
-The launcher is optional, so you can leave it out completely: set `launcher.isOn: false` and {@link PublicConfig.openChatByDefault | open the chat directly} instead. That combination is the common setup for a custom-element or embedded layout that carries no launcher at all.
+The launcher is optional. Turn it off with `launcher.isOn: false` and open the chat directly with {@link PublicConfig.openChatByDefault | openChatByDefault} — the common setup for embedded layouts with no launcher.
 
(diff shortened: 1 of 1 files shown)


AlexanderMelox (commented):
### Important

- `Launcher.md:~10` — **Over-simplification (meaning drift)**
  **Before:** `"A launcher is optional. Turn it off with \`launcher.isOn: false\` and open the chat directly with {@link PublicConfig.openChatByDefault} — the common setup for a custom-element or embedded layout that has no launcher at all."`
  **After:** `"The launcher is optional, so you can leave it out completely: set \`launcher.isOn: false\` and {@link PublicConfig.openChatByDefault | open the chat directly} instead. That combination is the common setup for a custom-element or embedded layout that carries no launcher at all."`
  **Fix:** `"leave it out completely"` and `"carries no launcher at all"` are redundant with each other and with `"isOn: false"`. The added phrase `"That combination is…"` is also wordier than the original dash clause. Suggested: `"The launcher is optional. Turn it off with \`launcher.isOn: false\` and open the chat directly with {@link PublicConfig.openChatByDefault} — the common setup for embedded layouts with no launcher."`

### Nits

- `Launcher.md:~14` — **Sentence length** — The rewrite of the "Configure the launcher" paragraph is accurate but runs noticeably longer than the original. The added explanation of `avatarUrlOverride` (`"Each of those device configs accepts an avatarUrlOverride…"`) doesn't appear in the original and is a scope addition. Worth flagging for the author.

AlexanderMelox (approved):

Issue #1608, opened 2026-06-16
Screen reader announcements for suggestion results


ethanwinters:
Part of #731. Blocked by #1610 (autocomplete focus model).

### Problem
The autocomplete / prompt-line suggestion list has no screen-reader announcements when suggestions load, change, or come back empty (WCAG 4.1.3 Status Messages). The presentational `cds-aichat-autocomplete` renders nothing when empty, so today there's no surface that could even announce "no results."

### Requirements
- A **polite** live region announces: results available (e.g. "5 suggestions available"), no results, and — async only — loading in progress. Selection/insertion is optional.
- Announcement strings are **injected and localizable**, not hardcoded in `@carbon/ai-chat-components`. The results-count string must be a function/template (`(count) => string`) since pluralization is locale-dependent and `@carbon/ai-chat` owns the formatter; the rest can be plain strings.
- **Active-option announcement is conditional on the focus model (#1610).** Today the component moves focus onto the option buttons, so the active option is announced via focus and needs **no** live region — do not double-announce it. If #1610 adopts the `aria-activedescendant` model (input keeps focus), focus no longer moves, and announcing the highlighted option during arrow navigation becomes part of this work.

### Where / how
- Own it in the **controller** (`AutocompleteControllerElement` in `packages/ai-chat-components/src/components/input/src/autocomplete-controller.ts`), not the presentational `cds-aichat-autocomplete` — the controller knows the load/empty/async transitions; the component renders `null` when empty.
- Reuse **`AriaAnnouncerManager`** (`packages/ai-chat-components/src/globals/utils/aria-announcer-manager.ts`); copy the pattern shipping in `chat-shell` (`.../chat-shell/src/shell.ts`): hidden live regions in `render()`, `connect()` in `firstUpdated`, `disconnect()` in `disconnectedCallback`, `announce()` at the trigger points.
- Hook points: `_kickoffResolve` (loading), after `_resolveItems` resolves (count / no-results), `select()` (optional). Inject strings as element properties (mirror the `*Announcement` props on `chat-shell`'s element).

### Notes / dependencies
- **Blocked by #1610.** The focus-model decision there determines whether active-option announcements during navigation are in scope here, so finalize that issue first.

PR #1775, opened 2026-07-15
feat: additional keyboard support and announcements in autocomplete

### LINKED ISSUE #1610: Adopt combobox focus model for autocomplete
### Problem
The autocomplete uses two conflicting keyboard-focus models:
- The presentational component (`cds-aichat-autocomplete`) moves **real DOM focus** onto each option button on Arrow navigation (`button.focus()` in `packages/ai-chat-components/src/components/autocomplete/src/autocomplete.ts`).
- The controller (`AutocompleteControllerElement`) **forwards Arrow/Enter/Escape from the editor to the list**, which only makes sense if the **editor keeps focus**.

These contradict each other, and neither fully matches the WAI-ARIA combobox pattern for a typeahead inside a text field. 

### Goal — adopt the combobox model (input keeps focus + `aria-activedescendant`)
- The input/editor retains DOM focus throughout; typing is never interrupted by navigating suggestions.
- The active option is indicated with `aria-activedescendant` on the input, pointing at the active option's `id` (options need stable ids).
- Options expose `role="option"` and `aria-selected` on the active one.
- The listbox and options are **not** in the tab order; stop calling `button.focus()` — Arrow keys update `aria-activedescendant` instead.
- Wire `aria-expanded` / `aria-controls` on the input to reflect open/closed state and the listbox id.

### Folds in these existing a11y findings
- Remove host `tabindex="0"` — the container shouldn't be a tab stop.
- Add option ids + `aria-selected` and `aria-activedescendant` wiring.
- `aria-expanded` / `aria-controls` on the input.

### Impact on announcements (#1608)
Under the current "focus moves into the list" model, screen readers announce the active option via focus, so #1608 only needs list-level announcements. Once focus stays in the input, the active option is **no longer announced via focus** — announcing it during navigation becomes part of the announcer's job. That's why this **blocks #1608**: the announcement scope can't be finalized until the focus model is settled.

### Reference
WAI-ARIA APG combobox pattern: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/

davidmenendez:
Closes #1610 & #1608 

adds some additional keyboard support to autocomplete as well as the announcement functionality for accessibility.

#### Changelog

**Changed**

- `packages/ai-chat-components/src/components/autocomplete/src/autocomplete.ts` main source of changes. most importantly; switching to a `listbox` role and adding in the announcement functionality using `AriaAnnouncerManager`

#### Testing / Reviewing

some additional testing has been included thanks to Bob. for manual testing: run the demo app - in the sidebar enable `Enable autocomplete suggestions` - start voiceover using `command + F5` - begin typing - review the dialog that's currently being announced; number of suggestions and currently highlighted suggestion - use arrow keys to move up and down and review the announcement dialog - hit `Enter` and verify the selection has been captured in the prompt-line


davidmenendez:
i'm not sure what happened to the commits. going to close and re-open

PR #1783, opened 2026-07-15
feat: additional keyboard support and announcements in autocomplete

### LINKED ISSUE #1610: Adopt combobox focus model for autocomplete

davidmenendez:
Closes #1610 & #1608 

adds some additional keyboard support to autocomplete as well as the announcement functionality for accessibility.

#### Changelog

**Changed**

- `packages/ai-chat-components/src/components/autocomplete/src/autocomplete.ts` main source of changes. most importantly; switching to a `listbox` role and adding in the announcement functionality using `AriaAnnouncerManager`

#### Testing / Reviewing

some additional testing has been included thanks to Bob. for manual testing: run the demo app - in the sidebar enable `Enable autocomplete suggestions` - start voiceover using `command + F5` - begin typing - review the dialog that's currently being announced; number of suggestions and currently highlighted suggestion - use arrow keys to move up and down and review the announcement dialog - hit `Enter` and verify the selection has been captured in the prompt-line


packages/ai-chat-components/src/components/autocomplete/src/autocomplete.ts
```diff
@@ -131,13 +144,44 @@ class AutocompleteElement extends LitElement {
...
     document.removeEventListener("click", this._handleClickOutside);
+    this._announcer.disconnect();
+    if (this._moveAnnouncePending !== null) {
+      clearTimeout(this._moveAnnouncePending);
+      this._moveAnnouncePending = null;
+    }
+  }
+
+  firstUpdated() {
+    this._politeRegion1 = this.shadowRoot!.querySelector<HTMLDivElement>(
```
ethanwinters: we really need to get our local check to match the ci-check here. Can we clean these up?

packages/ai-chat-components/src/components/autocomplete/src/autocomplete.ts
```diff
@@ -294,6 +363,8 @@ class AutocompleteElement extends LitElement {
   }
 
   private _dismiss() {
+    this._openAnnounced = false;
+    this._announcer.announce("Suggestions closed.");
```
ethanwinters: we probably need these to be strings that get passed in rather than english only.

packages/ai-chat-components/src/components/autocomplete/src/autocomplete.ts
```diff
@@ -131,13 +144,44 @@ class AutocompleteElement extends LitElement {
...
+        this._announcer.announce("No suggestions.");
+        this._openAnnounced = false;
+        return;
+      }
+      // Reset to first item and announce list opened (once per show).
+      this._focusedIndex = 0;
+      if (!this._openAnnounced) {
+        this._openAnnounced = true;
+        this._announcer.announce(
+          `${totalItems} suggestion${totalItems === 1 ? "" : "s"}. Use up and down arrows to move, Enter to pick, Escape to close.`,
```
ethanwinters: hard coded english ain't what we are looking for

ethanwinters (changes requested):

davidmenendez:
@ethanwinters Good call on the i18n concerns. i included changes for a built in translation function. let me know if this solution works 👍

packages/ai-chat-components/src/components/autocomplete/src/autocomplete.ts
```diff
@@ -347,7 +479,12 @@ class AutocompleteElement extends LitElement {
             : ""
         }
 
-        <div class="${blockClass}__items">
+        <ul
```
warrenmblood, line 482: The switch from `<div>` to `<ul>` added this top & bottom margin

Image

packages/ai-chat-components/src/components/autocomplete/src/autocomplete-item-group.ts
```diff
@@ -100,30 +108,32 @@ class AutocompleteItemGroupElement extends LitElement {
...
     return html`
-      <div class="${blockClass}" role="listbox" aria-label="${this.title}">
+      <li role="group" aria-label="${this.title}" class="${blockClass}">
         ${
           this.title
             ? html` <div class="${blockClass}__title">${this.title}</div> `
             : null
         }
-        <div class="${blockClass}__items">
+        <ul class="${blockClass}__items">
```
warrenmblood, line 117: visual regression of added margin & horizontal padding to `cds-aichat-autocomplete-item-group__items`
Image

warrenmblood (changes requested):
when using VoiceOver, the suggestion item text isn't being announced while navigating the autocomplete via arrow up/down. It only gets announced after it is selected or when the item's send button is focused by pressing Tab.

I'm still not totally clear on how to keyboard navigate to the autocomplete item's send button. Focus remains inside the input by default while navigating the autocomplete with arrow keys, so pressing Tab while an item is highlighted moves focus past the input field. However, if you first bring focus inside the autocomplete using Shift+Tab, then Tab will focus on the highlighted item's send button.

Pressing Tab w/ focus outside autocomplete:
https://github.com/user-attachments/assets/34308409-2abb-4e70-b575-9fc852467e0a

Pressing Tab w/ focus inside autocomplete:
https://github.com/user-attachments/assets/341ca413-97ed-4927-bd7d-09f55a21d8bf

davidmenendez:
@warrenmblood thanks for pointing this all out. i've finally become aware of the chromatic review checks for the visual regression haha.

as for the focusing, yeah i'm also a but uncertain about that. it's in a weird place where it's some in-between of the combobox / listbox implementation and something else. i'm not exactly sure what the expected behavior is for those scenarios you pointed out. i believe it should just be to close it when focus leaves the input, but i'll verify with design and accessibility on this 👍

warrenmblood (approved):
LGTM

ethanwinters (approved):

PR #1786, opened 2026-07-15
feat: add alpha release workflow

### LINKED ISSUE #1774: Need to add a flow for publishing alpha releases to our github workflows
Need this for releasing a prompt line alpha

The alpha versions need to get published to npm as an alpha version.

They need demo site, storybooks and documentation sites published under chat.carbondesignsystem.com/tag/alpha

I don't think we need to add a tag to github for any alpha release.

We don't not need a change log.

These alpha releases will only be produced rarely and handed off to teams we are validating new features with.

They will be published from feature integration branches like `feat/prompt-line`.

`feat/prompt-line` will be out first alpha release.

ethanwinters:
Closes #1774

Adds a manual, off-cadence flow for cutting **alpha** releases of `@carbon/ai-chat` and `@carbon/ai-chat-components` straight from a feature integration branch (e.g. `feat/prompt-line`), so preview builds can be handed to teams validating in-progress features. Both packages publish to the npm `alpha` dist-tag with independent, auto-incrementing alpha versions, and the demo/docs/Storybooks publish under `chat.carbondesignsystem.com/tag/alpha`. The flow intentionally creates no git tag, GitHub release, changelog, `versions.js` update, or PR, and it never commits or pushes — the version bumps happen only in the throwaway runner checkout, so the branch it runs from is left untouched.

Files worth a closer look:

- [`.github/workflows/release-alpha.yml`](.github/workflows/release-alpha.yml) — the whole flow. The npm publish and all five CDN-upload steps are gated on `!inputs.dry-run`, and publishing uses `npm publish --workspace … --tag alpha` (not `lerna publish`) so no clean working tree / local commit is needed after the in-place version bump.
- [`scripts/compute-alpha-versions.js`](scripts/compute-alpha-versions.js) — derives each package's next `-alpha.N` from what is already on npm (no git state), and the workflow pins `@carbon/ai-chat` to the exact components alpha built in the same run so the published alpha stack is self-consistent.

#### Changelog

**New**

- `Release alpha` GitHub Actions workflow (`workflow_dispatch`, run from a selected feature branch) that publishes `@carbon/ai-chat` + `@carbon/ai-chat-components` to the npm `alpha` dist-tag and the demo/docs/both Storybooks to `chat.carbondesignsystem.com/tag/alpha`, with a `dry run` default that computes versions and builds but publishes nothing.
- `scripts/compute-alpha-versions.js`, which computes independent per-package next-alpha versions from the npm registry.
- "Alpha release" section (and TOC entry) in [`docs/publishing-releases.md`](docs/publishing-releases.md).

**Changed**

- The demo and TypeDoc version-selector switchers now recognize `/tag/alpha/` and surface an "Alpha" entry, shown only when viewing the alpha site so alpha stays unlisted on the stable `latest` / `next` sites.

**Removed**

- N/A

#### Testing / Reviewing

This is release automation, not reachable from the demo toggles, so verification is manual:

- **Version compute (local, no side effects):** `node scripts/compute-alpha-versions.js` prints the next alpha per package (e.g. `@carbon/ai-chat -> 1.18.0-alpha.0`, `@carbon/ai-chat-components -> 1.8.0-alpha.0`); `BUMP=patch node scripts/compute-alpha-versions.js` previews a patch-based alpha.
- **Dry run (safe):** after this lands on `main`, Actions → **Release alpha** → Run workflow → pick a feature branch, leave `dry run` checked. Confirm the log prints the planned versions and that the build + Storybook steps succeed while no publish/upload steps run.
- **Real run:** re-run with `dry run` unchecked, then confirm `npm view @carbon/ai-chat dist-tags` / `npm view @carbon/ai-chat-components dist-tags` show the new `alpha` version with `latest` / `next` unchanged, the `tag/alpha` demo/docs/Storybook URLs return 200, and no new git tag or commit appears on the source branch.
- **Switchers:** the `/tag/(latest|next|alpha)/` detection and conditional "Alpha" option are covered by review; "Alpha" appears only on `/tag/alpha/` paths.

Reviewer note: the workflow YAML has no repo CI gate (prettier excludes `.yml`, eslint is `packages/`-scoped), so please eyeball it — running `actionlint` is recommended.

### DIFF
--- a/.github/workflows/release-alpha.yml
+++ b/.github/workflows/release-alpha.yml
@@ -1,0 +1,176 @@
+# Publish a manual alpha release of @carbon/ai-chat and
+# @carbon/ai-chat-components from a feature integration branch (e.g.
+# feat/prompt-line).
+#
+# Alpha releases are rare, off-cadence previews handed to teams validating
+# in-progress features. Unlike the RC / stable release pipeline, this workflow:
+#   - publishes both packages to npm under the `alpha` dist-tag (never moves
+#     `latest` or `next`), each with its own independent alpha version
+#   - publishes the demo, docs, both Storybooks and the components bundle to
+#     chat.carbondesignsystem.com/tag/alpha (and components/.../tag/alpha)
+#   - creates NO git tag, GitHub release, changelog, versions.js update or PR
+#
+# It never commits or pushes: the version bumps happen only in the throwaway
+# runner checkout, so the branch it runs from is left untouched.
+#
+# Run it from the branch you want to publish: Actions -> Release alpha ->
+# "Run workflow" -> pick the feature branch. A workflow_dispatch run uses (and
+# checks out) the selected ref, so the alpha is built from that branch's code.
+name: Release alpha
+run-name: "Release alpha (${{ inputs.bump }}) dry-run:${{ inputs.dry-run }} from ${{ github.ref_name }}"
+
+concurrency:
+  group: release-alpha
+  cancel-in-progress: false
+
+on:
+  workflow_dispatch:
+    inputs:
+      bump:
+        description: "Base bump off the current stable version (the alpha previews the next minor or patch)."
+        type: choice
+        options:
+          - minor
+          - patch
+        default: minor
+      dry-run:
+        description: "Dry run? Computes versions and builds everything, but publishes nothing to npm or the CDN."
+        type: boolean
+        default: true
+
+# id-token: write enables npm provenance (both packages set
+# publishConfig.provenance). Nothing is pushed, so contents: read is enough —
+# no PAT / contents: write, unlike the RC / stable release workflow.
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  release-alpha:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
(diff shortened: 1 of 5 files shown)


ethanwinters:
Of note, there is a "dry run" on this, and we are going to not really going to be able to test this until we run a dry run post merge.

.github/workflows/release-alpha.yml
```diff
@@ -0,0 +1,176 @@
...
+  release-alpha:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          fetch-depth: "0"
+
+      - name: Use Node.js version from .nvmrc
+        uses: actions/setup-node@v6.4.0
```
davidmenendez, line 58: super minor, but we should be using the hash version

davidmenendez (approved):
LGTM! actionlint passed and everything looks in order. one minor thing, there's a few spots in the `release-alpha.yml` where we're not installing dependencies using the commit hash, which we should continue to do for security reasons. although it looks like there's a couple dozen instances of this being the case in our `.github` folder, so might just be worth doing separately as part of a cleanup of our workflow files. i won't stop it from approval 😂

PR #1789, opened 2026-07-15
fix(prompt-line): visual regressions

warrenmblood:
Fixes visual regression where the width of the autocomplete list is compressed. Also fixes bug where the `<svg>` within an icon button receives focus on click.

#### Changelog

**Changed**

- `autocomplete-content` slot styles: `inline-size` and `inset-inline-start` set to account for padding and margin now applied to `.--input-uploads-and-autocomplete`
- `carbonIconToReact` util function:
  - Adds attributes `focusable: "false"` and `style: { pointerEvents: "none" }` to prevent focus on click events

#### Testing / Reviewing

- Go to demo deploy preview Chat config -> Input -> Enable autocomplete suggestions, expanded layout, and dummy menu options
- Verify the autocomplete list aligns with the width of the input (matches the designs)
- Click on the menu action buttons and make sure the svg cannot be focused by either clicking or pressing Tab between buttons
- Test in Fullscreen, Float, and Sidebar chat layouts

### DIFF
--- a/packages/ai-chat/src/chat/utils/carbonIcon.ts
+++ b/packages/ai-chat/src/chat/utils/carbonIcon.ts
@@ -75,7 +75,9 @@ export function carbonIconToReact(
         width: icon.attrs.width || 16,
         height: icon.attrs.height || 16,
         fill: icon.attrs.fill || "currentColor",
+        focusable: "false",
         tabIndex: -1,
+        style: { pointerEvents: "none" },
         ...transformedProps,
       },
       icon.content.map((child, i) =>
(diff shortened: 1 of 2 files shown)


packages/ai-chat/src/chat/utils/carbonIcon.ts
```diff
@@ -75,7 +75,9 @@ export function carbonIconToReact(
         width: icon.attrs.width || 16,
         height: icon.attrs.height || 16,
         fill: icon.attrs.fill || "currentColor",
+        focusable: "false",
```
ethanwinters, line 78: huzzah!

ethanwinters (approved):

PR #1795, opened 2026-07-17
feat: ChatSDK lifecycle facade, boot pipeline, and instance reuse

### LINKED ISSUE #1732: Add the acquire and boot pipeline in src/sdk/, and enforce the SDK boundary

### LINKED ISSUE #1733: Reuse the live chat instance across a host remount via featureFlags.reuseInstance
### Background

A host that unmounts and remounts the chat loses the conversation. React StrictMode, a changing `key`, a conditional render, or a props refactor is enough. The chat cold-boots, and the transcript does not come back on its own, because saved state carries view flags and no messages.

The human-agent case is worse than a lost transcript. On the fresh boot, the chat reads the saved connection flag. It then either forces a reconnect or ends the chat and writes "chat was ended" into the transcript. So an ordinary render can tell a user their conversation with an agent is over.

Hosts cannot reliably avoid this, because a remount can start anywhere above the chat. The only fix available today is advice: mount once, and never move the component.

Parent epic: #1728. ADR-0003 governs it, and decides that a chat's lifetime belongs to whoever acquired it rather than to the host mount.

### Goal

With `featureFlags.reuseInstance` on, an unmount and remount keeps the running conversation instead of cold-booting a new one. In-session only, held in memory, not across a page reload.

### Acceptance criteria

- [ ] With the flag on, a remount adopts the running chat. A spec sends a message, unmounts, remounts inside the window, and finds the same transcript and the same agent connection.
- [ ] With the flag off — the default — every mount cold-boots, exactly as today. The existing boot specs pass unchanged.
- [ ] Two mounts on one namespace adopt each other. Two on different namespaces do not.
- [ ] Each acquire takes its own claim, and the grace window opens only when the last claim drops. A spec with two claims releases one and asserts no unload timer started.
- [ ] Releasing the same claim twice is a no-op. It starts no second window and does not throw.
- [ ] An acquire inside the window adopts the chat and cancels the pending unload.
- [ ] When the window runs out, the chat unloads exactly as an immediate teardown would, agent connection included. Asserted against the same spec #1681 adds.
- [ ] The window defaults to 3000 ms and honors `featureFlags.reuseInstanceGraceMs`.
- [ ] An adopting acquire replaces the config rather than patching it. A spec adopts with a config that omits `serviceDeskFactory` and asserts the factory is gone, not inherited.
- [ ] A second live acquire from a *different* config on a claimed namespace logs a console error in a development build, naming the namespace and the fix. A production build logs nothing.
- [ ] A second live acquire from the same shell on the same namespace — the ordinary two-claim case — logs nothing.
- [ ] The registry imports no React and no Lit, so the fence from #1732 passes over it.
- [ ] `npm run build --workspace=@carbon/ai-chat` exits 0. The new config fields change published types, so TypeDoc link validation runs.
- [ ] The gate for `packages/ai-chat` passes, per the repository's definition-of-done commands.

Registry behavior is proved by `tests/services/spec/reuseInstanceRegistry_spec.ts`, adoption through the shells by `tests/config/spec/reuseInstanceAdoption_spec.tsx`, and the console error by `tests/config/spec/rebootWarning_spec.tsx`.

### Public API / contract

Two new public config fields, and nothing else. The `featureFlags` key is itself new, and `reuseInstance` is its first member.

```ts
/**
 * Opt into behaviors that are still stabilizing.
 *
 * @category Config
 */
export interface PublicConfigFeatureFlags {
  /**
   * Keep the live in-memory chat across a host unmount and remount instead of cold-booting a new
   * one, so the conversation survives a remount. In-session only — not carried across a page
   * reload. Defaults to `false`.
   *
   * @experimental
   */
  reuseInstance?: boolean;

  /**
   * How long, in milliseconds, to keep the chat alive after the last host unmounts before
   * disposing it. Applies only when {@link PublicConfigFeatureFlags.reuseInstance} is on.
   * Defaults to 3000.
   *
   * @experimental
   */
  reuseInstanceGraceMs?: number;
}

export interface PublicConfig {
  /** Opt into behaviors that are still stabilizing. See {@link PublicConfigFeatureFlags}. */
  featureFlags?: PublicConfigFeatureFlags;
  // …existing fields
}
```

Both are surfaced on the flattened web-component config beside the other boolean and numeric props.

**Namespace is the identity.** Two mounts sharing a namespace adopt each other even when their configs differ. So a host running two genuinely separate chats has to give them different namespaces. An adopting acquire replaces the config through the same path the shells already use for prop diffs. Anything the second config leaves out is removed rather than inherited. That includes a `serviceDeskFactory`, and dropping it severs a live agent conversation the first mount was holding. Development builds make that mistake loud. Production builds stay silent.

**Timing and ownership.** In 1.x a host reaches this through the shell only, and an unmount plays the part of a release. There is no early-teardown verb. A host that wants the chat gone waits out the window, three seconds by default. That cost is taken knowingly, and the default bounds it.

**No flag says whether an acquire adopted.** ADR-0003 rejects that, which is why the drafted `onAttach` callback and its `remount` detail are not built. A per-call flag is true only once, and it lies while boot is still in flight. A host with boot-once work asks a durable question instead — `instance.state.messages.get().length === 0` — because that answer is true whenever it is asked. The shells already hand an instance to `onBeforeRender` and `onAfterRender`. With the remount fact gone, a third callback would carry nothing.

**Repeat calls, and when a second claim is a mistake.** Releasing a claim twice is a no-op. Acquiring twice on one namespace yields two claims over one chat, not two chats — that is the mechanism a remount rides on, since the new mount claims before the old one releases. What earns the console error is narrower: a second live claim whose config differs from the running one, which means two hosts think they own different chats and one of them is about to lose its configuration. Matching configs stay silent.

**The docs do not promise a default flip.** ADR-0003 leaves open whether a later major turns reuse on by default, and says so precisely to stop the prototype's wording deciding it. So the published JSDoc says the flag is stabilizing, not that it is destined to become the default.

**Limits of the guarantee.** The registry is module-level state. Two copies of the package on a page do not share it, so reuse quietly does not happen across them. Reuse stays opt-in for all of 1.x. Whether a later major turns it on by default is deliberately left open.

Everything else is unchanged: no new events, no new instance members, no failure path a host can catch, and nothing announced to a screen reader. An adopted chat never re-renders from scratch, so focus stays where the host left it.

### Out of scope

- Surviving a page reload. That is the host's job through `customLoadHistory`.
- The public `release()` and `destroy()` verbs. Those are SDK surface, no earlier than 2.0.0.
- A public `instance.destroy()`. ADR-0025 keeps lifecycle off the instance; see #1681.
- `onAttach` and `OnAttachDetails`. Not built, per the contract above.
- The teardown the window expires into (#1681) and the acquire path itself (#1732).

### Related

- Parent epic #1728.
- Depends on #1732 (the acquire this registers against) and #1681 (the teardown the window expires into).
- Land #1669 first if the `reuseInstance` docs are to `{@link}` the `state.messages` store they point hosts at.
- Advances #1650, whose input-disabled symptom is a remount artifact.

ethanwinters:
Closes #1732
Closes #1733

Wave B of the `feat: persist api` decomposition (epic #1728): the internal **`ChatSDK` lifecycle facade + boot pipeline**, the **reuse-instance** groundwork, and mechanical enforcement of the **framework-agnostic SDK boundary**. This is the spine the earlier slot-state stores (#1742) and the later conversation-read API (#1669) hang off of.

It replaces the React-callback boot in `chat/utils/chatBoot.ts` with two layers:

- **`src/chat/sdk/`** — the headless, framework-agnostic surface a shell (or a 2.0 compose-your-own consumer) drives: `acquireChatSDK` (create-or-adopt), the `ChatSDK` class (`instance`, `slotStates`, `attach`, `release`, `destroy`, `runInitialViewChange`), and the `reuseInstanceRegistry`.
- **`src/chat/boot/appBoot.ts`** — shell plumbing run _around_ `acquireChatSDK` (container CSS, config defaults, the accidental-remount diagnostic), deliberately outside the fenced core.

Files worth reviewing carefully:

- [`src/chat/services/reuseInstanceRegistry.ts`](packages/ai-chat/src/chat/services/reuseInstanceRegistry.ts) — the ref-counted, grace-timer'd manager cache. Review the two-cold-boots-overlap handling in `registerServiceManager` and the "displaced manager disposes itself" identity checks in `release`/`evict`.
- [`src/chat/sdk/ChatSDK.ts`](packages/ai-chat/src/chat/sdk/ChatSDK.ts) — `acquireChatSDK` is **headless** (binds no DOM); the shell calls `sdk.attach({ container, customHostElement })` after acquire on both the cold-boot and reuse-adopt paths. Note the deliberate layering choice: the core does **not** hold a `serviceManager.sdk` back-reference (that would invert the layering that lets `sdk/` become `@carbon/ai-chat/sdk` in 2.0). Instead `ServiceManager` exposes a bare `onDestroy?: () => void` hook and `ChatSDK.ts` owns a module-level `WeakMap<ServiceManager, ChatSDK>`; `ChatInstanceImpl.destroy` calls `serviceManager.onDestroy?.()`.
- [`src/chat/ChatAppEntry.tsx`](packages/ai-chat/src/chat/ChatAppEntry.tsx) — the boot effect now subscribes the slot-state stores via `useSyncExternalStore`, and its cleanup owns teardown through a `cancelled` closure flag that covers an unmount mid-boot. Review the three `cancelled` checks and the `needsBootstrap` gate.
- [`src/web-components/cds-aichat-container/cds-aichat-internal.tsx`](packages/ai-chat/src/web-components/cds-aichat-container/cds-aichat-internal.tsx) — `disconnectedCallback` now defers React teardown to a microtask so a transient DOM move (disconnect→reconnect) keeps the live root instead of re-booting.
- Both WC hosts ([container](packages/ai-chat/src/web-components/cds-aichat-container/index.ts), [custom-element](packages/ai-chat/src/web-components/cds-aichat-custom-element/index.ts)) — per-element bus handlers moved into an `onAttach`-fired `registerInstanceHandlers`/`removeInstanceHandlers` pair so a fresh element adopting a reused instance re-wires cleanly and orphaned handlers don't keep firing on a detached node.

**On the regroup:** this PR intentionally delivers **two issues** — #1732 (ChatSDK facade + boot pipeline + boundary) and #1733 (reuse-instance). They were split in the original plan but are interwoven through the internals of these six files; stripping the reuse branches out would have forced temporary shapes the reuse work immediately reverts. The conversation-read API (#1669) stays a separate follow-up PR because it's additive (~13 lines touch shared files).

#### Major changes

**New**

- `featureFlags.reuseInstance` (opt-in) keeps a live `ServiceManager` in memory across a host unmount/remount (React StrictMode, a changing `key`, conditional rendering) instead of cold-booting and discarding the conversation. `featureFlags.reuseInstanceGraceMs` tunes the release grace window (default 3000ms).
- Internal `ChatSDK` lifecycle facade and headless `acquireChatSDK` create-or-adopt boot pipeline (`src/chat/sdk/`), plus the shell boot pipeline (`src/chat/boot/appBoot.ts`).
- `onAttach(instance, { remount })` callback on the React and web-component surfaces — fires on every mount/attach (unlike boot-once `onAfterRender`), so hosts can re-wire per-mount setup and run one-time setup only on the first attach.
- Framework-agnostic **SDK boundary** enforcement: an ESLint `no-restricted-imports` fence on `src/chat/{services,store,events,instance,schema,sdk}/` and a transitive import-graph guard (`tests/sdk/spec/sdkBoundary_spec.ts`) that walks the `sdk/index.ts` barrel.

**Changed**

- Boot moved off the React-callback `chatBoot.ts` onto `acquireChatSDK` + the core-owned slot-state value stores (`ChatAppEntry` now subscribes via `useSyncExternalStore`; the WC hosts subscribe to the shared stores instead of re-accumulating bus events locally).
- `ChatInstanceImpl.destroy` now routes through `serviceManager.onDestroy` (installed by the SDK layer) so teardown includes reuse-registry eviction without the core naming the SDK.

**Removed**

- `src/chat/utils/chatBoot.ts` and its spec (replaced by `sdk/` + `boot/`).

#### Minor changes

**Changed**

- `PublicConfig` gains a `featureFlags` field (`PublicConfigFeatureFlags`), flattened for the web-component surface (`feature-flags`) and exported from `aiChatEntry.tsx` / `serverEntry.ts` alongside `OnAttachDetails`.
- `onAfterRender` JSDoc across the surfaces clarified: it fires once per boot and does not gate rendering (its return value is not awaited); it does not re-fire on a reuse re-attach.
- Fixed a dead SSR guard in `ChatAppEntry` (`if (!isBrowser)` → `if (!isBrowser())` — `isBrowser` is a function, so the old form never short-circuited).

#### Testing / Reviewing

Not demo-reachable as a UI toggle (reuse is a lifecycle/remount behavior), so verification is unit + a manual remount check.

**Automated** — all green locally (`tsc --noEmit` clean, `npm run build --workspace=@carbon/ai-chat` clean, 296 specs across the touched areas):

- `tests/sdk/spec/chatSDK_spec.ts` — cold boot, headless acquire→attach, adopt-same-facade, grace-window disposal, `destroy()`.
- `tests/services/spec/reuseInstanceRegistry_spec.ts` — ref-count / grace-timer / evict bookkeeping.
- `tests/config/spec/reuseInstanceAdoption_spec.tsx`, `chatCustomElement_spec.tsx`, `rebootWarning_spec.tsx` — React remount adoption + the accidental-reboot diagnostic.
- `tests/web-components/spec/internalDisconnectedRender_spec.ts`, `slotStateAdoption_spec.ts` — the deferred-disconnect microtask and WC slot-state re-subscription across remount.
- `tests/sdk/spec/sdkBoundary_spec.ts` — the transitive boundary guard.

**Manual** — with a chat mounted, set `featureFlags.reuseInstance: true`, send a few messages, then unmount and remount the host within the grace window (toggle a `key`, or React StrictMode). Confirm the conversation, view state, and any user-defined-response / custom-footer slot content survive the remount without a cold boot.

**Reviewer note (test gap):** the unmount-*during*-the-async-`acquireChatSDK` race (the `cancelled` branch in `ChatAppEntry`'s boot effect) is exercised only indirectly. A regression that dropped the mid-boot `sdk.release()` would leak a `ServiceManager` on a fast unmount and wouldn't be caught by the current specs — worth a dedicated test in a follow-up.


ethanwinters:
This PR was fast-forward-merged into `sdk-persistence-integration` by mistake before review. The integration branch has been reset to its pre-merge state (`7ad8589c`) and the review is reopened as #1798 (same fork branch, same commits, same diff). Please review there — this PR can't be reopened once merged.

PR #1814, opened 2026-07-20
feat: allow saying an explicit version for the alpha

ethanwinters:
e.g. as is, if i publish an alpha for prompt line it comes out as 1.18.0-alpha.0... but I want to say 1.19.0-alpha.0

The alpha release workflow could only ever preview the *next* release: it derived the base version by bumping each package's current stable version, so an alpha cut from `feat/prompt-line` came out as `1.18.0-alpha.0` even though that work ships in `1.19.0`. This adds optional `ai-chat-base` / `components-base` workflow inputs that set the base version explicitly, leaving the existing `bump`-derived behavior as the default when both are blank.

- [`scripts/compute-alpha-versions.js`](scripts/compute-alpha-versions.js) — the base version is now either an explicit override or the previous `nextBase()` derivation, guarded by three validations worth reviewing together: the format check in `requestedBase()`, the "must be ahead of current" check in `computeAlpha()`, and `assertCompleteOverride()`, which rejects a half-filled override. The last one exists because the two packages version independently (`@carbon/ai-chat` is at 1.17.0, `@carbon/ai-chat-components` at 1.7.0) — overriding only one would silently pair an alpha of one release with an alpha of another. It runs before any `npm view` calls so a bad input fails in seconds.

#### Changelog

**New**

- `ai-chat-base` and `components-base` inputs on the Release alpha workflow, setting the exact base version each package's alpha previews (e.g. `1.19.0-alpha.0` instead of `1.18.0-alpha.0`).
- Validation of those inputs: they must be plain `x.y.z` versions, must be ahead of the package's current version, and must be set for both packages or neither. Failures exit 1 with a single-line message rather than a stack trace.

**Changed**

- `bump` is ignored for a package with an explicit base; its input description now says so. With both base inputs blank, version computation is unchanged.
- [`docs/publishing-releases.md`](docs/publishing-releases.md) gained an alpha-release checklist item covering when and how to set the bases.

**Removed**

- Nothing.

#### Testing / Reviewing

Not reachable from the demo site — this is release tooling, exercised through the script directly and through a workflow dry run.

Run the version computation locally from the repo root (it only reads `package.json` and queries npm; it publishes nothing):

```sh
# Default: unchanged behavior, derived from `bump`.
node scripts/compute-alpha-versions.js
# -> @carbon/ai-chat-components -> 1.8.0-alpha.0
# -> @carbon/ai-chat -> 1.18.0-alpha.0

# Explicit bases, the case this PR adds.
AI_CHAT_BASE=1.19.0 COMPONENTS_BASE=1.9.0 node scripts/compute-alpha-versions.js
# -> @carbon/ai-chat-components -> 1.9.0-alpha.0
# -> @carbon/ai-chat -> 1.19.0-alpha.0

# Patch bump still works.
BUMP=patch node scripts/compute-alpha-versions.js
```

Each of these should exit 1 with one line of output and no stack trace:

```sh
AI_CHAT_BASE=1.19.0 node scripts/compute-alpha-versions.js
# -> Base versions must be set for every package or none. Missing: COMPONENTS_BASE (@carbon/ai-chat-components).

AI_CHAT_BASE=1.19 COMPONENTS_BASE=1.9.0 node scripts/compute-alpha-versions.js
# -> Invalid base version "1.19" for @carbon/ai-chat: expected a plain x.y.z release version with no prerelease suffix.

AI_CHAT_BASE=1.16.0 COMPONENTS_BASE=1.9.0 node scripts/compute-alpha-versions.js
# -> Base version 1.16.0 for @carbon/ai-chat is not ahead of its current version 1.17.0.
```

### DIFF
--- a/.github/workflows/release-alpha.yml
+++ b/.github/workflows/release-alpha.yml
@@ -27,12 +27,20 @@ on:
   workflow_dispatch:
     inputs:
       bump:
-        description: "Base bump off the current stable version (the alpha previews the next minor or patch)."
+        description: "Base bump off the current stable version (the alpha previews the next minor or patch). Ignored for a package with an explicit base below."
         type: choice
         options:
           - minor
           - patch
         default: minor
+      ai-chat-base:
+        description: "Exact base version for @carbon/ai-chat, e.g. 1.19.0 when the branch lands a release later than the next one. Leave both base inputs blank to derive from `bump`; setting one requires the other."
+        type: string
+        default: ""
+      components-base:
+        description: "Exact base version for @carbon/ai-chat-components, e.g. 1.9.0. The two packages version independently, so this must be set whenever `ai-chat-base` is."
+        type: string
+        default: ""
       dry-run:
         description: "Dry run? Computes versions and builds everything, but publishes nothing to npm or the CDN."
         type: boolean
@@ -69,6 +77,8 @@ jobs:
       - name: Compute alpha versions
         env:
           BUMP: ${{ inputs.bump }}
+          AI_CHAT_BASE: ${{ inputs.ai-chat-base }}
+          COMPONENTS_BASE: ${{ inputs.components-base }}
         run: node scripts/compute-alpha-versions.js
 
       # Apply the versions in the runner checkout only (no commit, no tag). Pin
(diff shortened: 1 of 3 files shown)


davidmenendez (approved):
LGTM!

PR #1901, opened 2026-07-22
fix: pagination tooltip position

### LINKED ISSUE #1152: Tooltip labels overflow outside widget boundary
https://github.com/carbon-design-system/carbon-ai-chat/issues/1144 had two issues in it. Issue 1 is resolved, this is issue 2.

Tooltip labels overflow outside widget boundary

When hovering over interactive elements near the right edge of the chat widget, tooltip labels extend beyond the widget boundary and are clipped. This is visible on:

- The "Next page" pagination button tooltip on tables (both "Next page" in v1.9.1 and "Nächste Seite" in v1.6.0 are cut off).
- The copy button tooltip on code snippets ("In Zwischenablage kopieren" is truncated to "In Zwischenablage k...").

The tooltips do not reposition to remain within the visible widget area.

## Screenshots illustrating both issues are attached for reference:

**v1.6.0**
Image
Image

**v1.9.1**
Image
Image

### Reproduction / example

https://stackblitz.com/edit/github-3ybcxztj?file=src%2FApp.tsx

davidmenendez:
Closes #1152 

address small UI bug where the carbon table pagination tooltip was getting cut off in the float view by utilizing an updated pagination API from https://github.com/carbon-design-system/carbon/issues/22106

Screenshot 2026-07-22 at 2 37 34 PM

#### Changelog

**Changed**

- `packages/ai-chat-components/src/components/table/src/table-pagination.template.ts` adds `forward-text-tooltip-position="left"` to the `cds-pagination` component to avoid the tooltip being cutoff by the floating variant

#### Testing / Reviewing

go to the demo, enable float mode, enter `table` in the prompt line, scroll down to the pagination part of the table, and hover over the pagination icon button to verify that it's showing up appropriately on the left and not being cut off.

### DIFF
--- a/packages/ai-chat-components/src/components/table/src/table-pagination.template.ts
+++ b/packages/ai-chat-components/src/components/table/src/table-pagination.template.ts
@@ -13,6 +13,7 @@ import "@carbon/web-components/es/components/select/index.js";
 import { html } from "lit";
 import { ifDefined } from "lit/directives/if-defined.js";
 import { TableRowContent } from "./table.js";
+import { isDirectionRTL } from "../../../globals/utils/rtl-utils.js";
 
 // Import only the constants, not the class
 const POSSIBLE_PAGE_SIZES = [5, 10, 15, 20, 50];
@@ -81,7 +82,7 @@ function tablePaginationTemplate(props: TablePaginationProps) {
     .formatStatusWithDeterminateTotal=${ifDefined(getPaginationStatusText)}
     @cds-pagination-changed-current=${handlePageChangeEvent}
     @cds-page-sizes-select-changed=${handlePageSizeChangeEvent}
-    forward-text-tooltip-position="left"
+    forward-text-tooltip-position=${isDirectionRTL() ? "right" : "left"}
   >
     ${supportedPageSizes.map(
       (pageSize) =>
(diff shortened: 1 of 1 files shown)


packages/ai-chat-components/src/components/table/src/table-pagination.template.ts
```diff
@@ -81,6 +81,7 @@ function tablePaginationTemplate(props: TablePaginationProps) {
     .formatStatusWithDeterminateTotal=${ifDefined(getPaginationStatusText)}
     @cds-pagination-changed-current=${handlePageChangeEvent}
     @cds-page-sizes-select-changed=${handlePageSizeChangeEvent}
+    forward-text-tooltip-position="left"
```
warrenmblood: this won't work if text direction is Right-to-Left. We have a utility `isDirectionRTL` you can use to set this value as right or left

warrenmblood (changes requested):

packages/ai-chat-components/src/components/table/src/table-pagination.template.ts
davidmenendez: ty for reminding me that this exists!

davidmenendez:
@warrenmblood changes pushed 👍

warrenmblood (approved):
LGTM

Issue #1934, opened 2026-07-23
[prompt-line]: Enable option to toggle starters on/off


warrenmblood:
The prompt-line conversation starters are currently only triggered by the prompt line editor being focused and empty. There is no other way for users to turn the starters list on or off. Users may want to control this, for example through a prompt line action button that toggles the starter list on/off.

Add a config option to `StartersConfig` that allows users to enable/disable the starters list.

PR #1936, opened 2026-07-23
feat(prompt-line): conversation starters example

### LINKED ISSUE #1351: Example: Conversation starters
work with design on the "conversation starters" pattern and build an example showing it in use.

warrenmblood:
Closes #1351 #1934

Adds new example to showcase the usage of Starter prompts via `InputConfig.starters` to show suggestions in the prompt line while the editor is empty, focused, and editable.

Also adds option `StartersConfig.isOn` to grant users control over the starters list opening/closing.

#### Changelog

**New**
- Two new examples: the React and WC versions of `prompt-line-conversation-starters`
- Adds `StartersConfig` interface `(items: SuggestionItem[] + renderCustomList?)` and replaced all usage of `SuggestionItem[]` for `AutocompleteControllerOptions.starters` with it. This allows users to pass `renderCustomList` to replace the built-in list UI, for example to add a header above the items.
- Adds `StartersConfig.isOn`. When `isOn` is changed, push the new value into the Tiptap extension storage, then either re-focus the editor to show list or fire a no-op transaction that clears the trigger and hides the list.

**Changed**
- Docs updates to reflect changes in this PR
- Updates conversation starters tests in `packages/ai-chat/tests/components/spec/promptLineMode_spec.ts` & `packages/ai-chat-components/src/components/prompt-line/src/__tests__/autocomplete-controller.test.ts` to accept `StartersConfig` instead of `SuggestionItem[]`

#### Testing / Reviewing

Test the React & Web component example:
- `npm run start --workspace=@carbon/ai-chat-examples-react-prompt-line-conversation-starters`
- `npm run start --workspace=@carbon/ai-chat-examples-web-components-prompt-line-conversation-starters`


examples/react/prompt-line-conversation-starters/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "@carbon/ai-chat": "^1.18.0-rc.0",
+    "@carbon/ai-chat-components": "^1.8.0-rc.0",
+    "@carbon/icons-react": "^11.53.0",
+    "@carbon/web-components": "^2.58.1",
+    "build": "^0.1.4",
+    "lit": "^3.1.0",
+    "npm": "^12.0.1",
+    "react": "^19.0.0",
+    "react-dom": "^19.0.0",
+    "run": "^1.5.0"
```
ethanwinters: `build`, `npm`, and `run` are unrelated squatter packages on the public registry, not real dependencies. This is the signature of an `npm install npm run build` typo.

They pull **151 entries** into `package-lock.json` — essentially the entire 2515-line lockfile growth in this PR — and vendor the whole npm CLI into an example. The sibling `examples/react/prompt-line-typeahead/package.json` has none of them.

```bash
# delete the three lines, then:
rm -rf node_modules/{npm,build,run}
npm install
```

`@carbon/ai-chat-components` and `@carbon/icons-react` are genuinely imported by `App.tsx`, so those should stay.

packages/ai-chat-components/src/components/prompt-line/src/prompt-line-rich-runtime.ts
```diff
@@ -326,6 +326,7 @@ class RichController implements PromptLineController {
       this._editor.commands.setTextSelection({ from, to });
     }
     if (wasFocused) {
+      this._focusFromMouse = true;
```
ethanwinters: This tells the focus handler that a *programmatic* refocus was mouse-driven, which costs keyboard users their visible focus indicator.

`_focusFromMouse` gates `PM_KEYBOARD_FOCUS_CLASS` in `_wireEditorEvents`, so setting it to `true` here means the restored editor comes back with no focus outline, and `cds-aichat-prompt-focus` reports `keyboard: false` to hosts.

Verified in the React example — with the ring present, one click on the starters toggle gives:

```
keyboardFocusRing_before: true
keyboardFocusRing_after:  false
stillFocused:             true
```

That's WCAG 2.1 AA 2.4.7 Focus Visible.

Preserving the original focus provenance instead of hardcoding it should do:

```ts
const wasKeyboardFocus = this._editor.view.dom.classList.contains(PM_KEYBOARD_FOCUS_CLASS);
// ...read this before destroy(), the node is gone afterwards
if (wasFocused) {
  this._focusFromMouse = !wasKeyboardFocus;
  this._editor.commands.focus();
}
```

See the review summary for why this path runs on every `isOn` toggle rather than rarely.

packages/ai-chat-components/src/components/prompt-line/src/tiptap/types.ts
```diff
@@ -165,6 +165,31 @@ export interface AutocompleteConfig extends BaseSuggestionConfig {
...
+> {
+  /** The starter prompts to display. */
+  items: SuggestionItem[];
+  /**
+   * Controls whether the starters list is active. Defaults to `true`.
+   *
+   * When true, the list appears automatically whenever the editor is focused
+   * and empty. Set to false to suppress the list without removing the config
+   * entirely. Keeping the config present with isOn: false leaves the editor
+   * intact and lets you toggle the list on and off instantly.
```
ethanwinters, line 188: This claim isn't true today, and both example READMEs repeat it.

Toggling `isOn` destroys and recreates the Tiptap editor. Verified in the browser by tagging the live ProseMirror node: `sameEditorNode: false, recreated: true`.

Root cause and a suggested fix are in the review summary (it lives in `useInputExtensions.ts`, which this PR doesn't otherwise touch). Either fix the behavior or soften this sentence and the matching README lines.

packages/ai-chat-components/src/components/prompt-line/src/autocomplete-controller.ts
```diff
@@ -154,6 +158,26 @@ export class AutocompleteController {
...
+    // immediately (disable path).
+    if ("starters" in next && prevIsOn !== this._starters?.isOn) {
+      const editor = this._promptLine?.getEditor();
+      const storage = (editor?.storage as unknown as Record<string, unknown>)
+        ?.carbonStarterTrigger as StarterTriggerStorage | undefined;
+      if (storage && editor) {
+        storage.isOn = this._starters?.isOn !== false;
+        if (storage.isOn) {
+          editor.view.dom.dispatchEvent(
+            new PointerEvent("pointerdown", { bubbles: true }),
```
ethanwinters: Synthesizing a pointer event to flip a private flag in another module is action-at-a-distance, and it emits an unbalanced `pointerdown` with no matching `pointerup` that any other listener on that subtree will see as a phantom press.

The only purpose of this dispatch is to set `_focusFromMouse` inside `prompt-line-rich-runtime.ts` so the subsequent `.focus()` doesn't draw a keyboard focus ring. That coupling is invisible from either side.

The deeper issue is what `isOn` means. Right now `isOn: true` doesn't *show* the list — it only un-gates focus-driven display, so the controller has to manufacture visibility by faking a click and refocusing. A cleaner model is to make `isOn` an explicit visibility input that `maybeEmit` consults directly, rather than a filter layered on top of focus:

```ts
const isActive =
  !forceClear &&
  storage?.isOn !== false &&
  editor.isEditable &&
  editor.isEmpty;      // focus dropped from the gate
```

The `editor.isEmpty` term still handles "hide when there are no matches" — I confirmed typing `Gen` fires `trigger-change -> null` — and dropping focus from the condition removes the need for this synthetic event entirely.

packages/ai-chat-components/src/components/prompt-line/src/autocomplete-controller.ts
```diff
@@ -132,6 +134,8 @@ export class AutocompleteController {
       >
     >,
   ): void {
+    const prevIsOn = this._starters?.isOn;
```
ethanwinters: `isOn` defaults to on when omitted, but this compares raw values, so `undefined` and `true` read as a change.

A consumer going from `{ items }` to `{ items, isOn: true }` — no semantic change — takes the enable branch below and gets a surprise focus steal. Normalizing both sides fixes it:

```ts
const prevIsOn = this._starters?.isOn !== false;
// ...
if ("starters" in next && prevIsOn !== (this._starters?.isOn !== false)) {
```

examples/react/prompt-line-conversation-starters/src/App.tsx
```diff
@@ -0,0 +1,164 @@
...
+        // `isOn: false` suppresses the list without removing the config, keeping
+        // the rich editor alive so toggling back on is instant.
+        starters: { ...starters, isOn: startersEnabled },
+
+        // Single toggle action: enables or disables the conversation starters.
+        // Disabled when the input has text because starters only trigger on an
+        // empty editor.
+        actions: [
+          {
+            text: startersEnabled
```
ethanwinters, line 141: This toggle's on/off state is conveyed *only* by the tooltip text flipping between "Hide" and "Show", which requires hovering to discover and gives screen readers no toggle semantics. I checked the rendered button: `aria-pressed` is `null`.

`ToolbarAction` genuinely has no field for this today, so it isn't fixable inside this PR — filed as carbon-design-system/carbon-ai-chat#1963.

Since this example is what will teach the pattern, I'd either link that issue in a comment here, or demonstrate `isOn` with a control that carries its own visible state until #1963 lands.

packages/ai-chat-components/src/components/prompt-line/src/__tests__/autocomplete-controller.test.ts
```diff
@@ -35,10 +36,12 @@ const COMMANDS: SuggestionItem[] = [
   { id: "translate", label: "translate" },
 ];
 
-const STARTERS: SuggestionItem[] = [
-  { id: "hello", label: "Say hello" },
-  { id: "intro", label: "Introduce yourself" },
-];
+const STARTERS: StartersConfig = {
```
ethanwinters, line 39: This file only migrates the existing assertions to the new object shape; nothing here exercises `isOn`, which is the actual new feature.

Worth adding to this spec:

- `isOn: false` suppresses the starter trigger while the editor is empty + focused
- toggling `false -> true` re-emits the trigger
- `isOn: undefined` behaves as `true`
- a `renderCustomList` on `StartersConfig` reaches `_state.renderCustomList` (also new in this PR, also untested)

examples/react/prompt-line-conversation-starters/src/App.tsx
```diff
@@ -0,0 +1,164 @@
...
+      onSelect={(e: CustomEvent<{ item: (typeof STARTER_ITEMS)[number] }>) =>
+        onSelect(e.detail.item)
+      }
+      onDismiss={onDismiss}
+    />
+  );
+}
+
+function App() {
+  const renderCustomList = useMemo(() => StarterList, []);
```
ethanwinters: `StarterList` is a module-level function declaration, so its identity is already stable across renders. This `useMemo` doesn't do anything; `renderCustomList: StarterList` directly is equivalent and clearer for an example whose job is to teach the API.

packages/ai-chat-components/src/components/autocomplete/src/autocomplete.ts
```diff
@@ -361,6 +363,15 @@ class AutocompleteElement extends LitElement {
     );
   }
 
+  private _handleMousedown = (event: MouseEvent) => {
+    // Prevent the editor from losing focus when the user clicks inside the
+    // autocomplete component. Without this, mousedown transfers focus away from
+    // the editor, carbonStarterTrigger.onTransaction fires, sees
+    // editor.isFocused === false, and dismisses the list before the click event
+    // can trigger onSelect.
+    event.preventDefault();
```
ethanwinters, line 375: The blanket `preventDefault()` on the host also blocks text selection inside the list, since suppressing the mousedown default kills selection along with focus transfer.

Scoping the listener to the item rows would keep the focus fix without that side effect. The explanatory comment above is good and earns its place.

examples/web-components/prompt-line-conversation-starters/src/carbon-icons.d.ts
```diff
@@ -0,0 +1,18 @@
...
+ *  @license
+ */
+
+/*
+ * TypeScript ambient declarations for `@carbon/icons/es/*` icon modules.
+ * These modules export a single default object (a Carbon icon descriptor)
+ * that is compatible with the `icon` field on `ToolbarAction`.
+ */
+declare module "@carbon/icons/es/*" {
+  const icon: any;
```
ethanwinters: `CarbonIcon` is exported from `@carbon/ai-chat-components` and would type-check the `icon` field properly instead of widening to `any`.

ethanwinters (changes requested):
Reviewed the full diff and verified behavior live in both new examples.

The API reshape from `SuggestionItem[]` to `StartersConfig` is clean and consistently threaded through both packages, docs, and tests, and the two examples follow the README/Indexer contract closely. Both run, and the core flow works: starters appear on empty+focused, `renderCustomList` renders the "Prompt suggestions" header, selection auto-sends, and typing correctly dismisses the list.

Requesting changes on three things: a set of accidental npm dependencies, a verified keyboard focus-ring regression, and an `isOn` design whose documented behavior is the opposite of what actually happens. Details are in the inline comments; the last one needs more room, so it's below.

---

## The `isOn` toggle recreates the editor

`StartersConfig.isOn`'s JSDoc promises that keeping the config present with `isOn: false` "leaves the editor intact and lets you toggle the list on and off instantly", and both example READMEs repeat that claim.

I tagged the live ProseMirror node in the React example and clicked the toggle:

```
sameEditorNode: false
recreated: true
```

The editor is destroyed and rebuilt on **every** toggle — exactly what the design says it avoids.

The root cause is in `packages/ai-chat/src/chat/hooks/useInputExtensions.ts`, which this PR only touches at the import, so I couldn't anchor a comment there:

```ts
const normalizedStarters = useMemo(
  () => transformStarterItems(starters),
  [starters],          // <- isOn lives inside this object
);
```

`normalizedStarters` is memoized on the whole `starters` object identity, and `isOn` is a field on that object. Flipping `isOn` produces a new object, which churns the `extensions` array, which trips the editor-recreation path that the same file warns about a few lines further down:

> No carbon bundle: return a stable reference [...] so a `buildCarbonExtensions` load that yields nothing [...] does not churn the prompt-line's `extensions` prop and needlessly recreate the live editor (which would drop its content).

One way out is to key the memo on the fields that actually affect extension construction, letting `isOn` flow through storage without a rebuild:

```ts
const normalizedStarters = useMemo(
  () => transformStarterItems(starters),
  [starters?.items, starters?.renderCustomList],
);
```

...and let the existing storage patch in `AutocompleteController.updateOptions` carry `isOn` on its own.

This is also what makes the focus-ring regression fire: because the editor is recreated on every toggle, `_recreateEditor`'s hardcoded `_focusFromMouse = true` runs every time. Fixing the recreation makes that path rare, but it should be fixed too — see the inline comment.

Either fix the code or fix the docs, but they can't both ship as-is.

---

## Test / verification gaps

- **`isOn` has no test coverage at all.** `autocomplete-controller.test.ts` only migrates existing assertions to the new object shape; nothing exercises the new flag. Worth adding: `isOn: false` suppresses the trigger while empty+focused, toggling `false -> true` re-emits, and `undefined` behaves as `true`.
- **`renderCustomList` on starters is newly enabled and untested.** `_resolveRenderCustomList` used to return `undefined` for the starter trigger and now returns the consumer's callback — no test asserts it reaches `_state.renderCustomList`.
- **Nothing pins the editor-recreation behavior.** A test asserting that the `extensions` array identity is stable across an `isOn` flip would have caught the contradiction described above before it shipped.
- **Manual a11y pass still needed once the focus-ring fix lands**: keyboard-only tab into the editor, arrow through starters, Enter to select, Escape to dismiss, and confirm the focus ring survives a toggle.

## Tracked separately

Two things came out of this review that are follow-ups rather than objections to this PR:

- carbon-design-system/carbon-ai-chat#1963 — `ToolbarAction` has no pressed/selected state, so a toggle built from `input.actions` can only signal on/off via tooltip text.
- carbon-design-system/carbon-ai-chat#1964 — `cds-aichat-autocomplete` options expose `listitem` rather than `option` in the flattened a11y tree, and the input has no combobox relationship attributes. Predates this PR and affects the typeahead examples equally.

Nice work on the docs — `PromptLine.md` and the two example READMEs are unusually clear, and the comment explaining the mousedown fix in `autocomplete.ts` is exactly the kind of non-obvious *why* the repo asks for.

warrenmblood:
@ethanwinters changes pushed

ethanwinters (approved):

Issue #1992, opened 2026-07-27
feat: migrate float examples to launcher component


AlexanderMelox:
## Summary
Can be started after merging #1996 

Update the four float examples to use the new standalone launcher component instead of `ChatButton` / `cds-aichat-button` with the mismatched icon sizing.

## Scope
- Update [`examples/react/custom-element-as-float`](examples/react/custom-element-as-float)
- Update [`examples/react/custom-element-as-float-lazy-load`](examples/react/custom-element-as-float-lazy-load)
- Update [`examples/web-components/custom-element-as-float`](examples/web-components/custom-element-as-float)
- Update [`examples/web-components/custom-element-as-float-lazy-load`](examples/web-components/custom-element-as-float-lazy-load)

## Acceptance criteria
- [ ] The two React float examples use [`Launcher`](packages/ai-chat-components/src/react/launcher.ts)
- [ ] The two web-components float examples use [`cds-aichat-launcher`](packages/ai-chat-components/src/components/launcher/src/launcher.ts)
- [ ] The current float open/close behavior is preserved
- [ ] Affected example builds/tests pass
- [ ] Both package gates still pass after the example updates

## Notes
This is follow-up work to the launcher port. `packages/ai-chat` is out of scope for this issue.

PR #1996, opened 2026-07-27
refactor: port simple launcher component

### LINKED ISSUE #1426: Port the Launcher to @carbon/ai-chat-components
## Summary

Port the launcher UI from `components-legacy` into `@carbon/ai-chat-components` as a Lit element. `Launcher` (`launcher/Launcher.tsx`) is a props-driven, presentational component; `LauncherContainer` (`launcher/LauncherContainer.tsx`) holds the store wiring and stays in `@carbon/ai-chat` **for 1.x only** — #1548 deletes it when 2.0 removes the built-in launcher.

This element is not just an internal tidy-up. From 2.0 the chat renders no launcher of its own (#1548), and **this element is what consumers wire up themselves**. It needs to ship on 1.x, before the built-in launcher goes away, so there is somewhere for consumers to go.

## Scope

- **Port:** `packages/ai-chat/src/chat/components-legacy/launcher/Launcher.tsx`
- **Stays in `@carbon/ai-chat` (1.x):** `LauncherContainer.tsx` — relocates to `components/` and becomes the wiring container around the ported element. It is deleted in 2.0 by #1548.

## Settled direction

The design follow-up that used to block this issue is resolved:

- **No redesign.** The ported element is today's launcher.
- **Simple standalone launcher only.** The "expanded" / call-to-action launcher is not ported. Its config is already deprecated and is removed in 2.0 (#1532, subsumed by #1548).
- **Carbon core tokens drive the colors.** The ported element does not read the `--cds-aichat-launcher-*` custom properties — those are removed with the built-in launcher in #1548.
- **Accessibility.** #1240 raised a11y concerns with the current launcher. Those are a design conversation about this element and about host usage; they do not gate this port.

## Approach

1. Build the Lit element in `@carbon/ai-chat-components` with the full artifact set (element + `.scss`, `index.ts`, Lit + React stories, tests, regenerate `custom-elements.json` + React wrapper).
2. Map colors onto Carbon core theme tokens; leave the `--cds-aichat-launcher-*` custom properties untouched in `@carbon/ai-chat` (#1548 removes them).
3. Relocate `LauncherContainer` into `@carbon/ai-chat`'s `components/`, wiring it to the ported element.
4. Update import sites; remove the legacy `launcher/Launcher.tsx`.

## Acceptance criteria

- [x] The simple launcher renders as a `@carbon/ai-chat-components` element, with Lit and React stories and tests.
- [ ] Colors resolve from Carbon core theme tokens; no dependency on `--cds-aichat-launcher-*`.
- [x] The expanded / call-to-action behavior is not carried into the ported element.
- [ ] `LauncherContainer` drives the element from `@carbon/ai-chat` with no behavior change on 1.x.
- [ ] Legacy launcher file removed from `components-legacy`.
- [ ] Build + tests green for both packages.

## Out of scope

- Removing the built-in launcher, its config, and its view surface — #1548.
- A redesign or new launcher variants.

## Related

- Parent: #1425
- Blocks: #1548 — 2.0 removes the built-in launcher and points consumers at this element.
- Related: #1240 (launcher accessibility / design), #1532 (call-to-action config removal), #1617 (storybook coverage)

AlexanderMelox:
Closes #1426

Port the simple standalone Launcher into `@carbon/ai-chat-components`, add Storybook coverage for both Lit and React usage, and bring over the minimal styling and test coverage needed to make the component usable and reviewable on 1.x. 

> [!IMPORTANT]
> Per follow-up clarification, this PR does not rewire `packages/ai-chat`; it only exposes the standalone launcher component and supporting artifacts. 

#### Changelog

**New**

- Added the standalone Launcher web component in `@carbon/ai-chat-components`.
- Added the generated React wrapper for Launcher.
- Added React Storybook stories and docs for Launcher.
- Added focused Launcher web-component tests for unread badge rendering and toggle event dispatch.

**Changed**

- Applied minimal simple-launcher styling so the standalone component matches the expected visual treatment more closely. This includes adding a block-size of 56px to achieve the new larger icons.

**Removed**

- None in this PR.

#### Testing / Reviewing

- Validate the component package build:
  - `npm run build --workspace=@carbon/ai-chat-components`
- Validate Launcher web-component tests:
  - `npm run test:web-components --workspace=@carbon/ai-chat-components -- src/components/launcher/__tests__/launcher.test.ts`
- Review the Launcher in Storybook:
  - Lit Storybook: `http://localhost:6006`
  - React Storybook: `http://localhost:7007`
- In both Storybooks, verify:
  - Default launcher rendering
  - AI-enabled icon rendering
  - Unread count badge rendering
  - Avatar rendering
  - Toggle interaction/event behavior
- Note: standalone launcher styling in `packages/ai-chat-components/src/components/launcher/src/launcher.scss` was intentionally kept to the minimal simple-launcher port needed for this issue. If any remaining visual polish gaps are found, they can be handled in follow-up work.
- Follow-up example migration work is tracked separately in #1992.


AlexanderMelox:
Alternative for avatar URL is using this component but we need it in string or url format 

https://github.com/carbon-design-system/carbon-ai-chat/blob/65fae83977efbe17065f189e3eb05a4a4fe78acc/packages/ai-chat/src/chat/components-legacy/Avatar.tsx#L25

packages/ai-chat-components/src/components/launcher/src/launcher.ts
```diff
@@ -0,0 +1,189 @@
...
+ * @fires {CustomEvent} cds-aichat-launcher-toggle - Fired when the user clicks the launcher button.
+ */
+@carbonElement(`${prefix}-launcher`)
+class CDSAIChatLauncher extends LitElement {
+  static styles = [commonStyles, styles];
+
+  /**
+   * Hides the launcher and removes it from the tab order when `true`.
+   */
+  @property({ type: Boolean, attribute: 'launcher-hidden' })
```
ethanwinters: I don't think we need this? Seems like a classname they could add on their own.

packages/ai-chat-components/src/components/launcher/src/launcher.ts
```diff
@@ -0,0 +1,189 @@
...
+   */
+  @property({ attribute: 'unread-label' })
+  unreadLabel?: string;
+
+  /**
+   * Value for the `data-testid` attribute on the inner button element.
+   * @internal
+   */
+  @property({ attribute: 'data-test-id' })
+  dataTestId?: string;
```
ethanwinters: Lets remove this

packages/ai-chat-components/src/components/launcher/src/launcher.ts
```diff
@@ -0,0 +1,189 @@
...
+        composed: true,
+      })
+    );
+  }
+
+  /**
+   * Moves focus to the launcher button.
+   */
+  requestFocus() {
+    this.shadowRoot?.querySelector<HTMLElement>('cds-aichat-button')?.focus();
```
ethanwinters: I think this is supposed to return a boolean (false if there was nothing to focus)

packages/ai-chat-components/src/components/launcher/src/launcher.ts
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  render() {
+    const tabIndex = this.launcherHidden ? -1 : undefined;
+    const tooltipPosition = isDirectionRTL()
```
warrenmblood: small thing. We're using `isDirectionRTL` to detect the document text direction and set the Launcher's tooltip position. That works within our chat app because of where the launcher is positioned, but to me this doesn't make sense in the standalone component. You can see in the storybook this makes it so the launcher tooltip always gets cut off.

I think it makes sense to add an `align` property so the tooltip position can be set by the user instead, similar to `IconButton` in core. We can then use `isDirectionalRTL` to set the launcher tooltip in the main chat component.

warrenmblood (changes requested):
Looks good! Just want the tooltip position addressed so it's at least visible in the storybook. See below comment.

packages/ai-chat-components/src/components/launcher/src/launcher.ts
AlexanderMelox: @ethanwinters From the original launcher it looks like it calls a function called `doFocus which does the following`

```
/**
 * Requests focus be moved to the element referenced by the given react ref if the referenced element is defined.
 *
 * @param ref The reference to the element to move focus to.
 * @param defer Indicates if the focus should be executed now or if it should be deferred to another event loop.
 * @param preventScroll Indicates if scrolling should be prevented as a result of the focus change.
 */
```

If we feel the need of not exposing this to the public API we can totally remove just wanted to double check

AlexanderMelox: So it looks like its on a imperative handle for the react parent. If we feel like this was just react glue for the LauncherContainer we can remove it here 👍🏻 

```
 useImperativeHandle(launcherRef, () => ({
    requestFocus: () => {
      doFocusRef(buttonRef);
    },
    buttonElement: () => buttonRef.current ?? undefined,
    containerElement: () => extendedContainerRef.current ?? undefined,
    launcherContainerElement: () => extendedContainerRef.current ?? undefined,
  }));
```

ethanwinters: Yeah lets just remove it

warrenmblood (approved):
LGTM!

PR #2001, opened 2026-07-29
feat: align config on isOn and default the focus shortcut off

### LINKED ISSUE #1916: Honor the documented is_on default for the message focus toggle shortcut
## Background

`ChatShortcutConfig.isOn` has no owner. Three call sites each resolve its default differently, and two carry comments asserting opposite behavior:

| Site | Code | Its own comment |
| --- | --- | --- |
| [AppShell.tsx:645](packages/ai-chat/src/chat/AppShell.tsx#L645) | `shortcutConfig.is_on === true` | "default to **false** if not specified" |
| [MessagesComponent.tsx:732](packages/ai-chat/src/chat/components-legacy/MessagesComponent.tsx#L732) | `shortcutConfig.is_on !== false` | "default to **true** if not specified" |
| [Input.tsx:386](packages/ai-chat/src/chat/components/input/Input.tsx#L386) | `shortcutConfig?.is_on` | none — no default merge at all |

The root cause is that **there is no defaults-merge layer for `keyboardShortcuts`**. `createAppConfig` in [doCreateStore.ts](packages/ai-chat/src/chat/store/doCreateStore.ts) already runs `mergeDefaultsDeep` for `header`, `layout`, and `launcher` and stores the results on `AppConfig.derived` — but `keyboardShortcuts` is passed through raw. Every read site therefore invents its own default, and they disagree.

## The failures

**A screen reader is told to press a key that does nothing.** `ChatShortcutConfig` requires `key` and `modifiers` but not `is_on`, so this is a conforming config:

```ts
keyboardShortcuts: { messageFocusToggle: { key: "F6", modifiers: {} } }
```

The `||` fallback at [AppShell.tsx:641-642](packages/ai-chat/src/chat/AppShell.tsx#L641-L642) substitutes the **whole object**, so it only fires when `messageFocusToggle` is entirely absent. Here it is present, `is_on` stays `undefined`, and `=== true` fails. Meanwhile `MessagesComponent` reads the same config as enabled and picks the detailed scroll-handle label, announcing "Press F6 to toggle between the message list and the input field." F6 is inert.

Swapping `||` for `??` would not fix this. The merge has to be **field by field**.

**The announcement drops modifier keys.** [Input.tsx:387](packages/ai-chat/src/chat/components/input/Input.tsx#L387) passes the bare `shortcutConfig.key`, while [MessagesComponent.tsx:740](packages/ai-chat/src/chat/components-legacy/MessagesComponent.tsx#L740) passes `formatShortcutForDisplay(config)`. A `{ key: "k", modifiers: { ctrl: true, shift: true } }` config announces "Press **k**" on first focus but "Press **Ctrl + Shift + K**" on the scroll handle.

**The test that covers this asserts nothing.** [keyboardShortcuts_spec.ts:573](packages/ai-chat/tests/instance/spec/keyboardShortcuts_spec.ts#L573) is titled "should enable shortcuts by default when is_on is undefined (inherits from DEFAULT which is true)". It builds the exact failing config, fires F6, and closes with `expect(true).toBe(true)`. It is not alone: **16 of the suite's 17 tests** end that way. The suite also fires keys on `document` while the listener is attached to the chat container, and never sets `openChatByDefault`, so the listener is never even attached.

## Goal

One resolver owns the default. Every consumer of the shortcut config agrees on whether it is enabled.

**The default is `false`.** The shortcut is off unless a host opts in with `isOn: true`. This reverses the direction this issue was originally filed with — see [Decisions](#decisions).

## Approach

1. **Add the merge to the layer that already exists for it.** Extend the `mergeDefaultsDeep` call in `createAppConfig` ([doCreateStore.ts](packages/ai-chat/src/chat/store/doCreateStore.ts)) to fold `DEFAULT_MESSAGE_FOCUS_TOGGLE_SHORTCUT` into `AppConfig.derived.keyboardShortcuts`, alongside `header`/`layout`/`launcher`. `mergeDefaultsDeep` skips `undefined` sources, so it merges field by field and an explicit `isOn: false` survives.
2. **Route all three sites through `derived`.** Delete the two contradicting comments and the local `||` fallbacks. No site keeps a default of its own.
3. **Fix the announcement** to pass `formatShortcutForDisplay(...)` instead of the bare `key`, so the announced and displayed shortcut agree about modifiers. The `input_keyboardShortcutAnnouncement` placeholder keeps its `{key}` name — renaming it would break any host overriding that string through `strings`.
4. **Make `key` and `modifiers` optional** on `ChatShortcutConfig`. With the merge in place, `{ isOn: true }` alone resolves to the full default binding, so a host that just wants to turn the shortcut on no longer has to restate `key: "F6", modifiers: {}` — and the demo's hand-rolled backfill in `config-manager.ts`, which duplicated the merge purely to satisfy the type, is deleted.
5. **Rewrite the spec** so it boots an open chat, fires on the container, and asserts real behavior.

### Why not a standalone resolver in `keyboardUtils.ts`

A helper called at each site would still return a **fresh object per call**, which is why the original plan for this issue needed a warning about `mapStateToProps` and `shallowEqual`. `AppConfig.derived` is where this codebase already solves that: the merged object is built once per config replace and passed through `reconcileObjectReferences`, so `MessagesComponent`'s selector keeps a stable identity with no memoization.

## Decisions

Two decisions were taken while scoping the fix, both of which change this issue as originally written:

- **The default is `false`, not `true`.** This issue originally argued for honoring the documented `true` default. The shortcut is now off out of the box, and the JSDoc, `DEFAULT_MESSAGE_FOCUS_TOGGLE_SHORTCUT`, and the demo switcher all say so. Consequences: `input_keyboardShortcutAnnouncement` does not fire in the unconfigured case, and the scroll handles use the `...NoShortcut` label variants. Both now match the behavior instead of contradicting it.
- **`isOn` stays optional.** Rather than making it required to force an explicit choice, `isOn` remains `isOn?: boolean` — consistent with `HeaderConfig`, `LauncherConfig`, `HistoryConfig`, and `HomeScreenConfig`, and with `UploadConfig`/`DisclaimerPublicConfig`, which are loosened to optional in the same change.

## Acceptance criteria

- [ ] One merge resolves the effective shortcut config, and `AppShell`, `MessagesComponent`, and `Input` all read it from `AppConfig.derived`.
- [ ] `{ key: "F6", modifiers: {} }` with no `isOn` leaves the shortcut **disabled**, and the scroll handles omit the shortcut clause — the two agree.
- [ ] A partial config still inherits the default `modifiers` and `isOn` field by field, rather than being replaced wholesale.
- [ ] `isOn: false` still disables it and still suppresses the shortcut clause.
- [ ] `isOn: true` enables the shortcut, the detailed labels, and the first-focus announcement together.
- [ ] The announcement and the scroll handles print the **same** formatted shortcut, modifiers included.
- [ ] `{ isOn: true }` alone resolves to the full default binding, and the demo switcher dispatches just that.
- [ ] `keyboardShortcuts_spec.ts` asserts real behavior throughout — no `expect(true).toBe(true)` remains, events are fired on the chat container, and the chat is actually open.
- [ ] No added re-renders — `selectMessagesState` stays `shallowEqual`-stable.
- [ ] Build and tests pass for both packages.

## Out of scope

- The keyboard navigation redesign — #1209.
- The missing focus indicator on the per-message focus handle, and scroll-handle discoverability. Both belong to the #1209 follow-ups.
- Extending `KeyboardShortcuts` with shortcuts beyond `messageFocusToggle`.
- Collapsing the duplicated `messages_scrollHandle*Detailed` / `...DetailedNoShortcut` strings in `en.json` — tracked on #1538.

## Related

- Found while scoping #1209.
- Ships together with #1995 (the `is_on` → `isOn` rename), which touches the same fields.
- `keyboardShortcuts` reaches the web component via `flattenedPublicConfig.ts`, so verify the fix through that path too.

### LINKED ISSUE #1995: Align UploadConfig and ChatShortcutConfig on `isOn`; drop the snake_case `is_on`
## Background

The public config surface names its on/off toggles `isOn` — `HeaderConfig.isOn`, `HomeScreenConfig.isOn`, `HomeScreenStarterButtons.isOn`, `HistoryConfig.isOn`, `LauncherConfig.isOn`, `DisclaimerConfig.isOn`.

Two config toggles spell it `is_on` instead:

| Field | File |
| --- | --- |
| `UploadConfig.is_on` | `packages/ai-chat/src/types/config/UploadConfig.ts` |
| `ChatShortcutConfig.is_on` | `packages/ai-chat/src/types/config/ShortcutConfig.ts` |

Neither is wire format. Both are host-authored config, and every field beside them is already camelCase — `maxFileSizeBytes`, `maxFiles`, `onFileUpload` on `UploadConfig`; `modifiers`, `messageFocusToggle` on the shortcut config. The snake_case is drift, not a contract.

Both interfaces are `@experimental`, so this does not have to wait for the 2.0 window and does not need a deprecation cycle. Fixing it now is cheaper than fixing it after the experimental tag comes off.

## Goal

```ts
// packages/ai-chat/src/types/config/UploadConfig.ts
export interface UploadConfig {
  /**
   * Whether file upload is enabled. When `true`, the chat renders a file attachment
   * button in the prompt line. Defaults to `false`.
   */
  isOn: boolean;
  // ...
}

// packages/ai-chat/src/types/config/ShortcutConfig.ts
export interface ChatShortcutConfig {
  /**
   * Whether the keyboard shortcut is enabled. Defaults to true.
   */
  isOn?: boolean;
  // ...
}
```

Straight rename. No behavior change, no default change, no alias — `is_on` is gone.

## Acceptance criteria

- [ ] `UploadConfig.is_on` is `UploadConfig.isOn`; `ChatShortcutConfig.is_on` is `ChatShortcutConfig.isOn`.
- [ ] `DEFAULT_MESSAGE_FOCUS_TOGGLE_SHORTCUT` in `ShortcutConfig.ts` uses `isOn`.
- [ ] Consumers follow: `ChatActionsImpl.ts` (upload gate), `MessagesComponent.tsx` (shortcut gate), and the demo switchers / `config-manager.ts` / `utils.ts`.
- [ ] Examples follow: `examples/react/prompt-line-file-upload` (`App.tsx` + `README.md`) and the `upload.is_on` rows in `examples/react/README.md`.
- [ ] Docs follow: `docs/PromptLine.md`, `docs/StructuredData.md`, and the `@link`ed JSDoc that names `is_on` in prose.
- [ ] Tests follow, including `tests/instance/spec/uploadUI_spec.ts`.
- [ ] TypeDoc builds clean; the generated `docs/api/markdown/UploadConfig.md` shows `isOn`.
- [ ] Release note calls out the rename on both `@experimental` interfaces.

## Out of scope

**The `is_on` fields in message payloads stay as they are.** `message_item_options.feedback.is_on` and `custom_footer_slot.is_on` (`packages/ai-chat/src/types/messaging/Messages.ts`) are server response options. snake_case is the correct convention on the wire format, and renaming them would be a real breaking change to what a backend sends. This issue touches host-authored config only.

Also out of scope:

- `InputConfig.isVisible` → `isOn`, and the `InputConfig` → `PromptLineConfig` rename around it. Both ride with #1701 in the 2.0 window, because they are breaking on non-experimental API.
- Any behavioral change to file upload or keyboard shortcuts.

## Related

- Sibling naming work in the 2.0 window: #1701
- Toggle convention this aligns to: `HeaderConfig.isOn`, `HistoryConfig.isOn`, `LauncherConfig.isOn`

ethanwinters:
Closes #1995
Closes #1916

Renames the last two config toggles still spelled `is_on`, and gives the message focus toggle shortcut one owner for its default. The two ship together because they touch the same fields.

The shortcut default had no owner. `AppShell` read it as `=== true`, `MessagesComponent` as `!== false`, and `Input` applied no default at all — so a conforming `{ key: "F6", modifiers: {} }` left the handler off while the scroll handle told a screen reader the key worked. Root cause was a missing merge: `createAppConfig` already merged `header`, `layout`, and `launcher` into `AppConfig.derived` but passed `keyboardShortcuts` through raw.

Worth a closer look:

- `packages/ai-chat/src/chat/store/doCreateStore.ts` — the new merge. It relies on `mergeDefaultsDeep` skipping `undefined` sources, which is what keeps an explicit `isOn: false` from being overwritten by the default.
- `packages/ai-chat/tests/instance/spec/keyboardShortcuts_spec.ts` — full rewrite, 755 lines changed. The old suite asserted nothing.

#### Changelog

**Changed**

- `UploadConfig.is_on` → `UploadConfig.isOn`. `@experimental`.
- `ChatShortcutConfig.is_on` → `ChatShortcutConfig.isOn`. `@experimental`.
- The message focus toggle shortcut now defaults to **off**. Opt in with `isOn: true`.
- `isOn` is optional on `UploadConfig` and `ChatShortcutConfig`. `@experimental`.
- `DisclaimerPublicConfig.isOn` is optional. Not `@experimental` — this one is a public API change, additive and source-compatible: every config that compiled before still compiles, and an omitted `isOn` reads as off, same as `false`.
- `key` and `modifiers` are optional on `ChatShortcutConfig`, so `{ isOn: true }` alone resolves to the default F6 binding.
- The chat warns at startup when `upload.onFileUpload` is set without `isOn`. An omitted `isOn` reads as off, and the type no longer catches it.
- The first-focus announcement formats modifiers. `Ctrl+Shift+K` announced "Press k" before.
- `formatShortcutForDisplay` returns the modifiers alone for a config with no `key`. It rendered a trailing separator — `Ctrl + ` — that a screen reader announces verbatim.
- Shortcut defaults merge once into `AppConfig.derived`; `AppShell`, `MessagesComponent`, and `Input` all read that.

**Removed**

- The demo's `messageFocusToggle` backfill and its hardcoded `"F6"`. The library supplies both now.

Wire-format `is_on` on message payloads is untouched — `message_item_options.feedback.is_on` and `custom_footer_slot.is_on` stay snake_case.

Also included, unrelated but small:

- `.eslintignore` was missing the storybook output dirs, so any build left 21,668 lint errors in generated files.
- Two non-null assertions the linter warned on.
- A stale `is_on` in the homescreen JSDoc.

#### Testing / Reviewing

`npm run aiChat:start`

**Shortcut default is off, and everything agrees**

1. Open the chat and send a message. Press F6 — nothing happens.
2. Tab to the scroll handle above the transcript. It reads "Press escape to exit the message list." with no mention of F6.
3. Check "Enable F6 keyboard shortcut" in the sidebar.
4. Press F6 — focus moves between the transcript and the input. The scroll handle now ends "Press F6 to toggle between the message list and the input field."

Before this change, step 2 named a key that did nothing whenever a host passed a partial config.

**Upload**

Set "File uploads" to on in the sidebar; the attach button appears. Config now spells it `upload.isOn`.


packages/ai-chat/tests/store/spec/dynamicConfigUpdates_spec.ts
```diff
@@ -84,6 +84,61 @@ describe('Dynamic Config Updates', () => {
...
+      expect(
+        serviceManager.store.getState().config.derived.keyboardShortcuts
+          .messageFocusToggle.isOn
+      ).toBe(true);
+
+      await applyConfigChangesDynamically(
+        serviceManager.store.getState().config.public,
+        {},
+        serviceManager
+      );
```
AlexanderMelox, line 164: Caught a logic bug here, can you verify?

The second `applyConfigChangesDynamically` call passes` {} `as newConfig. `applyConfigChangesDynamically` calls `createAppConfig(newConfig) `directly — it does not merge onto the previous config. So `createAppConfig({})` returns the factory default (isOn: false) and the assertion passes trivially, regardless of whether the merge logic can actually retract an earlier isOn: true.

packages/ai-chat/src/chat/utils/keyboardUtils.ts
```diff
@@ -91,7 +97,8 @@ export function formatShortcutForDisplay(config: ChatShortcutConfig): string {
   }
 
   // Add the key (capitalize single letters, keep special keys as-is)
-  const key = config.key.length === 1 ? config.key.toUpperCase() : config.key;
+  const configKey = config.key ?? '';
```
AlexanderMelox, line 102: **Problem**: `formatShortcutForDisplay` returns a dangling separator when key is absent

`config.key ?? ''` guards the assignment, but `parts.push(key) `still pushes `''`, so `formatShortcutForDisplay({ modifiers: { ctrl: true } })` produces "Ctrl + ". 

A screen reader would announce the trailing separator verbatim. In production every call goes through `Required<ChatShortcutConfig>` so key is always present — but the function's signature accepts the broader `ChatShortcutConfig`, so the gap is real for any other caller. 

**Suggested fix:**
```
const configKey = config.key ?? '';
if (!configKey) {
  return parts.join(' + ');
}
const key = configKey.length === 1 ? configKey.toUpperCase() : configKey;
parts.push(key);
```

AlexanderMelox (changes requested):
Looks good just two important things and a minor change:

Important:
- Logic bug in the "turn off at runtime" test
- formatShortcutForDisplay returns a dangling separator when key is absent

Small change: 
- `DisclaimerPublicConfig.isOn` optionality is an unlisted public API change
DisclaimerPublicConfig carries no @experimental tag. Changing isOn: boolean → isOn?: boolean is additive and safe, but it's not mentioned in the PR's changelog. Worth calling it out explicitly there since it's a non-experimental public type.

AlexanderMelox (approved):

Issue #1543, opened 2026-06-03
[Epic]: 2.0.0 - Carbon 12


ethanwinters:
PLACEHOLDER INFO:

## Carbon 12 Timeline

| Date | Gate | What we do |
| --- | --- | --- |
| **Sep 10, 2026** | Alpha (private preview) | Get access. Throwaway branch. Bump and see. |
| **Nov 5, 2026** | Beta — MVP components complete | **Our real gate.** At this point we are all in on Carbon 12. We need to bring all our issues to the Carbon team ASAP |
| **Jan 2027** | RC — API frozen, Upgrade Assistant ships | We should start shipping alphas, quietly. |
| **Feb 2027** | RC — stable API, Design Harness ships | Design validates the chat UI against the new look and feel. We should be announcing public alphas/betas by now. |
| **Mar 2027** | **GA** — `@carbon/web-components@3.0.0` | Ship both packages at 2.0.0. |
| **May 2027** | Think — new look and feel unveiled | Product teams will ask when chat looks like that. |

**One planning caution:** alpha to GA in six months is aggressive. Carbon's v11 prerelease ran eight months across four betas. Plan against these dates, but don't make a downstream commitment that only works if every one of them holds.

---

## Where we stand

After everything now planned for 2.0.0, we render **22 Carbon components in 75 places**.

Our exposure is concentrated around **`icon-button`, `button` and `overflow-menu`.** These are also the elements changing to being rounded that can have some new padding and spacing fallout, so I would assume that all the elements that use those are at risk of needing re-working.

**Our published elements:**

| Our element | Carbon components it renders | Carbon helpers it uses | In 2.0.0 |
| --- | --- | --- | --- |
| `feedback` | `button`, `checkbox`, `icon-button`, `tag`, `textarea` | icon loader ×2 | — |
| `table` | `data-table`, `pagination`, `select`, `icon-button` | icon loader | Moves under `markdown` |
| `card` | `button`, `icon-button`, `icon-indicator`, `loading` | icon loader ×2 | **Removed in favor of native tile/card** |
| `chat-history` | `icon-button`, `overflow-menu`, `search`, `skeleton-text` | **16** — icon loader ×6, focus and host-listener mixins ×9, spread directive | — |
| `code-snippet` | `button`, `skeleton-text` | icon loader, focus mixin | — |
| `chat-shell` | `icon-button`, `overflow-menu` | icon loader | — |
[...]

PR #2002, opened 2026-07-29
fix(workspace-shell): audit fixes for toolbar and header

### LINKED ISSUE #1503: [Workspace shell] Dev: Audit fixes
Link to [Workspace shell] Design: Audit fixes #1504 for reference

AlexanderMelox:
Closes #1503

Addresses design audit fixes for the workspace-shell component, including layout corrections to the header, toolbar title styling, and chat-button quick action styles.

#### Changelog

**New**

- ~~Added `workspace` boolean property to `cds-aichat-toolbar` that applies `heading-03` type style to the toolbar title when used inside a workspace-shell~~ — superseded, see **Changed** below; the toolbar no longer takes a new attribute
- Added `--cds-aichat-toolbar-title-padding-inline-start` custom property to `cds-aichat-toolbar` (defaults to `$spacing-04`) so the workspace-shell can override the title indent from outside the shadow boundary
- Added `lowContrast` to the demo notification in `WorkspaceWriteableElementExample`

**Changed**

- `workspace-shell-header`: switched to `display: flex` layout with `flex-direction: column` (row on `md+`), added `align-items: flex-start`, and `gap: $spacing-05`
- `workspace-shell-header`: moved `header-action` slot inside `.header-content` for both collapsible and non-collapsible variants so it aligns correctly within the flex container
- `workspace-shell-header`: added `align-items: start` and `inline-size: 100%` to `.header-content`, and `inline-size: 100%` to the title row
- ~~`workspace-shell`: moved `::slotted([slot='notification'])` rule to host level and removed `!important` on `max-inline-size`; added `margin-block-end: 0`~~ — the `!important` removal was a regression, restored in `65368459` (see below). The move to host level and `margin-block-end: 0` stand.
- `workspace-shell`: restored `max-inline-size: 100% !important` on the slotted notification, with a comment explaining why it is load-bearing. Carbon caps inline notifications at a responsive 18/38/46/52rem ladder. In the React build that rule targets a light-DOM `div.cds--inline-notification` from the **document** tree, which beats `::slotted()` regardless of specificity — measured on the deploy preview, the notification rendered **736px inside a 1196px shell**, and raising the selector's specificity did not move it. `!important` restores it to 100% (1194px). The web components build is unaffected either way: there Carbon applies the same ladder via `:host(cds-inline-notification)` **inside** the notification's shadow root, which our outer-tree `::slotted()` already overrides without the flag.
- ~~`toolbar`: changed `padding-inline` to `padding-inline-start: $spacing-07` on the title element~~ — superseded by the two bullets below
- `workspace-shell`: the existing `::slotted(cds-aichat-toolbar)` rule now applies `heading-03` and sets `--cds-aichat-toolbar-title-padding-inline-start: $spacing-07`. Font properties are inherited, so the type style reaches the toolbar title across the shadow boundary on its own; padding is not inherited, hence the custom property. Carbon's buttons, tooltips, and overflow menu all set their own type styles, so nothing else inside the toolbar is affected.
- `toolbar`: split `padding-inline: $spacing-04` on the title into `padding-inline-end: $spacing-04` plus a `padding-inline-start` that reads the new custom property. The existing "zero the start padding when the navigation slot is filled" rule still takes precedence over it.
- `chat-button` quick-action: changed border from `$chat-button` token to `$border-interactive`; fixed property order (`border-width` before `border-color`) on `:focus`; reformatted multi-line calc expressions to single line
- ~~Stories: updated `QuickAction` story default to `isSelected: true`; added `workspace` prop to workspace-shell toolbar story slots~~
- Stories: updated `QuickAction` story default to `isSelected: true`. The workspace-shell toolbar story slots need no extra prop — the shell styles the slotted toolbar itself.

**Removed**

- ~~N/A~~
- Removed the `workspace` boolean property from `cds-aichat-toolbar`, along with its `:host([workspace])` style block, its three attribute-reflection tests, the workspace-shell test asserting the attribute, and the prop from both Storybook stories and both demo files (React and web-components)

#### Testing / Reviewing

1. Open the workspace-shell demo (`WorkspaceWriteableElementExample`) in both React and web-components variants
2. ~~Verify the toolbar title renders with `heading-03` sizing when `workspace` attribute is set~~ Verify the toolbar title renders with `heading-03` sizing inside the workspace-shell with **no** attribute set, and that a toolbar outside a workspace-shell is unchanged. Measured on the deploy preview for `b0656978` (`components-workspace-shell--default`): the title computes to `font-size: 20px`, `line-height: 28px`, `font-weight: 400`, `padding-inline-start: 32px`, with `workspace` absent from the element. On `components-toolbar--default` (outside any shell) it is `16px`/`16px` with `padding-inline-start: 12px`, and the custom property is unset so it falls back to `$spacing-04`.
3. Verify the `header-action` slot content appears correctly aligned inside the header content area on mobile (stacked) and desktop (row) layouts
4. Verify the warning notification in the demo no longer has a heavy contrast background
5. Verify the quick-action chat-button border uses the correct interactive border token and the focus ring uses `border-width: 2px`
6. Run `npm test --workspace=packages/ai-chat-components` — toolbar and workspace-shell tests should pass


packages/ai-chat-components/src/components/workspace-shell/src/workspace-shell-header.scss
```diff
@@ -10,20 +10,24 @@
...
   padding: rem(24px) rem(32px);
   border-block-end: solid rem(1px) theme.$chat-bubble-border;
   color: theme.$text-secondary;
+  gap: spacing.$spacing-05;
 
   @include type.type-style('body-01');
 
-  @include breakpoint-down('md') {
-    gap: 16px;
-    grid-auto-flow: row;
```
AlexanderMelox, line 26: This was dead code that wasn't being used

packages/ai-chat-components/src/components/toolbar/src/toolbar.scss
```diff
@@ -100,3 +101,9 @@ $block-class: #{$prefix}-toolbar;
     min-inline-size: layout.$spacing-08;
   }
 }
+
+:host([workspace]) {
```
ethanwinters, line 105: oh hmm, interesting. I wonder if the move in 2.0.0 will be to make the title a slot vs having to do this? Can't change it in 1.x without it breaking, but I think its worth making an issue to investigate under https://github.com/carbon-design-system/carbon-ai-chat/issues/1543. You mind adding that issue @AlexanderMelox

ethanwinters (dismissed):

ethanwinters (dismissed):
Hmm... it looks like on the chromatic snapshot that all uses of toolbar got an extra padding/margin start on the title. I don't think that was the intent?

packages/ai-chat-components/src/components/toolbar/src/toolbar.scss
AlexanderMelox: issue created https://github.com/carbon-design-system/carbon-ai-chat/issues/2015

AlexanderMelox:
~~@ethanwinters this was one of the audit fixes~~ I see what you mean, double checking if the intent was for all toolbar or just workspace
Screenshot 2026-07-31 at 11 54 45 AM

demo/src/react/WorkspaceWriteableElementExample.tsx
```diff
@@ -146,6 +146,7 @@ function WorkspaceWriteableElementExample({
         slot="toolbar"
         actions={toolbarActions}
         overflow
+        workspace
```
ethanwinters, line 149: I wonder if a "workspace" attribute is the right call here or if its too feature specific. Is there a better way to say "big mode"?

ethanwinters (commented):
Looks good, but lets chat this one out in slack or guild.

packages/ai-chat-components/src/components/toolbar/src/toolbar.scss
```diff
@@ -100,3 +101,11 @@ $block-class: #{$prefix}-toolbar;
...
+  .#{$block-class}__title {
+    @include type.type-style('heading-03');
+
+    padding-inline-start: layout.$spacing-07;
```
AlexanderMelox: Let's move this to workspace shell styles instead

AlexanderMelox: Move all the styles into workspace shell and just override it there, and then make an issue on making titleText a slot and add to this epic: https://github.com/carbon-design-system/carbon-ai-chat/issues/1543

AlexanderMelox: Issue opened here: https://github.com/carbon-design-system/carbon-ai-chat/issues/2230

packages/ai-chat-components/src/components/workspace-shell/src/workspace-shell.scss
```diff
@@ -67,7 +61,19 @@
   }
 }
 
+::slotted([slot='notification']) {
+  box-sizing: border-box;
+  margin-block-end: 0;
+  max-inline-size: 100%;
```
warrenmblood:
```suggestion
  /* stylelint-disable-next-line declaration-no-important */
  max-inline-size: 100% !important;
```

warrenmblood: Without the `!important` flag, `max-inline-size` may get overridden by styles coming from the core component. Noticed this in the React "Default" story:
Image

warrenmblood (changes requested):

packages/ai-chat-components/src/components/workspace-shell/src/workspace-shell.scss
AlexanderMelox: Hey! Is this something we are okay with? I thought we generally want to avoid important

AlexanderMelox: I reverted the important since it was there before this issue was solved so we can leave as is and come back to it later

warrenmblood (approved):
LGTM!

warrenmblood (approved):

PR #2003, opened 2026-07-29
refactor: consolidate autocomplete

### LINKED ISSUE #1964: Fix listbox/option ARIA wiring in cds-aichat-autocomplete
## Background

`cds-aichat-autocomplete` renders a combobox-style suggestion list, but the ARIA relationships that make one usable are broken by shadow-DOM boundaries. This predates PR #1936 and affects every surface that uses the component — typeahead, mention, command, and now starters.

Two structural problems:

**1. The listbox owns no options.** `autocomplete.ts:493-498` renders `<ul role="listbox">` in the `cds-aichat-autocomplete` shadow root, but its children are `<cds-aichat-autocomplete-item>` host elements. The actual `<li role="option" aria-selected id="${item.id}--option">` lives one shadow boundary deeper, in each item's own shadow root (`autocomplete-item.ts:163-170`). `role="listbox"` requires owned `option` children, and the boundary breaks that ownership.

Verified in Chrome against `examples/react/prompt-line-conversation-starters` on the flattened a11y tree:

```
listbox "Autocomplete options" orientation="vertical"
  listitem level="1"     ← no option role, no accessible name
  listitem level="1"
  listitem level="1"
  listitem level="1"
```

The `role="option"` and `aria-selected` set inside the item shadow roots never reach the tree, and the options expose no accessible name.

**2. The input is not wired to the listbox.** The ProseMirror editable that owns focus has none of the combobox relationship attributes. Verified live — all null:

```
role: "textbox", aria-expanded: null, aria-controls: null,
aria-activedescendant: null, aria-haspopup: null
```

So a screen reader user gets no signal that a popup exists, no way to know it opened, and no announcement of the active option as they arrow through it. The per-item ids exist but are unreferenceable across shadow roots, and the popup renders in a different tree from the editor entirely, so `aria-controls` would have to cross a boundary too.

Partial existing mitigation: a polite live region announces "N suggestions. Use up and down arrows to move, Enter to pick, Escape to close." on open. That covers discovery, but not per-option state as the user navigates.

## Goal

The suggestion list exposes a correct listbox/option structure with named options and active-option tracking, so screen reader users get the same information sighted users get from the visual highlight.

## Acceptance criteria

- [ ] Options expose `role="option"` with an accessible name in the flattened a11y tree, as direct owned children of the `listbox`.
- [ ] The active option exposes `aria-selected="true"`, and exactly one option is selected at a time.
- [ ] Arrowing through the list moves the active option and a screen reader announces the newly active option's label.
- [ ] The focused input conveys that a popup is available and whether it is currently open.
- [ ] Behavior holds for all four trigger types — typeahead/autocomplete, mention, command, starters — and for both the built-in list and a consumer `renderCustomList`.
- [ ] Grouped items (`autocomplete-item-group.ts:111`, `role="group"`) still expose their group label correctly under the new structure.
- [ ] Existing keyboard behavior is unchanged: Up/Down moves, Enter picks, Escape closes, focus stays in the editor throughout.
- [ ] WTR tests in `packages/ai-chat-components/src/components/autocomplete/__tests__/` assert the roles, names, and active-option tracking.
- [ ] Verified with NVDA and VoiceOver on the `prompt-line-typeahead` and `prompt-line-conversation-starters` examples.

## Approach notes

Not prescribing an implementation, but the shadow boundary is the crux and there are two broad directions:

- **Flatten the options into the listbox's shadow root** so `<li role="option">` are real DOM children of `<ul role="listbox">` and their ids are addressable from within that root. Most direct fix for problem 1, but `cds-aichat-autocomplete-item` currently owns per-item rendering (avatar, typed/remainder label split, send button), so this is a real refactor of that split.
- **Keep the current element split and reach for `aria-owns` / a delegated id strategy.** Lighter-touch, but `aria-owns` across shadow roots has uneven AT support and tends to be fragile.

Problem 2 is partly constrained by the popup rendering outside the editor's tree — worth checking whether the existing live-region announcer can be extended to cover active-option changes if a true `aria-activedescendant` relationship proves impractical. `AriaAnnouncerManager` (`packages/ai-chat-components/src/globals/utils/aria-announcer-manager.ts`) is the sanctioned path per `packages/ai-chat-components/AGENTS.md`; don't hand-roll live regions.

## Out of scope

- The `ToolbarAction` pressed-state gap (separate issue).
- The keyboard focus-ring regression in `prompt-line-rich-runtime.ts`, which is being handled as a blocker on #1936.
- Visual/styling changes to the list.

## Related

- PR #1936 — the review that surfaced this, though the gap predates it.
- `packages/ai-chat-components/src/components/autocomplete/src/autocomplete.ts:493-498` — listbox render.
- `packages/ai-chat-components/src/components/autocomplete/src/autocomplete-item.ts:158-170` — option render.
- `references/accessibility.md` — repo WCAG 2.1 AA checklist and announcer patterns.
- APG combobox pattern: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/

davidmenendez:
Closes #1964 

consolidates `autocomplete-item` and `autocomplete-item-group` into `autocomplete` in order to alleviate accessibility issues that arise from the shadow dom boundary. by moving the previously componentized code back into `autocomplete` we avoid issues where aria attributes are unable to properly communicate with each other.

as part of the refactor there's also some cleanup being done to remove code that's become unnecessary due to the restructuring and to optimize code to reduce redundancy.

#### Changelog

**Changed**

- `packages/ai-chat-components/src/components/autocomplete/src/autocomplete.ts` consolidated code from item and groups into main component

**Removed**

- TBD

#### Testing / Reviewing

run autocomplete from the storybook and the app. verify that it's working as intended and there are no regressions in functionality. then verify that it's working as intended with voiceover.

note: i haven't consolidated the tests yet or deleted old files. I will do that after preliminary review of the code is done and successful.


davidmenendez:
@warrenmblood made the adjustments to the tests since we confirmed that the consolidated code is working as intended

packages/ai-chat-components/src/components/card/__stories__/card.stories.js
warrenmblood, line 1: Are this and `packages/ai-chat-components/src/components/card/__stories__/placeholder.png` related to this PR?

warrenmblood (changes requested):
Looks good. Didn't find any autocomplete regressions. Just noticed a few props missing from the Component API table in the React storybook overview page. These should be added to `packages/ai-chat-components/src/components/autocomplete/__stories__/autocomplete-react.stories.jsx` (starting on line 148):

```
export default {
  title: 'Preview/Autocomplete',
  argTypes: {
    items: {
      control: false,
      description:
        'Array of suggestion items to display.',
    },
    groups: {
      control: false,
      description:
        'Array of grouped suggestion items to display. These will be displayed after any provided `items`.',
    },
    headerConfig: {
      control: false,
      description:
        'Optional header configuration object with `showHeader` and `title` properties.',
    },
    i18n: {
      control: false,
      description:
        'Localized strings for announcements and labels. Defaults to English via `defaultAutocompleteI18n`.',
    },
    // ...existing argTypes
```

packages/ai-chat-components/src/components/card/__stories__/card.stories.js
davidmenendez: placeholder was added to fix a flakey test that i saw fail in the CI

warrenmblood (approved):
LGTM

PR #2011, opened 2026-07-30
fix(chat-button): audit quick-action states styles; adds selected and active styles

### LINKED ISSUE #1491: [Chat button] Dev: Audit fixes
Link to [Chat button] Design: Audit fixes #1492 for reference

AlexanderMelox:
Closes #1491

Audits the `cds-aichat-button` quick-action visual states to align with Carbon token standards, and adds the missing disabled "selected" state.

#### Changelog

**New**

- Added `QuickActionSelected` Storybook story to demonstrate the disabled selected state for quick-action buttons.

**Changed**

- Replaced custom `chat-button` design tokens (`theme.$chat-button`, `theme.$chat-button-hover`, `theme.$chat-button-selected`, etc.) with standard Carbon tokens (`theme.$border-interactive`, `theme.$link-primary`, `theme.$background-hover`, `theme.$background-active`, `theme.$background-selected`, `theme.$text-secondary`) in `chat-button.scss`.
- Disabled border now uses `button-tokens.$button-disabled` instead of `theme.$border-disabled` for consistency with Carbon button internals.
- Removed the `.--btn--selected` CSS class path; selected state is now represented by setting `disabled` when `isSelected` is true via `_normalizeButtonState`. The btn--selected was dead code since it was never added to the DOM.

**Removed**

- Removed the now-redundant `.--btn--selected` CSS rules (border, background, focus ring overrides) from `chat-button.scss`.
- Removed the `&:not(.--btn--selected)` guard from the `[disabled]` style block — all disabled quick-action buttons share the same base disabled style, with the selected variant overridden via the `[isselected]` host attribute selector.

#### Testing / Reviewing

1. Open Storybook and navigate to **Chat Button → Quick action (selected)** — the button should appear in a selected/disabled visual state with `background-selected` background and `text-secondary` colour.
2. Verify the standard **Quick action** story still shows the bordered ghost style using `border-interactive` and `link-primary` colour.
3. Hover, focus, and active states on an unselected quick-action button should use standard Carbon tokens (no custom `chat-button-*` tokens).
4. Confirm a disabled (non-selected) quick-action button shows `button-disabled` border and `text-disabled` colour.
5. Run `npm test` in `packages/ai-chat-components` to confirm no regressions.


kiran-ravindra-ibm:
States themselves look good. I noticed one potential bug: if I toggle on "selected" and "disabled", but then toggle off only "disabled", the button appears to toggle "selected" off as well until I toggle off "selected" and toggle it back on again (and vice versa). Recording attached.

https://github.com/user-attachments/assets/6eca542f-65cb-45f8-8c14-3f9d3bdded57

AlexanderMelox:
This is what selected looks like in practice in our demo

**Before:**
Screenshot 2026-08-03 at 3 54 59 PM

**After:**
Screenshot 2026-08-03 at 3 53 05 PM

AlexanderMelox:
@ethanwinters I was looking at places where we need to "dogfood" our API's and found that`HomeScreen.tsx` maps over `starters.buttons `but never reads `starter.isSelected`, the config type supports it and the test uses it, but it's not forwarded to the ChatButton as isSelected. If the intent is for previously-clicked starter buttons to render in the selected (non-interactive) state, that prop needs to be wired up there.

Screenshot 2026-08-03 at 4 16 50 PM

The other place is where we expected being used inside of the Option message type component (like the screenshots above)

packages/ai-chat-components/src/components/chat-button/src/chat-button.scss
```diff
@@ -71,18 +67,11 @@
     }
 
     &:active:not([disabled]) {
-      background: theme.$chat-button-active;
-    }
+      @include type.type-style('body-compact-01');
```
warrenmblood:
```suggestion
```

Typography doesn't change for the active state, so applying this is unnecessary. You can also remove the import on line 10 since it is only used here.

packages/ai-chat-components/src/components/chat-button/src/chat-button.scss
```diff
@@ -2,14 +2,16 @@
  *  Copyright IBM Corp. 2025, 2026
  *
  *  This source code is licensed under the Apache-2.0 license found in the
- *  LICENSE file in the root directory of this source tree.
+ *  Apache-2.0 license found in the LICENSE file in the root directory of this source tree.
```
warrenmblood:
```suggestion
 * LICENSE file in the root directory of this source tree.
```

warrenmblood (changes requested):
Looks good. Aside from 2 minor changes below, since the `isSelected` control is already set up in the "Quick action" story, we may want to use the "Quick action (selected)" story to show the selected button in context within a list of disabled ones:

Image

warrenmblood (approved):
LGTM

PR #2014, opened 2026-07-31
feat: render file attachments in the user message bubble

### LINKED ISSUE #1990: Render file attachments in the user message bubble from `structured_data`
### Background

Files uploaded through `UploadConfig.onFileUpload` become `file`-typed `StructuredField` entries on `input.structured_data`. The chat never renders them: they appear as chips in the input, are cleared on send (`CLEAR_STRUCTURED_DATA`), and the user's bubble then shows text only. The attachment is invisible both live and after a history reload.

The only file rendering that exists is the human-agent path, which mints one `MessageRequest` per file with an internal `message_type: "file"`. That enum is exported from neither `aiChatEntry.tsx` nor `serverEntry.ts`, the per-file requests never reach the event bus, and the chat persists no conversation text — so a host cannot practically reproduce those messages through `customLoadHistory`.

### Goal

Render `file`-typed structured fields in the user bubble, so uploaded files are visible when sent and when restored from history.

`structured_data` is already an ordinary part of `MessageRequest`, so it round-trips through `HistoryItem` with no casts and no new persistence. History restore falls out of the rendering rather than being separate work.

### Acceptance criteria

- [ ] The user bubble renders one chip per `structured_data.fields` entry with `type === "file"`.
- [ ] Chips render from an `ExternalFileReference` (name / mime_type / size) with no `File` present — i.e. after a history reload.
- [ ] An `InlineFile` restored from history renders name-only. A raw `File` cannot serialize into a `HistoryItem`; this asymmetry is documented, not fixed.
- [ ] No download link is rendered from `ExternalFileReference.url` — deliberately deferred, can be added opt-in later.
- [ ] Docs: `StructuredData.md` states what the chat renders and the inline-vs-reference restore behaviour; `CustomHistory.md` documents the file round-trip.
- [ ] Tests cover restoring a request carrying file fields from history. No such coverage exists today.
- [ ] An example demonstrating uploaded files surviving a history reload.

### Public API / contract

Additive. `structured_data` gains rendering behaviour it did not have. This raises the cost of changing `StructuredField` later — worth deciding whether `structured_data` stays `@experimental` once the chat renders it.

**Open design question, settle first:** `cds-aichat-file-upload-item` requires a real `File` (it calls `URL.createObjectURL` for previews). Rendering from an `ExternalFileReference` needs a reference mode on that component or a sibling. This gates the visual design. Possibly related to #1839.

### Out of scope

The human-agent path keeps its existing bubble for now. Download links.

### Related

- #1989 — removes the vestigial `InlineFile.status` / `InlineFile.error` fields.
- #1839, #1987 — adjacent message-bubble work.

ethanwinters:
Closes #1990

Files uploaded through `onFileUpload` land on `input.structured_data`. Nothing rendered them. The attachment vanished when the message was sent, and again after a reload. A sent message now renders one read-only chip per `file` field, below its bubble.

No new persistence. `structured_data` is an ordinary part of a request, so history round-trips it already.

Sending a file with no caption also works now. The Send control and the send path read one predicate, instead of disagreeing.

Worth a closer look:

- `file-upload-item.ts` — the chip now takes `FileUpload | FileAttachment`. `_resolved` is where the two shapes meet. `_syncInjectedStyles` patches Carbon's shadow root for read-only mode.
- `MessageTypeComponent.tsx` — the bubble's text-item gate is relaxed. A request with only attachments renders no bubble, rather than an empty one.

#### Changelog

**New**

- File chips in the user's message bubble, live and after a history restore.
- `read-only` and `fallback-label` on `cds-aichat-file-upload-item`, plus a `FileAttachment` shape for a file with no live `File`.
- `history-file-attachments` example, React and web components.
- `MessageFileAttachments` component and `getMessageFileAttachments` helper.

**Changed**

- Sending a file with no text works.
- The Send control stays disabled while a file uploads, matching `doSend`.
- An empty send is silent only when it has no attachments either.
- `StructuredData.md` and `CustomHistory.md` cover what a file field renders, and the history round trip.

#### Testing / Reviewing

Demo (`npm start --workspace=@carbon/ai-chat-examples-demo`):

1. Sidebar → Input config → **Enable file uploads**.
2. Attach a file. Send with no text. You get a chip, not an empty bubble.
3. Attach an image and send. The chip shows a thumbnail. A PDF or CSV shows its file-type icon.
4. Attach a file and press Send mid-upload. The control is disabled until the upload lands.

Reload path: `npm run start --workspace=@carbon/ai-chat-examples-react-history-file-attachments`, then reload the page. The chip comes back from `structured_data` alone.

Also, upload a file from the demo site.


packages/ai-chat-components/src/components/file-uploads/src/file-upload-item.ts
```diff
@@ -59,49 +81,130 @@ class FileUploadItemElement extends LitElement {
...
-    if (!this._hasMediaPreview()) {
-      return;
-    }
-    const uploaderItem = this.shadowRoot?.querySelector(
+    this._syncInjectedStyles();
+  }
+
+  updated() {
+    this._syncInjectedStyles();
+  }
```
warrenmblood: `updated()` calls `_syncInjectedStyles()` unconditionally on every Lit render cycle. The bail-out on `_injectedStyle?.textContent === nextText` prevents duplicate DOM writes, but the shadow DOM query still runs every update.

Consider gating on changed properties:
```
updated(changedProperties: PropertyValues) {
    if (changedProperties.has('readOnly') || changedProperties.has('upload')) {
        this._syncInjectedStyles();
    }
}
```

packages/ai-chat/src/chat/components-legacy/MessageComponent.tsx
```diff
@@ -1260,6 +1265,21 @@ class MessageComponent extends PureComponent<MessageProps, MessageState> {
...
+          (message as MessageRequest).input?.structured_data
+        )
+      : [];
+
+    // An attachment-only message has nothing to put in a bubble, so it renders the
+    // chips alone rather than an empty bubble above them. `renderRequest` asks the
+    // same question, so the two cannot disagree.
+    const hasBubbleContent =
+      !messageIsRequest ||
+      hasRequestBubbleContent(localMessageItem, message as MessageRequest);
```
warrenmblood, line 1281: Noticing a strange bug in the demo where after the first message, the empty bubble renders on attachment-only messages. Wasn't able to trace the cause, but since this works properly on the first message, I suspect there's a stale value somewhere impacting `hasBubbleContent`:
Image

packages/ai-chat/src/chat/utils/fileAttachments.ts
warrenmblood, line 1: File attachments in sent messages do not currently support media preview like the comments suggest. `toAttachment` preserves the file type icon from the upload chip, but the image/video previews get lost. `MessageFileAttachment` needs to optionally carry the live File so the chip can use it for `objectURL` generation.

warrenmblood (changes requested):

warrenmblood (approved):
LGTM!

PR #1211, opened 2026-04-09
feat: typeahead/mentions/commands

ethanwinters:
Closes #730 #1193 

The goal of this PR is to serve as our foundation for prompt line work. 

Introduces a new typeahead, mentions, and commands system built on ProseMirror in `@carbon/ai-chat-components`, integrated into the `@carbon/ai-chat` Input component. This is a large commit with side-effect and temporary work — the `autocomplete` and `file-uploads` components will be rewritten with a new design in a follow-up.

The primary focus of this PR is the new `packages/ai-chat-components/src/components/input/` folder and how it is consumed upstream in `@carbon/ai-chat` and the relevant examples.

---

#### New

- **ProseMirror-based input system** in `packages/ai-chat-components/src/components/input/` — a plugin-based rich-text editor replacing the legacy `ContentEditableInput`. Includes plugins for value sync, trigger detection, token rendering, paste handling, auto-size, typing indicators, and max-length enforcement.
- **`SuggestionConfig` system** for configuring mentions (`@`), commands (`/`), and free-form typeahead triggers. Supports static item lists, async item fetching with debounce, custom list renderers, and custom token renderers.
- **`PublicConfig.input`** — new `InputConfig` type with `maxInputCharacters`, `isVisible`, `isDisabled`, and `suggestions` (an array of `SuggestionConfig`). All suggestion-related types are marked `@experimental`.
- **`StructuredFieldType`** extended with `"mention"` and `"command"` values for structured data in outgoing messages.
- **8 new example applications** (4 React, 4 Web Components):
  - `typeahead` — basic autocomplete suggestions
  - `typeahead-custom` — custom suggestion list renderer
  - `mentions-and-commands` — `@` mentions and `/` commands with structured data
  - `mentions-and-commands-custom-render` — custom token chip rendering
- **New React wrappers** in `@carbon/ai-chat-components`: `input-shell`, `input-send-control`, `stop-streaming-button`, `file-uploads`, `autocomplete`.
- Input shell moved file uploads on top of the input field instead of underneath
- Added typeahead for all the responses in the dropdown in the demo site
- Both the markdown component and the input field will bump font size from 14px to 16px when in mobile

#### Changed

- **`Input.tsx`** in `@carbon/ai-chat` rewritten to compose the new `InputShellElement`, `InputSendControl`, `FileUploads`, and `Autocomplete` components from `@carbon/ai-chat-components`. Manages suggestion fetching, file uploads, and Redux state integration.
[...]

PR #2022, opened 2026-07-31
feat(prompt-line): add storybook

### LINKED ISSUE #1347: Implement Storybook for prompt-line, input-shell, etc

warrenmblood:
Closes #1347

Adds prompt line to Storybook in React & Web Components. Also migrates autocomplete component files inside of prompt line (the only context in which autocomplete is used).

Additionally fixes bug in conversation starters trigger via mouse focus.

#### Changelog

**New**

- `packages/ai-chat-components/src/components/prompt-line/__stories__/`: Adds the following stories in React & WC versions to cover different configurations and use cases of prompt line:
  - Default (non-expanded)
  - Expanded
  - Commands and mentions
  - Conversation starters
  - File uploads
  - Typeahead
- Overview page providing documentation and code examples. Follows same pattern used in existing components.
- `autocomplete.ts`: Added an anchorElement property to the AutocompleteElement; the outside-click handler now treats clicks on the anchor element (the editor) as inside-clicks so they don't dismiss the list.
- `trigger-utils.ts`: Added a resetTriggerChangeState export that deletes an editor's entry from the coalescing map, allowing the same detail to be re-dispatched after an external dismissal.
- `carbon-starter-trigger.test.ts`: New test file covering events that trigger the display of the conversation starters overlay.

**Changed**

- `packages/ai-chat-components`: Moved `autocomplete` component files inside of `prompt-line` and updated relevant import paths.
- `autocomplete-controller.ts`: After dismissal, resets trigger-utils coalescing state so the editor can re-emit the same trigger detail on the next focus/update; sets anchorElement on the autocomplete element pointing to the editor DOM; registers a mousedown handler on the controller element that calls preventDefault to stop the autocomplete list from stealing focus from the editor.
- `carbon-starter-trigger.ts`: The onTransaction hook now only calls maybeEmit when the editor is focused; the onBlur hook that was force-clearing the trigger was removed entirely since blur-to-dismiss is now handled by the controller.
- `useChatAutocomplete.tsx`: Sets anchorElement on the list element after mounting. Adds a handleContainerMousedown callback that calls preventDefault and wires it to both the CustomElementHost wrapper div and the CustomReactNodePortal host element. Both host components now accept and wire up an onMousedown prop.

#### Testing / Reviewing
- Go to storybook deploy preview
- Test all new storybook pages under "Prompt line". Make sure setting controls works properly.
- Run autocomplete from the storybook and the demo app. Ensure no functionality broke when migrating files.


packages/ai-chat-components/.storybook-react/main.js
```diff
@@ -46,6 +46,13 @@ export default {
     name: getAbsolutePath('@storybook/react-vite'),
     options: {},
   },
+  typescript: {
+    // react-docgen's Babel parser does not support the `@decorator export
+    // class {}` syntax used by Lit / @carbon/web-components source files that
+    // are transitively imported by the React stories. Disabling docgen here
+    // is safe because all argTypes in the React stories are declared manually.
+    reactDocgen: false,
```
ethanwinters: **Blocker** — this disables prop docgen for *all* React stories to work around a single file.

`packages/ai-chat-components/src/components/prompt-line/src/prompt-line.ts:61` is the only file in the package using the `@decorator export class {}` form:

```ts
@carbonElement(`${prefix}-prompt-line`)
export class PromptLineElement extends LitElement {
```

Every other component uses the bare-class form react-docgen's Babel parser accepts — `card.ts:23`, `launcher.ts:33`, `autocomplete.ts:119` are all `@carbonElement(...)` + `class X extends ...`, exported via `export default X` at the bottom. `prompt-line.ts` already has `export default PromptLineElement;` on line 468, so the fix is to drop the inline `export`:

```ts
@carbonElement(`${prefix}-prompt-line`)
class PromptLineElement extends LitElement {
```

...adding `PromptLineElement` to a named export at the bottom if the named binding is still needed.

The cost of the global switch is measurable: with `reactDocgen: false` the React `<ArgTypes>` tables lose their type information. `Components/Launcher → Overview` on :7007 currently renders **Name | Description | Default** with no Type column and every default `-`, and 25 React Overviews use `<ArgTypes of={...}/>`. That's a regression to components this PR otherwise doesn't touch.

Note the story can't dodge the parse error by importing less — `react/prompt-line.tsx:13` imports the Lit element directly and `prompt-line/index.ts` imports `./src/prompt-line.js`, so `prompt-line.ts` is in the React story's module graph either way. Fixing the syntax is the only route.

packages/ai-chat-components/src/components/prompt-line/__stories__/prompt-line.stories.js
```diff
@@ -0,0 +1,979 @@
...
+    el.enableSendButton = enableSendButton;
+    return el;
+  },
+};
+
+// ---------------------------------------------------------------------------
+// Conversation starters — starters overlay + toggle action only
+// ---------------------------------------------------------------------------
+
+export const ConversationStarters = {
```
ethanwinters, line 889: **Blocker** — this story and its React twin don't work when the editor is focused by mouse: the list flashes and disappears. Root cause, evidence, and the suggested patch are in the review summary — it's the `document` click listener registered in `autocomplete.ts` `connectedCallback` firing on the same click that opens the list.

Pre-existing rather than introduced here, but it's the story this PR is adding, so it should go in with the fix.

packages/ai-chat-components/src/components/prompt-line/__stories__/prompt-line-react.stories.jsx
```diff
@@ -0,0 +1,790 @@
...
+  </div>
+);
+
+// ---------------------------------------------------------------------------
+// Story meta
+// ---------------------------------------------------------------------------
+
+export default {
+  title: 'Preview/Prompt line',
+  argTypes: {
```
ethanwinters, line 119: **Important** — these `argTypes` and `args` are a verbatim copy of the Lit meta at `prompt-line.stories.js:650`, ~60 lines duplicated. [storybook.md](https://github.com/carbon-design-system/carbon-ai-chat/blob/main/packages/ai-chat-components/references/storybook.md) has the React side spread the Lit story's so the two can't drift:

```jsx
import { default as WCMeta } from './prompt-line.stories';

export default {
  title: 'Preview/Prompt line',
  argTypes: { ...WCMeta.argTypes },
  args: { ...WCMeta.args },
};
```

Same for the per-story overrides below — `args: { ...Default.args, … }` rather than restating. This costs nothing on the docgen front: the Lit sources are already in this module's graph via `react/prompt-line` and `../index`.

Titles and story names do match across the two sides, which is the part that's easy to get wrong — nice.

packages/ai-chat-components/src/components/prompt-line/__stories__/prompt-line-react.mdx
```diff
@@ -0,0 +1,444 @@
...
+      <td><code>onFileRemove</code></td>
+      <td><code>cds-aichat-file-remove</code></td>
+      <td>Fired when the user clicks the remove button on a file chip.</td>
+      <td>• <code>fileId</code> (string)</td>
+    </tr>
+  </tbody>
+</table>
+
+
+## Component API
```
ethanwinters, line 373: **Nit** — these hand-maintained prop tables will drift the first time a wrapper prop changes; the Lit Overview gets its equivalents generated via six `<ArgTypes of="cds-aichat-…"/>` blocks.

Acknowledging this is more than a one-liner: prompt line is a six-component composite and the React meta has no `component:`, so there's no single `<ArgTypes of={Stories.Default}/>` that mirrors the Lit side — it needs a ref per sub-component. Worth doing once the `reactDocgen` fix lands (until then `<ArgTypes>` would render nothing useful anyway), or tracking as follow-up.

packages/ai-chat-components/src/components/prompt-line/__stories__/story-styles.scss
```diff
@@ -0,0 +1,36 @@
...
+ * .prompt-line-story-hint    — muted instructional text above the input.
+ */
+
+.prompt-line-story-wrapper {
+  inline-size: 640px;
+  max-inline-size: 100%;
+}
+
+.prompt-line-story-hint {
+  color: var(--cds-text-helper, #6f6f6f);
```
ethanwinters: **Nit** — hardcoded hex fallbacks and font stacks where Carbon tokens exist. `card/__stories__/story-styles.scss` is the good precedent in this package:

```scss
@use '@carbon/styles/scss/theme';
@use '@carbon/styles/scss/type' as *;
@use '@carbon/layout';

.prompt-line-story-hint {
  @include type-style('helper-text-01');

  color: theme.$text-helper;
  margin-block-end: layout.$spacing-03;
}
```

The `var(--cds-*)` references do resolve across themes, so this is cosmetic rather than a dark-theme bug — but the type sizes won't track the Carbon scale.

packages/ai-chat-components/src/components/prompt-line/__stories__/prompt-line-react.stories.jsx
```diff
@@ -0,0 +1,790 @@
...
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+const Wrapper = ({ children }) => (
+  <div style={{ width: '640px', maxWidth: '100%' }}>{children}</div>
+);
+
+// Gives the prompt line breathing room above it for the autocomplete overlay.
+// Mirrors .prompt-line-story-wrapper--bottom in story-styles.scss.
```
ethanwinters: **Nit** — stale reference: there's no `.prompt-line-story-wrapper--bottom` in `story-styles.scss` (it defines only `.prompt-line-story-wrapper`, `.prompt-line-story-hint`, and `.prompt-line-story-hint code`). Either add the modifier or drop the second line of the comment.

packages/ai-chat-components/src/components/prompt-line/__stories__/prompt-line.stories.js
```diff
@@ -0,0 +1,979 @@
...
+    const toggleIcon = this._startersEnabled ? ChatOff16 : Chat16;
+    const toggleLabel = this._startersEnabled
+      ? 'Hide conversation starters'
+      : 'Show conversation starters';
+
+    return html`
+      <style>
+        ${styles}
+      </style>
+      <div style="width:640px;max-inline-size:100%;margin-block-start:320px">
```
ethanwinters: **Nit** — this inline style duplicates `.prompt-line-story-wrapper` (which four of the six stories use, e.g. lines 455, 745, 814) and reintroduces the physical `width` property the SCSS avoids in favour of `inline-size`. Same at line 287.

The `margin-block-start: 320px` spacer is also repeated inline at lines 148, 460, and 600 — worth a modifier so the magic number lives in one place:

```scss
.prompt-line-story-wrapper--bottom {
  margin-block-start: 320px;
}
```

That also makes the comment on `prompt-line-react.stories.jsx:52` true.

ethanwinters (dismissed):
Two blockers, both fixable in this PR.

## 1. Conversation starters is broken on mouse focus (Lit and React)

Click into the editor and the suggestion list mounts and self-dismisses ~4ms later. Verified in the running Storybook:

- Mouse click → `cds-aichat-autocomplete` added, then removed within 4ms; controller trigger ends `null`.
- Keyboard <kbd>Tab</kbd> into the same editor → list stays open (640×203, 4 items).
- Typed triggers (`@` mention, typeahead) unaffected — both verified working on mouse.

Captured stack: `_handleClickOutside` → `_dismiss()` → `controller.dismiss()`.

`autocomplete.ts:196` registers the `document` click listener inside `connectedCallback`, which runs *during* the click that opens the list. That same click then bubbles to `document`, `this.contains(event.target)` is false (the target is the ProseMirror editor), and the list dismisses itself. Starters is the only trigger that opens on focus, so it's the only one that collides — which is why mentions and typeahead look fine.

The bug is pre-existing: the diff only changes import lines in `autocomplete.ts`, and it reproduces in the `prompt-line-conversation-starters` example on `main`. But since this PR is what puts it on display, better to fix it here than ship two stories that don't work. For the record, `refactor: consolidate autocomplete` (#2003) didn't touch this path — `_handleClickOutside` dates to #1211; the newer open-on-focus starters trigger is what made it reachable.

Smallest fix — register the listener after the opening click has finished dispatching:

```ts
connectedCallback() {
  super.connectedCallback();
  this.addEventListener('keydown', this._handleKeydown);
  // Deferred: registering synchronously would catch the tail of the very
  // click that opened the list and dismiss it immediately.
  this._clickOutsideTimer = setTimeout(() => {
    document.addEventListener('click', this._handleClickOutside);
  });
}

disconnectedCallback() {
  super.disconnectedCallback();
  clearTimeout(this._clickOutsideTimer);
  this.removeEventListener('keydown', this._handleKeydown);
  this._listboxEl?.removeEventListener('mousedown', this._handleMousedown);
  document.removeEventListener('click', this._handleClickOutside);
}
```

Worth a regression spec in `prompt-line/autocomplete/__tests__/autocomplete.test.ts`: mount the element while a click is dispatching and assert no `cds-aichat-autocomplete-dismiss` fires.

(Longer term, dismissal arguably belongs to `AutocompleteController` — it already owns trigger lifecycle — rather than every list element wiring its own `document` listener. Out of scope here.)

## 2. `reactDocgen: false` is a global fix for a one-file problem

See the inline comment on `.storybook-react/main.js`.

---

I did not run the full test suites or an a11y-addon pass over the six new stories — please confirm `npm run test --workspace=@carbon/ai-chat-components` and `npm run build --workspace=@carbon/ai-chat-components` before merge.

packages/ai-chat-components/src/components/prompt-line/__stories__/prompt-line-react.mdx
ethanwinters: If not clear, this can be a follow up

packages/ai-chat-components/src/components/prompt-line/src/autocomplete-controller.ts
```diff
@@ -266,6 +267,17 @@ export class AutocompleteController {
...
+    // Reset the trigger-utils coalescing state so the extension can re-emit
+    // the same detail on the next onFocus/onUpdate without being swallowed.
+    const editor = this._promptLine?.getEditor();
+    if (editor) {
+      resetTriggerChangeState(editor);
+      // Fire a no-op transaction so onUpdate re-evaluates immediately (handles
+      // the case where the editor is still focused and empty after dismissal).
+      if (editor.view?.dispatch) {
+        editor.view.dispatch(editor.state.tr);
+      }
```
ethanwinters: **Blocker** — this synthetic transaction makes Escape re-open the list it just closed.

With `onTransaction` now emitting whenever the editor is focused (`carbon-starter-trigger.ts:49-53`), the sequence is:

> Escape → `dismiss()` → `resetTriggerChangeState` clears the coalescing guard → `dispatch(state.tr)` → `onTransaction` → editor still focused and empty → re-dispatches `starter` → list reopens.

Verified on `preview-prompt-line--conversation-starters`: after one Escape press the trigger-change trace was `['starter', 'starter']` and the overlay was still open. Mentions and commands are unaffected — the editor isn't empty there, so `maybeEmit` sends `null` — which makes this specific to conversation starters, the surface this PR's new story showcases.

The fix is to drop the dispatch and keep the reset:

```ts
const editor = this._promptLine?.getEditor();
if (editor) {
  resetTriggerChangeState(editor);
}
```

`resetTriggerChangeState` on its own already covers what it was added for: after a blur-dismissal the coalescing map still holds `{starter}`, so a re-focus would be swallowed — clearing it fixes that without self-re-triggering.

I applied this locally and confirmed all four paths, then reverted:

| Path | Result |
| --- | --- |
| Escape | list closes, editor keeps focus ✅ |
| Blur → refocus | closes, then reopens ✅ (what the reset was added for) |
| Mouse-select a starter | inserts text, closes, keeps focus ✅ (this PR's fix, intact) |
| `@` mention Escape | unaffected ✅ |

Worth noting the comment on line 275-276 describes the intended behaviour as re-evaluating "the case where the editor is still focused and empty after dismissal" — that case is exactly Escape, where staying closed is what the user asked for.

ethanwinters (changes requested):
Thanks for turning the last round around — the React stories now spread `WCMeta`, `story-styles.scss` is on Carbon tokens and logical properties, the `--autocomplete` modifier landed, and dropping the inline `export` on `PromptLineElement` let `reactDocgen` stay on for the whole package. `index.ts` still re-exports it, so nothing downstream breaks.

The mouse-focus bug is genuinely fixed — I verified in the running Storybook that clicking a conversation starter inserts the text, closes the list, and keeps editor focus. Keyboard nav and blur-dismiss check out too.

Requesting changes for **one regression the fix introduced**, which I confirmed live rather than inferred from the diff: **Escape no longer dismisses the conversation-starters overlay.** Details in the inline comment — it's a two-line change that I applied locally and verified across all four interaction paths.

Please also add a regression test for it. The new `carbon-starter-trigger.test.ts` is a good addition and passes 3/3 across Chromium, Firefox and WebKit, but it exercises the extension in isolation — this regression lives in the extension↔controller interaction, which is precisely why it got through.

Everything else I found is non-blocking, and it's filed as #2065 (assigned to you) so it doesn't hold this up. One of those items is worth flagging as my doing: the `*API` / `*ArgTypes` carrier exports are your answer to my earlier note about hand-maintained React prop tables, and the mechanism works — but because every named export in a CSF file is a story, it adds 12 entries to the story index. I confirmed against the static build that Chromatic consumes (`storybook:react:build`): all 12 are `type: story` with `tags: ['test', 'manifest']`, so `!dev` hides them from the sidebar but not from snapshots. The issue has the suggested shape.

Two things I checked and am explicitly **not** flagging: the autocomplete directory move is not a breaking change (that path is absent from every release tag through `@carbon/ai-chat-components@1.8.0`, so there are no published consumers), and the Overview MDX hero-Canvas shape is a repo-wide alignment matter rather than this PR's to fix.

ethanwinters (approved):

PR #2058, opened 2026-08-04
refactor(image): consolidate ClickableImage into Image

### LINKED ISSUE #1434: Consolidate ClickableImage into Image
## Background

`ClickableImage` wraps `Image` in a button or a link, and folding the two together is still the right move — that part of this issue hasn't changed. What has changed is where the merged result should live, which is now an open question.

| File | Lines | Coupling |
| --- | --- | --- |
| `components-legacy/responseTypes/image/Image.tsx` | 229 | **None direct.** Every string already arrives as a prop |
| `components-legacy/responseTypes/util/ClickableImage.tsx` | 82 | **None** |

Parent epic #1803.

## The open question

This issue originally said to relocate `Image` into `@carbon/ai-chat`. An audit of the tree found `Image` has no direct store or service-manager coupling at all, which makes it the strongest port candidate in the epic under the criteria in #1425.

Its only coupling is indirect, through two children:

- `InlineError`, which is itself being ported.
- `TextBlock`, which is itself being ported and whose markdown becomes a slot.

Once both of those move, nothing chat-specific is left behind.

The question underneath this one is now answered: **the `image` response type is a keep** (#1845). It survives 2.0 with no deprecation, so the placement question is worth asking rather than moot.

## The scope split

The placement question and the fold are separable, so this issue only carries the fold.

**In scope here, on 1.x:** fold `ClickableImage` into `Image`, migrate the two call sites, delete the file. That's a React-to-React consolidation. It removes a duplicate wrapper regardless of where `Image` eventually lives, and nothing gates it any more.

**Not here:** whether the merged `Image` ports to Lit or relocates. That's the same shape of question every other component in #1863 answers, so it rides with that epic, post-2.0. Two facts worth carrying over when it's asked — `renderIcon` is a React element type today and becomes a slot in Lit while `useAITheme` becomes a plain attribute, and `Image` composes `cds-aichat-card` for its frame.

## Goal

`ClickableImage` no longer exists, and `Image` covers the button and link behavior itself.

## Approach

1. Fold `ClickableImage`'s button and link behavior into `Image`.
2. Migrate both call sites: `BaseButtonItemComponent` and `IFramePreviewCard`.
3. Delete `ClickableImage.tsx`.

## Notes

- **`needsAnnouncement` is dead**, and now has its own issue — #1912. The switch passes it to `Image` at `MessageTypeComponent.tsx:461` but the prop is never read. The field itself stays: `MessageComponent.tsx:390`/`:403` read it, and `reducers.ts:1510` / `actions.ts:405` clear it.
- The `IFramePreviewCard` call site sits in the iframe subtree. `iframe` is also a keep (#1845), so that subtree survives, but it still gates on #906 for the sandboxed-iframe primitive — a thin shim may be needed if it moves later.
- `Image` handles a 10-second load timeout and a skeleton placeholder, and both travel with it.
- `Image` also has an `inline` mode that drops the card frame, which needs to keep working.

## Acceptance criteria

- [ ] `ClickableImage` is folded into `Image`, and `ClickableImage.tsx` is deleted.
- [ ] Both call sites are migrated.
- [ ] Images render as they do today, including the inline mode, the skeleton, and the load timeout.
- [ ] Build and tests pass.

## Out of scope

- Whether the merged `Image` ports to Lit or relocates — #1863, post-2.0.
- The dead `needsAnnouncement` prop — #1912.
- The iframe subtree.

## Related

- Parent epic: #1803
- Unblocked by: #1845 (`image` is a keep), under #1913
- Placement question rides with: #1863
- Split out: #1912
- Coordinate with: #1847, #1848, #1852, #1856

AlexanderMelox:
Closes #1434

Folds `ClickableImage` into `Image`. When `onClick` or `isLink` is passed, `Image` wraps itself in a `<button>` or `<a>` — no separate component needed.

#### Changelog

**Changed**

- `Image` accepts five new props: `onClick`, `isLink`, `disabled`, `buttonAltText`, `target`
- `BaseButtonItemComponent` and `IFramePreviewCard` now use `Image` directly

**Removed**

- `ClickableImage` component (`ClickableImage.tsx`, `ClickableImage.scss`)
- Two dead CSS selectors that targeted `text-holder-tile` class names no longer in the render tree
- Removed unused useRef from Image

#### Testing / Reviewing

In the demo, type `image` in the chat input. This exercises all affected paths:

- Plain image card (non-clickable) — no wrapper rendered
- URL button with `image_url` — renders as `<a>` (link variant)
- POST_BACK / SHOW_PANEL / CUSTOM_EVENT buttons with `image_url` — render as `<button>`
- IFrame preview card — button variant, opens the frame panel on click

Hover over the link-variant and button-variant image cards. Confirm the image dims and the title underlines.


packages/ai-chat/src/chat/components-legacy/responseTypes/image/Image.tsx
```diff
@@ -132,6 +162,35 @@ function Image(props: ImageProps) {
       </div>
     </Card>
   );
+
+  if (isLink) {
```
davidmenendez, line 173: `isLink` feels a little redundant here with `displayURL`. i suggest we just use `href` and that way it can be easily inferred without the ned for extra `is_` props. similar to how it's inferred just below with `onClick`

packages/ai-chat/src/chat/components-legacy/responseTypes/image/Image.tsx
```diff
@@ -132,6 +162,35 @@ function Image(props: ImageProps) {
...
+    return (
+      <a
+        className="cds-aichat--clickable-image"
+        href={displayURL}
+        rel="noopener noreferrer"
```
davidmenendez: super minor, but lets just add a prop for `rel` and use `noopener noreferrer` as the default / fallback

packages/ai-chat/src/chat/components-legacy/responseTypes/image/Image.tsx
```diff
@@ -132,6 +162,35 @@ function Image(props: ImageProps) {
...
+        target={target}
+        onClick={onClick}>
+        {baseImageCard}
+        {buttonAltText && <VisuallyHidden>{buttonAltText}</VisuallyHidden>}
```
davidmenendez, line 182: i'm a bit thrown by the usage here of these `VisuallyHidden` components. anchors with embedded images don't actually need additional labels. the `alt` attribute in the `img` tag is all that's needed  when the child is an image

davidmenendez (approved):
i had a few thoughts here, but i'm going to just approve because i know we're trying to eliminate legacy components in the not too distant future, so there's no point in overthinking them. but feel free to check it out!

packages/ai-chat/src/chat/components-legacy/responseTypes/image/Image.tsx
AlexanderMelox: @davidmenendez I just ported it over exactly 1:1 to `ClickableImage` I can go ahead and refactor these too if it makes sense

packages/ai-chat/src/chat/components-legacy/responseTypes/image/Image.tsx
AlexanderMelox: I agree, there are some more nested components and parents ill have to update for this change and can be a little out of scope for this issue. Maybe we can make a followup issue for clean up?

PR #2098, opened 2026-08-07
Prompt line review fixes

### LINKED ISSUE #2097: Prompt line review fixes
The following issues were flagged during the prompt line review. Additional issues requiring further discussion with design are tracked in #2085.

Acceptance criteria for this issue is to fix the following:

### 1) Error state border
In expanded layout with `rounded=false`, the red error border isn't visible on the bottom. Swap the css `border` for `outline` to avoid conflicts with the top/bottom borders used as dividers between the slots. Also use `outline-offset` so the outline no longer adds space around the prompt shell.

### 2) Video preview image not rendering in file upload chips
This appears to be a regression. The preview is still working for images, but the `<video>` elements aren't getting their `src` attribute properly set.

### 3) Use overflow handler in storybook for prompt line action buttons
`@carbon/ai-chat` uses the `createOverflowHandler` utility to collapse the row of action buttons as a window resize shrinks the width of the prompt line. Implement this in Storybook so it illustrates the same behavior.

### 4) Conversation starters don't have send button inside autocomplete items
Following the same pattern as typeahead, the Conversation starters examples and stories should have the send button enabled inside autocomplete.

### 5) Active color shows on click when send button is disabled
Clicking the send button while it is disable displays the interactive color.

warrenmblood:
Closes #2097

Fixes series of issues identified in prompt line review.

One of the review items was enabling the send button inside conversation starter items — the same feature already working in the typeahead story. Wiring that up exposed several gaps that had to be fixed together.

The deepest one is in the coalescing layer in `trigger-utils.ts`, which prevents re-dispatching the same trigger detail on every editor transaction. When a user clicks the send button inside a starter item, the controller called `dismiss()` — which reset that coalescing state. That meant the starter trigger immediately re-fired on the next transaction while the editor was still focused and empty, reopening the list the user had just sent from. The fix adds a `keepCoalesced` flag to `dismiss()`: the send path passes `true`, leaving the coalescing state intact so the same detail is swallowed. The recovery path is `onBlur` on `carbonStarterTrigger` — when the editor loses focus, coalescing state is cleared so the starter list can reappear normally on the next focus. Without both halves the starters either bounce back immediately after send or fail to reopen.

Enabling the send button also exposed that `CustomListProps` had no `onSend` callback — a custom list renderer had no sanctioned way to invoke the send path, so the starters story couldn't wire the in-item send button at all. And `useChatAutocomplete` had the `onSelectItem` hook option partially wired: it existed in the interface but wasn't threaded through the custom-element and custom-React-node render paths. Both gaps are fixed here.

#### Changelog

**Changed**

- Error state: replace `border` with `outline` on the input container; adjust `error-message.scss` separator math to account for the outline offset — fixes the red border being clipped on the bottom in non-rounded layout. Also removes unnecessary `z-index: 0` from `::after` pseudo-element.
- Video upload chip: use `objectURL` (blob URL from `URL.createObjectURL`) as `<video src>`, not the raw `url` field — fixes blank video thumbnails in chips.
- Send button: add `:not([disabled])` to the `:active`/`:focus` fill selector — stops the interactive color from flashing when the disabled send button is clicked.
- `AutocompleteController.dismiss` gains a `keepCoalesced` parameter (default `false`); `carbonStarterTrigger` calls `resetTriggerChangeState` on blur — together these fix conversation starters re-opening immediately after the in-item send button is clicked.
- `CustomListProps` gains an `onSend` callback — previously, a custom list renderer had no sanctioned way to trigger the send path; without it the conversation starters story couldn't wire the in-item send button at all.
- `useChatAutocomplete` threads `onSelect` and `onSend` through all three render paths (custom element, custom React node, built-in autocomplete) and surfaces them via `onSendItem` / `onSelectItem` hook options.
- Prompt line Storybook: action button row uses `createOverflowHandler` to add responsive collapsing behavior. This matches the exact pattern used in `@carbon/ai-chat`'s `InputActionsInline`.
- Conversation starters story exposes `enableSendButton` and `attached` controls (previously disabled) and fires `cds-aichat-autocomplete-select` / `cds-aichat-autocomplete-send` storybook actions.

**New**
- Added the following tests:
- `autocomplete-controller-element.test.ts`: WTR integration tests verifying `cds-aichat-autocomplete-item-selected` and `cds-aichat-autocomplete-item-send` fire with the correct `bubbles`, `composed`, and `detail` shape
- `autocomplete-controller.test.ts`: two new cases for `dismiss(keepCoalesced)` — one confirming the default path resets coalescing so the same trigger re-fires, one confirming `keepCoalesced: true` suppresses re-fire until `resetTriggerChangeState` is called (simulating blur)
- `file-upload-item.test.ts`: regression guard confirming the video preview `<video src>` is a `blob:` URL

#### Testing / Reviewing

- Go to Storybook deploy preview
- Verify all issues from original issue (#2097) are fixed
- Check prompt line examples for any regressions. At minimum test the following examples for React & WC:
  `prompt-line-conversation-starters`
  `prompt-line-typeahead`


packages/ai-chat-components/src/react/hooks/useChatAutocomplete.tsx
```diff
@@ -260,17 +280,54 @@ function CustomElementHost({
...
+        (e as CustomEvent<{ item: SuggestionItem }>).detail?.item
+      );
+    const sendListener = (e: Event) =>
+      onSendRef.current?.((e as CustomEvent<{ text: string }>).detail?.text);
+
+    element.addEventListener(
+      'cds-aichat-autocomplete-dismiss',
+      dismissListener
+    );
+    element.addEventListener('cds-aichat-autocomplete-select', selectListener);
```
ethanwinters: **Blocker** — these auto-listeners double up with the callbacks already handed to `renderCustomList`, and they make `CustomListProps` mean opposite things in the two render paths.

`renderCustomList` is called with `onSelect: handleSelect` / `onSend: handleSend` / `onDismiss: dismiss` (lines 203-209), and then this effect *also* listens for `cds-aichat-autocomplete-select` / `-send` / `-dismiss` on the returned element and calls the same handlers. Those events are `bubbles: true, composed: true` (`autocomplete.ts:411-420`), so both fire.

A consumer who wires the props the way `types.ts:80-87` documents them ("Callback to invoke when user selects an item") therefore gets the handler twice. What that costs differs by path — I checked both against the running story by dispatching each event twice:

- **select — absorbed.** `AutocompleteController.select()` calls `dismiss()`, which nulls `_trigger`; the second call hits the `if (!trigger || !editor)` guard and bails. Text is inserted once. `onSelectItem` does still fire twice.
- **send — not absorbed.** `handleSend` (lines 160-166) has no equivalent guard: `dismiss(true)` early-returns on the second call, but `onSendItemRef.current?.(text)` runs unconditionally. Confirmed two `cds-aichat-autocomplete-send` actions from two dispatches. `@carbon/ai-chat` wires `onSendItem: (text) => onSendInput(text)` (`Input.tsx:506`), so that's **the message sent twice**.

The select-path wiring is exactly what both examples used before this PR, which is why both had to have it deleted here — any external consumer of `PublicConfig.input.starters.renderCustomList` is on that pattern, silently and with no deprecation.

Meanwhile `AutocompleteControllerElement._renderCustomList` (`autocomplete-controller.ts:729-735`) passes the same props and attaches nothing, so on the pure-WC path you still *must* wire them — as `prompt-line.stories.js:283-292` does. One documented interface, two opposite requirements.

Simplest fix that keeps back-compat: drop these listeners (and the matching block at 420-422 in `CustomReactNodePortal`) and leave the props as the single contract, matching the Lit path. If you'd rather keep the auto-wiring, then add the same listeners to `_renderCustomList` in the controller element, remove `onSelect`/`onSend` from `CustomListProps`, and call it out as a breaking change.

ethanwinters (commented):
One thing I'd want fixed before merge.

**`CustomListProps` now means opposite things in the two render paths.** The React hook passes `onSelect`/`onSend`/`onDismiss` into `renderCustomList` *and* attaches its own listeners for the matching events; the Lit controller element passes the same props and attaches nothing. So a consumer who wires these the way `types.ts` documents gets them invoked twice in React. On the select path that's absorbed by the controller's own guard, but `handleSend` has no such guard — `onSendItem` fires twice, which `@carbon/ai-chat` turns into a duplicate message send. Either way the same `renderCustomList` can't be written once for both flavors. The tell is that both examples in this PR had to have their wiring deleted — external consumers on that pattern break silently, with no deprecation. Details inline.

The rest is important-but-not-blocking: the `onSend` JSDoc documents a contract the code doesn't implement, and the Storybook overflow row diverges from the `InputActionsInline` it's mirroring in three ways (menu not anchored to its trigger, `data-fixed` missing, `useEffect` deps).

### Test / verification gaps

- Nothing covers the double-wiring hazard. A case where `renderCustomList` wires `onSelect` from the element event and asserts `select()` fires **once** would have caught it.
- `autocomplete-controller-element.test.ts` uses items where `label === value`, so it can't detect the `label`-vs-`value` mismatch in the send path. Add an item with a distinct `value`.
- The overflow row has no coverage at all. Story-only code, so that's defensible — but a smoke test would have caught the anchoring and `data-fixed` issues.
- The `dismiss(keepCoalesced)` unit tests are good, and the behavior holds up in the browser. No gap there.
- Per `packages/ai-chat-components/AGENTS.md`, the new `@fires` entries on the controller element mean `npm run custom-elements` should be rerun locally to confirm the manifest regenerates cleanly.

Also worth noting for the PR body: the changelog says `onSendItem`/`onSelectItem` "existed in the interface but weren't threaded through" — `onSendItem` did, but `onSelectItem` is added by this PR.

packages/ai-chat-components/src/components/prompt-line/src/tiptap/types.ts
```diff
@@ -76,10 +76,15 @@ export interface CustomListProps {
...
   query: string;
-  /** Callback to invoke when user selects an item. */
-  onSelect: (item: SuggestionItem) => void;
   /** Callback to invoke when list should be dismissed. */
   onDismiss: () => void;
+  /** Callback to invoke when user selects an item. */
+  onSelect: (item: SuggestionItem) => void;
+  /**
+   * Callback to invoke when user clicks the send-directly-from-item button.
+   * The text is the item's value that was sent.
```
ethanwinters, line 85: **Important** — this documents a contract the code doesn't implement.

`Autocomplete._handleSend` dispatches `detail: { text: item.label }` (`autocomplete.ts:360`), not `value`. `AutocompleteController.select()` uses `item.value ?? item.label` (`autocomplete-controller.ts:308`), so select and send disagree for any item where the two differ — and `@carbon/ai-chat` feeds this straight through (`Input.tsx:506` → `onSendInput(text)`), so such a starter would send the wrong string.

The new integration test can't catch it because its fixtures use `label === value`.

Prefer aligning the code with this doc:

```ts
detail: { text: item.value ?? item.label },
```

and add a test item with a distinct `value`. If `label` is intentional, reword this line to say so instead.

packages/ai-chat-components/src/components/prompt-line/__stories__/prompt-line.stories.js
```diff
@@ -44,29 +47,180 @@ import {
...
+
+        ${this._fixedActions.map((a) => this._renderButton(a))}
+        ${
+          this._menuOpen && hiddenActions.length > 0
+            ? html`
+                <cds-menu
+                  open
+                  label="More actions"
+                  ${ref(this._menuRef)}
+                  style="position: absolute;"
```
ethanwinters: **Important** — the menu isn't anchored to its trigger.

`InputActionsInline` positions it with `useFloatingMenuPosition(open, triggerRef, menuRef)` and renders the `<Menu>` as a sibling *outside* the measured container. Here that hook is dropped and replaced with a bare `position: absolute` and no coordinates, so the menu lands at the container's origin. Confirmed in the browser at a 380px viewport: it opens on top of the five still-visible action buttons rather than beside the "⋮".

That's also why `_overflowButtonRef` and `_menuRef` (lines 76-77, 176, 191) are assigned and never read — leftovers from the copy that advertise an anchoring the story doesn't do. Same in the React story (`triggerRef` at line 105, `menuRef` at 182).

Either port the positioning (set `x`/`y` on `cds-menu` from the trigger's `getBoundingClientRect()`), or accept container-anchored placement and delete the two dead refs.

packages/ai-chat-components/src/components/prompt-line/__stories__/prompt-line.stories.js
```diff
@@ -44,29 +47,180 @@ import {
...
+          }}>
+          ${iconLoader(OverflowMenuVertical16, { slot: 'icon' })}
+          <span slot="tooltip-content">More actions</span>
         </cds-icon-button>
-      `
-    )}
-  </div>
-`;
+
+        ${this._fixedActions.map((a) => this._renderButton(a))}
```
ethanwinters: **Important** — fixed actions are rendered without `data-fixed`, so the handler will collapse them anyway.

`createOverflowHandler` identifies pinned items by `item.hasAttribute('data-fixed')` (`@carbon/utilities/es/overflowHandler/index.js`); `InputActionsInline` sets `data-fixed=""` on them. Without it these are ordinary overflow items — the opposite of "fixed". The React story diverges the other way and never renders fixed actions at all (it maps only `nonFixedActions`).

Neither branch is exercised: no entry in `dummyActions` sets `fixed`. Two divergent, untested implementations of unused flexibility is exactly what the laziness ladder asks you to delete.

Simplest: drop the `fixed` split from both stories. If it's meant to demonstrate the behavior, add `data-fixed=""` here, render `fixedActions` in the React story, and give one `dummyActions` entry `fixed: true` so it actually shows.

packages/ai-chat-components/src/components/prompt-line/__stories__/prompt-line-react.stories.jsx
```diff
@@ -87,27 +96,116 @@ const CarbonIconSlot = ({ icon, slot }) =>
...
+          if (hidden === 0) setMenuOpen(false);
+        },
+      });
+      requestAnimationFrame(() => setMeasuring(false));
+    });
+    return () => {
+      cancelAnimationFrame(setupRaf);
+      handler?.disconnect();
+    };
+  }, []);
```
ethanwinters: **Important** — empty deps while the effect reads `actions`.

`nonFixedCount` is captured from the first render and the handler is never rebuilt if the action list changes; `InputActionsInline` uses `[actions]` for exactly this reason. Not reachable today (`dummyActions` is a module constant) and not lint-caught either — `npm run lint` is `eslint packages --ext .js,.ts,.tsx`, so `.jsx` stories aren't covered. But mirroring the production component is the stated goal of the change.

```js
  }, [actions]);
```

While you're here: the item buttons use `disabled={disabled || undefined}`, dropping `a.disabled`. The Lit story uses `?disabled=${this.disabled || a.disabled}` and `InputActionsInline` uses `disabled || opt.disabled`. Harmless today, but the two stories should agree.

packages/ai-chat-components/src/components/prompt-line/src/error-message.scss
```diff
@@ -32,14 +32,15 @@ $block-class: #{$prefix}-error-message;
...
     position: absolute;
-    z-index: 0;
     display: block;
     background: theme.$border-subtle-00;
     block-size: 1px;
     content: '';
-    inline-size: calc(100% + #{layout.$spacing-03} + #{layout.$spacing-03});
+
+    // account for padding and error outline on input container
+    inline-size: calc(100% + 2 * #{layout.$spacing-03} - 2px);
```
ethanwinters: **Nit** — these offsets hardcode the input container's `1px` outline from a different file (`prompt-line-shell.scss:100-101`). If that width ever changes, this separator silently misaligns with nothing to catch it.

Worth a shared value, e.g. a `$error-outline-width` in the prompt-line SCSS that both files `calc()` off of.

Separately: `z-index: 0` was dropped from this `::after` in the same hunk. It looks inert (the pseudo-element still paints above `__warning-icon-and-text` by tree order), but it's an unrelated edit riding along with the outline fix and it isn't in the changelog.

ethanwinters (approved):

PR #2118, opened 2026-08-10
feat(chat-button): add skeleton state

### LINKED ISSUE #2010: [Feature Request]: Add a skeleton state for chat button
### The problem

Chat button does not have a skeleton state.

### The solution

Add a skeleton state to chat button with stories example.

### Examples

_No response_

### Application / product

_No response_

No

### Business priority

Low Priority = release date is not dependent on fix or not upcoming

_No response_

### Additional contacts

@kiran-ravindra-ibm has designs for skeleton state

### Available extra resources

_No response_

### Code of Conduct

AlexanderMelox:
Closes #2010

Adds `<cds-aichat-button-skeleton>` — a pill-shaped loading placeholder for the chat button. Matches all three size variants.

#### Changelog

**New**

- `cds-aichat-button-skeleton` Lit element — extends `CDSButtonSkeleton`; supports `size` (`sm` / `md` / `lg`)
- `ChatButtonSkeleton` React wrapper
- Skeleton story in WC and React Storybooks

**Changed**

- `ChatButtonSize` type is now exported from `chat-button.ts`
- Skeleton host styles added to `chat-button.scss`
- `index.ts` registers the skeleton element

#### Testing / Reviewing

1. `npm run aiChat:start:storybook` → **Components / Chat button / Skeleton**
2. Toggle `size` — border-radius should match the live button
3. `npm run lint:styles` — 0 errors


packages/ai-chat-components/src/components/chat-button/__stories__/chat-button.stories.js
```diff
@@ -410,3 +402,34 @@ export const QuickActionSelectedWithDisabled = {
...
+export const Skeleton = {
+  argTypes: {
+    size: {
+      control: 'select',
+      description: 'Specify the size of the skeleton button.',
+      options: ['sm', 'md', 'lg'],
+    },
+    isQuickAction: {
+      table: { disable: true },
+    },
```
warrenmblood, line 429: Storybook still registers the `is-quick-action` reflected HTML attribute from the Lit decorator, so this shows in the story controls even though it has no effect on the skeleton. Add this to get rid of it:
```
'is-quick-action': {
    table: { disable: true },
},
```

packages/ai-chat-components/src/components/chat-button/__stories__/chat-button.stories.js
```diff
@@ -410,3 +402,34 @@ export const QuickActionSelectedWithDisabled = {
...
+    ...QuickAction.argTypes,
+  },
+  args: {
+    ...sharedArgs,
+    buttonText: 'Quick action',
+    isQuickAction: true,
+    isSelected: true,
+  },
+  render: baseButtonTemplate,
+};
```
warrenmblood: Let's move this story back before "Skeleton" so it matches the React storybook order.

packages/ai-chat-components/src/react/chat-button-skeleton.ts
```diff
@@ -0,0 +1,25 @@
...
+const ChatButtonSkeleton = withWebComponentBridge(
+  createComponent({
+    tagName: 'cds-aichat-button-skeleton',
+    elementClass: CDSAIChatButtonSkeleton,
+    react: React,
+  })
+);
+
+export default ChatButtonSkeleton;
+export { CHAT_BUTTON_SIZE };
```
warrenmblood: We don't need to export `CHAT_BUTTON_SIZE` since this enum value isn't actually used by the skeleton button. The standard pattern from `@carbon/web-components` is to only export these values from the base component, in this case `cds-aichat-button>`. Import on line 14 can also be removed.

warrenmblood (approved):
Looks good. No blockers, just a couple nit picks.

PR #2119, opened 2026-08-10
fix(markdown): namespace custom renderer slot names per element

### LINKED ISSUE #2099: [Bug]: customRenderers.codeBlock renders duplicate blocks in the first message and missing blocks in subsequent messages
### Browser

_No response_

### Package version

1.18.0

_No response_

### Summary

When `customRenderers.codeBlock` (or `customRenderers.table`) is configured via the `markdown` prop, slot-based custom renderers behave incorrectly across messages:

- **First message:** each custom-rendered block (code block, table) appears **duplicated** - multiple instances rendered where only one should be.
- **Subsequent messages:** the custom-rendered blocks are **missing** entirely - the surrounding text renders but the slot content is absent.

### Root cause

Carbon AI Chat internally mounts **two separate `<cds-aichat-markdown>` Lit element instances** for each committed message. The slot name used to identify a custom-rendered block is derived only from the token's start line number and its sibling index - with no reference to which element instance is rendering:

```js
// markdown-renderer.js
function slotNameFor(kind, token, options) {
    const startLine = token.map?.[0] ?? 0;
    const index = options.context?.currentIndex ?? 0;
    return `cds-aichat-markdown-renderer-${kind}-${startLine}-${index}`;
}
```

### Reproduction / example

https://stackblitz.com/edit/github-wumpmhfv

### Steps to reproduce

1. Configure `customRenderers.codeBlock` via the `markdown` prop:

```tsx
markdown={{
  customRenderers: {
    codeBlock: ({ slotName, code, language }) => (
      <div slot={slotName}>{code}</div>
    ),
  },
}}
```

2. Add two messages each containing a fenced code block:

```ts
instance.messaging.addMessage({
  output: {
    generic: [{
      response_type: MessageResponseTypes.TEXT,
      text: "```json\n{ \"name\": \"Alice\", \"age\": 30 }\n```",
    }],
  },
});
instance.messaging.addMessage({
  output: {
    generic: [{
      response_type: MessageResponseTypes.TEXT,
      text: "```json\n{ \"name\": \"Bob\", \"age\": 25 }\n```",
    }],
  },
});
```

**Expected:**
- First message: one code block rendered.
- Second message: one code block rendered.

**Actual:**
- First message: **two** duplicate code block shells rendered.
- Second message: **no** code block rendered - slot content is missing.

### Suggested severity

None

### Application / product

_No response_

### Code of Conduct

ethanwinters:
Closes #2099

Slot names were unique only within one `cds-aichat-markdown` element, but slot hosts are hoisted into one shared page-level container and projected back by name. Two messages that each opened with a fenced code block at line 0 minted the same name, so the first message's slot gathered both hosts and the second gathered none. Each element now mints a namespace at construction and appends it to every slot name.

The issue blames two markdown elements per message. That part is wrong — there is one, and the collision is between messages.

- `markdown/src/utils/slot-names.ts` — new. The counter sits on `globalThis` so the `es` and `es-custom` builds share one sequence, and the namespace is a suffix so `[slot^="cds-aichat-markdown-renderer-"]` selectors keep matching. A prefix breaks six existing assertions.
- `markdown/__tests__/markdown.test.ts` — the new harness mirrors the real containers: `preventDefault`, relocate the host, render one `<slot>` forwarder per name deduped by name. That dedupe is what starves the second message pre-fix.

Two commits, readable apart: the fix, then the examples.

#### Changelog

**New**

- Per-element namespace on every custom-renderer and plugin-fallback slot name.
- Four regression tests: distinct names per element, one projected host per element, plugin-fallback isolation across unmount, namespace survives reconnect.

**Changed**

- `slotName` is documented as unique per page and opaque. Format was never part of the API.
- The four markdown examples answer each turn with two messages, so the repeat case shows on load.
- The three chat containers keep name-keyed host maps, now with a comment naming the invariant they depend on. No behavior change — projection is name-based, so owner-keying would not prevent the failure.

#### Testing / Reviewing

Not reachable from the demo site — it has no `markdown` config. Use the examples.

```bash
npm run test  --workspace=@carbon/ai-chat-components
npm run test  --workspace=@carbon/ai-chat
```

The new tests bite. Comment out `slotNamespace: this.slotNamespace` in `markdown.ts` and three fail on all three engines, the first with `expected 'cds-aichat-markdown-renderer-table-0-0' to not equal 'cds-aichat-markdown-renderer-table-0-0'`.

Then, per example, confirm both messages render their own blocks:

```bash
npm run start --workspace=@carbon/ai-chat-examples-react-markdown-override
npm run start --workspace=@carbon/ai-chat-examples-web-components-markdown-plugin
```

Pre-fix, message one shows doubled blocks and message two shows none. Also check `document.querySelectorAll('[slot^="cds-aichat-markdown-renderer-"]')` — one host per block, names ending in different `-e<n>` suffixes.

Typed `fix`. No type member added, removed, or retyped, and the values were already unstable across page loads.

#2070 touches `markdown.ts` a few lines from the `slotNamespace` addition and needs a small rebase on top.


packages/ai-chat-components/src/components/markdown/__tests__/markdown.test.ts
```diff
@@ -1430,6 +1430,281 @@ HTTP: http://example.com
...
+          host.setAttribute('slot', detail.slotName);
+          if (host.parentElement !== harness) {
+            harness.appendChild(host);
+          }
+
+          if (!forwarderOwners.has(detail.slotName)) {
+            forwarderOwners.set(detail.slotName, owner);
+            const forwarder = document.createElement('slot');
+            forwarder.setAttribute('name', detail.slotName);
+            forwarder.setAttribute('slot', detail.slotName);
```
AlexanderMelox, line 1600: **Important** — `forwarder.setAttribute('slot', detail.slotName)` is a no-op here. The `<slot>` forwarder is appended to `owner` (a `cds-aichat-markdown` element) as a light-DOM child. The `slot` attribute routes a node into a *parent* shadow host's named slot — but `owner` is mounted inside the harness's shadow root, not slotted into any outer host, so the attribute has no effect. The actual projection tested by `owner.shadowRoot.querySelector('slot[name=...]').assignedElements()` at line 1632 comes entirely from the named shadow-root `<slot>` rendered by the markdown element's Lit template.

The test still passes because the assertion path is correct. But the attribute creates a false impression that the forwarder is doing work. Consider removing the line, or documenting why it would be needed in the real container context but is a no-op in the test:
```ts
// Real containers append the forwarder as a slotted child of the markdown
// element in the page tree; the `slot` attribute routes it onward through the
// outer host's shadow boundary. In this test the markdown element is mounted
// directly in the harness shadow root (not slotted), so this attribute is a
// no-op — remove it or keep it with a comment.
```

packages/ai-chat-components/src/components/markdown/src/utils/slot-names.ts
```diff
@@ -0,0 +1,77 @@
...
+ * counting from zero.
+ *
+ * The name is deliberately underscore-prefixed and camelCase: the `es-custom`
+ * build rewrites the bare tokens `cds` and `cds-aichat` in every emitted file
+ * (`tasks/build.js`), and neither word-boundary regex can match inside
+ * `__cdsAIChat…`, so both builds agree on the key.
+ *
+ * @internal
+ */
+const INSTANCE_SEQUENCE_KEY = '__cdsAIChatMarkdownInstanceSequence';
```
AlexanderMelox, line 36: **Nit** — `INSTANCE_SEQUENCE_KEY` is a module-level `const` (not exported), so it is dead code to the outside, but its comment explains the `globalThis` choice and the build rewrite constraint — both are genuinely non-obvious. The comment is long but earns its keep. One small grammar fix: "the `es` and `es-custom` builds are separate module scopes with separate class definitions, and would otherwise each start counting from zero" — the trailing clause ("and would otherwise…") could be its own sentence: "…separate class definitions. Without `globalThis`, each build would start counting from zero."

packages/ai-chat-components/src/components/markdown/src/utils/slot-names.ts
```diff
@@ -0,0 +1,77 @@
...
+ * shared counter, and short deterministic names stay readable in DevTools and
+ * in test output.
+ *
+ * @internal
+ */
+export function nextMarkdownInstanceId(): string {
+  const scope = globalThis as unknown as Record<string, number | undefined>;
+  const next = (scope[INSTANCE_SEQUENCE_KEY] ?? 0) + 1;
+  scope[INSTANCE_SEQUENCE_KEY] = next;
+  return `e${next.toString(36)}`;
```
AlexanderMelox, line 57: **Nit** — `e${next.toString(36)}` produces `e1`, `e2`, …, `e9`, `ea`, `eb`, …, `ez`, `e10`, …  The `e` prefix is a collision guard (prevents a bare digit-only string from looking like a CSS id selector) and is sound. One doc note worth adding: the `toString(36)` base-36 encoding means namespace identifiers can be confused with hex color values in long sessions (`e3f` etc.). Not a defect — just something to name in the comment so future readers don't assume decimal.

AlexanderMelox (commented):
## Review — `fix(markdown): namespace custom renderer slot names per element`

This is a precise, well-scoped fix. The root cause diagnosis is correct (page-level host container + name-only keying = cross-element collision), the chosen mechanism (construction-time namespace suffix on `globalThis`) is the right shape for the es/es-custom dual-build constraint, and the four regression tests exercise the failure mode directly. The PR description is exceptionally clear and worth citing as a model.

One **Important** finding on the test harness: `forwarder.setAttribute('slot', detail.slotName)` gives the harness's `<slot>` forwarder a `slot` attribute so it can re-project through the harness's own shadow DOM — but `owner` (the `cds-aichat-markdown` element) is mounted **inside the harness's shadow root** (`harness.shadowRoot?.appendChild(el)`), not in the harness's light DOM. A `slot` attribute on a child of `owner` only has meaning when `owner` is slotted into another shadow host, which it is not here. The harness works because the real markdown element renders the named shadow-root `<slot>` placeholders itself (from the renderer output), so the `slot` attribute on the appended `forwarder` element is a no-op and does not affect what `owner.shadowRoot.querySelector('slot[name=...]').assignedElements()` returns. The code passes, but the `forwarder.setAttribute('slot', detail.slotName)` line is misleading — it looks like it is wiring re-projection when it is actually dead.

Two **Nit**-level documentation items and a note on examples.

No blockers. Ready to approve once the team decides whether to fix the no-op `slot` attribute in the harness.

packages/ai-chat-components/src/components/markdown/__tests__/markdown.test.ts
ethanwinters: I checked this one against the runtime rather than by reading, and the premise doesn't hold — `slot` is load-bearing, and removing it breaks the test.

`owner` doesn't need to be slotted into an outer host, because **`owner` is itself a host**. `cds-aichat-markdown` renders `<slot name="…">` in its own shadow root, so a light-DOM child of `owner` carrying `slot="…"` is assigned into that slot. The forwarder needs both attributes, for two different reasons:

- `name` — the forwarder is itself a slot in the *harness's* shadow tree (it descends from `owner`, which lives in `harness.shadowRoot`), so it gathers the relocated host back out of the harness light DOM.
- `slot` — assigns the forwarder into the markdown element's own shadow slot, which is the second hop `assignedElements({ flatten: true })` resolves through.

Instrumented, with the line removed:

```
DIAG assigned.length 1 isHost false tag CDS-AICHAT-CARD parent SLOT
```

`assigned.length` stays 1, but it is the shadow slot's **fallback content** (the default `cds-aichat-card` code-block rendering), not the hoisted host — `flatten` returns fallback children when nothing is assigned. So `expect(assigned[0]).to.equal(host)` fails.

It fails expensively, too: chai stringifies both DOM subtrees to build the failure message, and the tab locks up hard enough that mocha's own per-test timer never fires. All three engines hit the 300s `testsFinishTimeout` with `0 passed, 0 failed`, against 82 passed in 9.1s with the line in. I reproduced that twice and then bisected it to this single test.

The harness isn't inventing the shape either — it mirrors what all three real containers render verbatim:

- `cds-aichat-container/index.ts:692` — `<slot name=${slot} slot=${slot}></slot>`
- `cds-aichat-custom-element/index.ts:411` — same
- `ai-chat-components/src/react/markdown.tsx:328` — `<slot key={slotName} name={slotName} slot={slotName} />`

You're right that it deserved a comment, though — it read as a no-op to you and would to the next reader. I've added one at the mint site rather than removing the attribute.

packages/ai-chat-components/src/components/markdown/src/utils/slot-names.ts
ethanwinters: Applied verbatim — the trailing clause is now its own sentence: "…separate class definitions. Without `globalThis`, each build would start counting from zero."

packages/ai-chat-components/src/components/markdown/src/utils/slot-names.ts
ethanwinters: Good catch on the encoding — documented. The comment now spells out that the count is base-36 rather than decimal, with the `e1`, `e2`, … `e9`, `ea`, … `ez`, `e10` sequence written out.

I didn't encode the stated rationale for the `e` prefix, though, because I don't think it holds: these names only ever appear inside quoted attribute values (`[slot^="…"]`, `slot[name="…"]`), never as an id selector, and the base name already ends in numeric position segments — so a digit-leading suffix wouldn't be a new problem. What the `e` verifiably *does* is keep the suffix from reading as one more of those position segments, so that's what the comment says:

> The `e` keeps the suffix from being read as one more of the numeric position segments the base name already ends in (`…-codeBlock-0-0-e3`, not `…-codeBlock-0-0-3`).

Happy to swap in the hex-confusion note as well if you still want it called out.

AlexanderMelox (approved):

PR #2128, opened 2026-08-11
feat(prompt-line): disableDirectSend prop

### LINKED ISSUE #2124: [Prompt-line]: Remove clickable send button inside autocomplete items
The prompt line design review clarified the usage of the send icon within autocomplete items (see below):
Image

In the above example, the suggestion should send directly to the chat regardless of whether the send icon or anywhere else within the suggestion item is clicked.

If a user needs the behavior of populating the editor with a suggestion before sending, then the autocomplete items should only emit the `cds-aichat-autocomplete-select` event and not handle send themselves.

Task:
- Update autocomplete's `enableSendButton` prop to control whether the entire item supports the "send" or "select" path. Rename prop to something more fitting (e.g. `enableDirectSend`)
- Update all tests and documentation to reflect the desired behavior. Because this is being done prior to the stable release of prompt line, we don't need to worry about backwards compatibility.

Note: Fixing this will also clear up an issue where the user cannot navigate to the send button via keyboard navigation

warrenmblood:
Closes #2124

Replaces the interactive `cds-icon-button` send button inside each autocomplete item with a lightweight visual-only send icon, and renames the controlling prop from `enableSendButton` to `disableDirectSend`. Clicking an item now always routes through the item click handler; `disableDirectSend` decides which event fires and whether the icon is shown.

#### Changelog

**Changed**

- `cds-aichat-autocomplete`: `enableSendButton` prop renamed to `disableDirectSend` (attribute `disable-direct-send`)
- `disableDirectSend=false` (default): clicking an item fires `cds-aichat-autocomplete-send` and shows a decorative send icon on hover/focus/active
- `disableDirectSend=true`: clicking an item fires `cds-aichat-autocomplete-select` (insert-into-editor); no send icon rendered
- `BaseSuggestionConfig` and `StartersConfig` gain `disableDirectSend?: boolean`; value flows through `AutocompleteController` → `AutocompleteControllerElement` → `useChatAutocomplete`
- `CustomListProps.onSelect` made optional; examples using starters drop the now-unnecessary `onSelect` wiring
- Send icon is now a non-interactive `<span>` with `pointer-events: none`; removed `cds-icon-button`, `isDirectionRTL` import, and `isRTL` state from `autocomplete.ts`

#### Testing / Reviewing

1. Start Storybook: `npm run storybook --workspace=@carbon/ai-chat-components`
2. Open **Prompt line / Autocomplete / Default**; toggle **disableDirectSend** in controls.
   - `false` (default): hovering an item reveals the send icon; clicking fires `cds-aichat-autocomplete-send` in the actions panel.
   - `true`: no icon; clicking fires `cds-aichat-autocomplete-select`.
3. In the `prompt-line-conversation-starters` example (`PORT=3000 npm start --workspace=...`), confirm clicking a starter sends the message without inserting text into the editor first.
4. Unit tests: `npm test --workspace=@carbon/ai-chat-components`


packages/ai-chat-components/src/components/prompt-line/autocomplete/src/autocomplete.ts
ethanwinters: we should update our lint rules so we catch this before it gets here. Lets get rid of the non-null assertion.

packages/ai-chat-components/src/components/prompt-line/src/tiptap/types.ts
```diff
@@ -109,6 +112,13 @@ export interface BaseSuggestionConfig {
 
   /** Replace the built-in suggestion list UI. */
   renderCustomList?: (props: CustomListProps) => HTMLElement | unknown;
+
+  /**
+   * When `false`, clicking a suggestion item fires `cds-aichat-autocomplete-select`
+   * and inserts the item into the editor rather than sending immediately.
+   * Defaults to `true`.
+   */
+  enableDirectSend?: boolean;
```
ethanwinters: One of the things I'd like to try to cut down on is props that default to true. Having to add something to set it to false always feels sloppy. Can we invert this to default to false and then add it as needed in the composed app?

packages/ai-chat-components/src/components/prompt-line/src/autocomplete-controller.ts
```diff
@@ -751,6 +773,7 @@ class AutocompleteControllerElement extends LitElement {
     return html`
       <cds-aichat-autocomplete
         .items=${items}
+        .enableDirectSend=${enableDirectSend ?? true}
```
ethanwinters: **Blocker** — this default applies direct-send to *every* trigger type. That breaks mention and command insertion for anyone who has not opted out.

To confirm, I ran `examples/react/prompt-line-mentions-and-commands` with the two `enableDirectSend: false` lines removed. That is the config the example has on `main` today. I typed `Hey @Ja` and clicked "Jane Smith":

- the chat sent a message reading **"Jane Smith"**
- the mock backend replied **"No mentions or commands detected in structured data"**. `mention.onSelect` never fired, so the example's `updateStructuredData` call never ran
- the editor still held the half-typed **"Hey @Ja"**, no chip

The path: `_resolveEnableDirectSend()` returns `undefined`, so `?? true` wins. `_handleItemClick` then calls `_handleSend`, which calls `_sendItem`. That dismisses the list and emits `cds-aichat-autocomplete-item-send`. `@carbon/ai-chat` wires the event straight to `onSendInput(text)` (`Input.tsx:506-508`), so the chip-insertion branch at `:319-348` never runs.

That is why all four mention and command examples in this diff needed `enableDirectSend: false`. Those edits absorb a regression rather than demonstrate a feature. Direct-send suits starters and typeahead. Mention and command insert into the editor by definition.

Inverting the default fixes this, per the `types.ts:121` comment. A starter config that wants send opts in, and mention and command keep working untouched.

If the default stays `true`, it must at least exclude mention and command:

```ts
private _resolveEnableDirectSend(): boolean | undefined {
  const trigger = this._trigger;
  if (!trigger || trigger.type === 'mention' || trigger.type === 'command') {
    return false;
  }
  // ...
}
```

packages/ai-chat-components/src/react/hooks/useChatAutocomplete.tsx
```diff
@@ -235,6 +235,7 @@ export function useChatAutocomplete(
         slot="autocomplete-content"
         items={state.items}
         attached={attached}
+        enableDirectSend={state.enableDirectSend ?? true}
```
ethanwinters: **Blocker** — same default as `autocomplete-controller.ts:776`. It carries a second consequence for starters: it bypasses `isSendDisabled`.

`AutocompleteController.select()` gates the starter send behind the flag (`autocomplete-controller.ts:313`):

```ts
if (!this._isSendDisabled) {
  this._onStarterSelected?.(projectRawValue(editor));
}
```

`handleSend` on line 163 has no such gate, and `Input.tsx:506-508` calls `onSendInput(text)` unconditionally.

I confirmed this in `examples/react/prompt-line-conversation-starters`. With `input.isSendDisabled: true` and starters on the built-in list, clicking a starter **sent the message**. The Send button beside it correctly rendered `disabled`. I then added `enableDirectSend: false` to the same starters config and clicked again. That time the text landed in the editor and nothing sent, so the new default is what flips it. Hosts set `isSendDisabled` to hold sends off while a response streams. A starter click should not punch through it.

The direct-send path also skips `setRawInputValue` and `sendCurrentValue()` at `Input.tsx:499-505`. The send no longer runs through `projectRawValue(editor)`, so the turn loses any structured-data sidecar.

Inverting the default fixes the common case. The direct-send path should still honour `isSendDisabled`, since a consumer can opt in explicitly.

packages/ai-chat-components/src/components/prompt-line/autocomplete/src/autocomplete.ts
```diff
@@ -353,6 +349,7 @@ class AutocompleteElement extends LitElement {
       return;
     }
     this._focusedIndex = index;
+    this._announcer.announce(this.i18n.itemInserted(item.label));
```
ethanwinters: **Important** — the direct-send path announces `itemInserted`, but nothing was inserted. The message was sent. A screen-reader user clicking a starter hears "Generate a chart for key metrics inserted." after the message has already gone to the chat.

Adding an announcement is the right instinct; the old icon-button path announced nothing. It just needs its own string. `AutocompleteI18n` requires every user-visible string as a field, and #2124 waives back-compat pre-stable, so add one:

```ts
/** Announced when an item is sent directly to chat. */
itemSent: (label: string) => string;
```

```ts
itemSent: (label) => `${label} sent.`,
```

Then call `this._announcer.announce(this.i18n.itemSent(item.label));` here, and leave `itemInserted` to `_selectItem` at `:407`.

packages/ai-chat-components/src/components/prompt-line/src/tiptap/types.ts
```diff
@@ -78,13 +78,16 @@ export interface CustomListProps {
...
-   * The text is the item's value that was sent.
+   * Callback to invoke when user sends an item directly to chat, bypassing
+   * the editor.
    */
   onSend: (text: string) => void;
+  /**
+   * Callback to invoke when user selects an item to insert into the editor
+   * without sending to chat. Optional.
+   */
+  onSelect?: (item: SuggestionItem) => void;
```
ethanwinters: **Important** — making `onSelect` optional weakens the contract, and no producer omits it. Both call sites always pass it: `autocomplete-controller.ts:755` and `useChatAutocomplete.tsx:206`.

The only effect is a null-check on every consumer, which this diff then absorbs. `CustomSuggestionList.tsx:73,100,103` and `custom-suggestion-list.ts:109` all grew `?.` or `| undefined` for a callback that always arrives. A consumer who ignores `onSelect` just leaves it out of the destructure. That already worked when it was required.

Suggest restoring `onSelect: (item: SuggestionItem) => void;` and dropping the example edits that follow from it. `enableDirectSend` on `BaseSuggestionConfig` at `:121` is a separate question — see the comment there.

packages/ai-chat-components/src/components/prompt-line/__stories__/prompt-line-react.stories.jsx
```diff
@@ -403,9 +403,9 @@ const CommandsAndMentionsStory = ({
   React.useEffect(() => {
     const el = shellRef.current?.querySelector('cds-aichat-autocomplete');
     if (el) {
-      el.enableSendButton = enableSendButton;
+      el.enableDirectSend = enableDirectSend ?? false;
```
ethanwinters: **Important** — this pushes `enableDirectSend` onto an element the controller owns and re-renders. `useChatAutocomplete` sets the same prop on every render (`useChatAutocomplete.tsx:238`). The assignment here only sticks because the effect re-runs on `autocompleteContent`. Any render that leaves that dep alone silently reverts the control.

The examples in this diff show the right wiring. Put it on the trigger config, which is the documented public API:

```js
const mention = useMemo(
  () => ({ trigger: '@', enableDirectSend, items: ..., onSelect: ... }),
  [enableDirectSend]
);
```

That also drops the `?? false` here. It duplicates `args: { enableDirectSend: false }` at `:458`, and it contradicts the `?? true` used at `:478` and `:774` for the same control.

examples/react/prompt-line-conversation-starters/README.md
```diff
@@ -6,7 +6,7 @@
 
 - Using `input.expanded: true` so the editor occupies its own full-width row and the action buttons render inline beneath it.
 - Configuring `input.starters` so the suggestion list appears immediately when the editor is focused and empty — no typing required. Selecting a starter inserts the text and auto-sends in one action.
```
ethanwinters: **Important** — this bullet still describes the old flow, as does the intro at line 3: "Selecting a starter inserts the text and auto-sends in one action." Direct send never inserts the text. `_sendItem` dismisses the list and emits, leaving the editor untouched. I watched this in the running example: the editor stays empty and the message goes straight out. Only the third bullet was updated.

Suggested replacement:

> Selecting a starter sends it straight to the chat, skipping the editor.

The web-components twin carries the same stale line in both places (`examples/web-components/prompt-line-conversation-starters/README.md:3,8`).

packages/ai-chat-components/src/components/prompt-line/autocomplete/src/autocomplete.ts
```diff
@@ -432,10 +429,14 @@ class AutocompleteElement extends LitElement {
   }
 
   private _handleItemClick(index: number) {
```
ethanwinters, line 440: **Nit** — the `else` branch sets `_focusedIndex`, and `_handleSend` sets it again at `:351`. The null guard is duplicated too. This reads flatter:

```ts
private _handleItemClick(index: number) {
  if (this.enableDirectSend) {
    this._handleSend(index);
    return;
  }
  const item = this._getItemAtIndex(index);
  if (!item) {
    return;
  }
  this._focusedIndex = index;
  this._selectItem(item);
}
```

packages/ai-chat-components/src/components/prompt-line/autocomplete/src/autocomplete.scss
```diff
@@ -157,20 +157,24 @@ $group-class: #{$item-class}-group;
     color: theme.$text-helper;
   }
 
-  .#{$item-class}__send {
+  .#{$item-class}__send-icon {
     z-index: 1;
+    display: flex;
+    align-items: center;
+    align-self: center;
+    padding-inline: layout.$spacing-03;
```
ethanwinters: **Nit** — worth checking against the design spec in #2124. I measured this in Storybook, on Prompt line / Autocomplete / Default. `$spacing-03` puts the send glyph 8px from the item's trailing edge. The label sits on a 16px leading gutter (`$spacing-05`). The icon reads tighter to the edge than the text does.

The `cds-icon-button` it replaced was 40px wide with the 16px glyph centred, so about 12px. `$spacing-05` here would make the two gutters symmetric, if that matches the mock.

packages/ai-chat-components/src/components/prompt-line/autocomplete/__stories__/autocomplete.mdx
```diff
@@ -39,9 +39,7 @@ Here's a quick example to get you started.
   .items=${[
     { id: "1", label: "When is the best time to eat?" },
     { id: "2", label: "When is the sun rising today?" }
-  ]}
-  attached
-  enable-send-button>
+  ]}>
```
ethanwinters: **Nit** — correcting my own reasoning here; my first pass overstated this.

I'd written that `enable-direct-send="false"` in markup is a no-op. That's true, but it's just how boolean attributes work everywhere — `<input disabled="false">` is disabled too — so it isn't a defect in this component, and the JS property (`el.enableDirectSend = false`) works fine. Consumers get this element from the controller or build it with `createElement`, so the markup path is mostly hypothetical anyway.

What's actually left is small: because the property defaults to `true`, the attribute can never express "off", so it's inert as an input and only reflects outward. This page uses the attribute spelling when it says "When `false`, clicking inserts the item into the editor", which points a reader at the one spelling that can't produce `false`. Either mention that consumers set the property, or invert the default and the attribute starts carrying its weight. Not blocking either way.

The live version of this problem is in `autocomplete.stories.js` — see the separate comment there.

Separately, and unrelated to the above: the event table below says send fires when the item is "clicked/activated". Say that this sends the message immediately. A reader should not have to infer it.

packages/ai-chat-components/src/components/prompt-line/autocomplete/__stories__/autocomplete-react.mdx
```diff
@@ -50,11 +50,6 @@ function MyComponent() {
   return (
     <CDSAIChatAutocomplete
       items={suggestions}
```
ethanwinters, line 52: **Nit** — the getting-started snippet dropped `attached` along with the send props. `attached` has nothing to do with this rename, and it defaults to `true`, which suppresses bottom corner rounding. That is the less obvious of the two states, so showing it helped. Suggest keeping `attached={true}`.

ethanwinters (changes requested):
Reviewed everything except the two comments already on `autocomplete.ts:385` and `types.ts:121`.

The core swap is right. The interactive `cds-icon-button` is gone, the item is the only click target, and the keyboard trap from #2124 is fixed. The Lit suite passes, 38/38 across Chromium, Firefox, and WebKit.

Both blockers trace to one thing: `enableDirectSend` defaults to `true`. The controller applies that default to every trigger type, not just starters (`autocomplete-controller.ts:776`, `useChatAutocomplete.tsx:238`). I built the packages and ran the examples to confirm. The line comments have the detail; here is the short version.

**Mentions are broken.** I removed the two `enableDirectSend: false` lines this PR adds to `examples/react/prompt-line-mentions-and-commands`. That is the config the example has on `main` today. Then I typed `Hey @Ja` and clicked "Jane Smith":

- the chat sent a message reading **"Jane Smith"**
- the mock backend replied **"No mentions or commands detected in structured data"**. `onSelect` never fired, so `updateStructuredData` never ran
- the editor still held the half-typed **"Hey @Ja"**, no chip

**Starters bypass `isSendDisabled`.** I set `input.isSendDisabled: true` and put starters on the built-in list. Clicking a starter sent the message, while the Send button beside it correctly rendered `disabled`. I then added `enableDirectSend: false` to the same config and clicked again. That time the text landed in the editor and nothing sent. The default is what flips it.

Both findings reinforce the invert-the-default comment on `types.ts:121` rather than restating it. Inverting also fixes the attribute problem noted on `autocomplete.mdx`.

One retraction: ignore any earlier note about `telemetry.yml`. `release-base.yml` regenerates and commits it on full release. It self-heals, and it is not yours to update.

**Findings without a diff line to anchor to:**

- **Important** — Two JSDoc blocks still say "send button", but no button exists now. See `useChatAutocomplete.tsx:55` on `onSendItem`, and `autocomplete-controller.ts:617` on `@fires cds-aichat-autocomplete-item-send`.
- **Important** — `prompt-line.stories.js:604` and `:700` push the prop imperatively, same as the React story below.

**Test coverage — could you open a follow-up issue instead of growing this PR?**

The gaps are worth tracking, but they should not hold up the fix:

- `autocomplete-controller.ts` gained `_resolveEnableDirectSend()`, which resolves across four config types. It also gained a state field and a property binding. Neither `autocomplete-controller.test.ts` nor `autocomplete-controller-element.test.ts` changed.
- Coverage dropped at `autocomplete.test.ts:219`. The old grouped test asserted that `enableSendButton=false` reached grouped items. The replacement only covers the default-true case, so nothing checks that grouped items honour `enableDirectSend=false`.
- Nothing covers Enter-to-activate with `enableDirectSend=false`; only the click path at `:169`. Nothing covers the announcement text either, which is where the `itemInserted` finding hides.
- Please note in the issue that the tests belong at the controller level, not just the element. A mention click must insert a chip. A starter click must respect `isSendDisabled`. Those two cases would have caught both regressions.

packages/ai-chat-components/src/components/prompt-line/src/autocomplete-controller.ts
```diff
@@ -513,11 +515,27 @@ export class AutocompleteController {
...
+    if (!trigger || trigger.type === 'mention' || trigger.type === 'command') {
+      return true;
+    }
+    if (trigger.type === 'starter') {
+      if (this._isSendDisabled) {
+        return true;
+      }
+      return this._starters?.disableDirectSend;
+    }
+    return this._autocomplete?.disableDirectSend;
```
ethanwinters: **Blocker** — this gate only reaches the built-in `<cds-aichat-autocomplete>`, so a `renderCustomList` consumer still sends on click while the Send button renders disabled. Reproduced in `examples/react/prompt-line-conversation-starters` with `input.isSendDisabled: true`: clicking a starter posted the message and the mock backend replied, and the rendered list carried `disableDirectSend: false` with four send icons. Line 530 leaves the same hole for the `autocomplete` trigger — move the check into `_sendItem()`, which every send path goes through.

With `renderCustomList` removed from that example the built-in path behaves correctly, which is what narrows this to the custom-list route.

packages/ai-chat-components/src/components/prompt-line/__stories__/prompt-line.mdx
```diff
@@ -281,16 +281,16 @@ Here's a minimal implementation.
...
       <td><code>cds-aichat-autocomplete-send</code></td>
-      <td>Fired when the send button inside an item is clicked.</td>
+      <td>Fired when an item is clicked or activated via keyboard and <code>enable-direct-send</code> is <code>true</code> (default). The item text is sent directly to chat.</td>
       <td>• <code>text</code> (string)</td>
     </tr>
+    <tr>
+      <td><code>cds-aichat-autocomplete-select</code></td>
+      <td>Fired when an item is clicked or activated via keyboard and <code>enable-direct-send</code> is <code>false</code>. Use this when the item should be inserted into the editor rather than sent immediately.</td>
+      <td>• <code>item</code> (SuggestionItem)</td>
+    </tr>
```
ethanwinters, line 293: **Blocker** — this table still names `enable-direct-send` and its polarity is inverted against the shipped prop, so a Web Components reader wires up an attribute that does not exist. The React table at `prompt-line-react.mdx:333` was updated; this one was missed.

```suggestion
    <tr>
      <td><code>cds-aichat-autocomplete-send</code></td>
      <td>Fired when an item is clicked or activated via keyboard and <code>disable-direct-send</code> is <code>false</code> (default). The item text is sent directly to chat.</td>
      <td>• <code>text</code> (string)</td>
    </tr>
    <tr>
      <td><code>cds-aichat-autocomplete-select</code></td>
      <td>Fired when an item is clicked or activated via keyboard and <code>disable-direct-send</code> is <code>true</code>. Use this when the item should be inserted into the editor rather than sent immediately.</td>
      <td>• <code>item</code> (SuggestionItem)</td>
    </tr>
```

packages/ai-chat-components/src/components/prompt-line/src/tiptap/types.ts
```diff
@@ -109,6 +112,13 @@ export interface BaseSuggestionConfig {
 
   /** Replace the built-in suggestion list UI. */
   renderCustomList?: (props: CustomListProps) => HTMLElement | unknown;
+
+  /**
+   * When `true`, clicking a suggestion item fires `cds-aichat-autocomplete-select`
+   * and inserts the item into the editor rather than sending immediately.
+   * Defaults to `false`.
+   */
+  disableDirectSend?: boolean;
```
ethanwinters, line 122: **Important** — `TriggerSuggestionConfig` extends this interface, but `_resolveDisableDirectSend()` hard-returns `true` for `mention` and `command`, so setting `disableDirectSend` on either config does nothing and the "Defaults to `false`" line above is wrong for them. Either honor the value for those triggers, or drop the field from `TriggerSuggestionConfig` via `Omit<BaseSuggestionConfig, 'disableDirectSend'>` and say here that mention and command always insert.

ethanwinters (commented):
Fix blockers before merge.

Mention and command now insert instead of sending - that blocker is fixed, confirmed in the browser. The second one is not: `isSendDisabled` still only guards the built-in list, so a `renderCustomList` consumer sends while the Send button renders disabled. One doc table still names the old attribute.

ethanwinters (approved):
Left three things to be fixed before you merge.

PR #2129, opened 2026-08-11
fix: toolbar overflow after 2

### LINKED ISSUE #1832: [Feature Request] [Toolbar] Only Show Overflow Menu When It Contains Multiple Actions
### The problem

Currently, at certain viewport widths, the overflow menu appears even when it contains only a single action. Although this behavior is based on the available width, the user experience would be improved if the overflow menu only appeared when there are two or more actions to display.
When only one action is moved into the overflow, it doesn't meaningfully reduce the toolbar's width and instead adds an unnecessary click for users to access that action.

https://github.com/user-attachments/assets/c3558029-de4d-47b4-bb39-1df33f885015

### The solution

_No response_

### Examples

_No response_

### Application / product

_No response_

None

### Business priority

None

_No response_

### Additional contacts

_No response_

### Available extra resources

_No response_

### Code of Conduct

davidmenendez:
Closes #1832 

Updates the overflow handler logic in `toolbar.ts` so that the overflow menu doesn't show up until more than one item needs to be hidden, because it doesn't make sense to just replace one hidden item with the overflow menu

#### Changelog

**Changed**

- `packages/ai-chat-components/src/components/toolbar/src/toolbar.ts` updates the overflow logic

#### Testing / Reviewing

go to the `toolbar` story in storybook - resize the window until the overflow menu shows up - verify that when it appears that there are two items in the overflow menu

### DIFF
--- a/packages/ai-chat-components/src/components/toolbar/src/toolbar.ts
+++ b/packages/ai-chat-components/src/components/toolbar/src/toolbar.ts
@@ -285,10 +285,17 @@ class CDSAIChatToolbar extends LitElement {
   }
 
   render() {
-    const { hiddenActions, visibleActions } = this.getActions();
-    const showOverflowMenu = hiddenActions.length > 0;
+    const {
+      hiddenActions: rawHiddenActions,
+      visibleActions: rawVisibleActions,
+    } = this.getActions();
+    const showOverflowMenu = rawHiddenActions.length > 1;
+    const visibleActions = showOverflowMenu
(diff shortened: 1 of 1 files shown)


ethanwinters:
@AlexanderMelox assigned you as the review for this instead since you also have a toolbar PR out

AlexanderMelox (approved):

AlexanderMelox:
@ethanwinters I checked and shouldn't conflict with workspace shell work 👍🏻

PR #2172, opened 2026-08-13
fix: maintain space around chips

### LINKED ISSUE #2155: [Bug]: Sent message bubble drops spaces around mention and command chips
## Background

Release blocker. The sent message bubble glues text to mention and command chips, so the message the user sees is not the message they composed.

Reproduce on `examples/react/prompt-line-mentions-and-commands`, both directions:

| Composed         | Bubble renders  | Backend receives   |
| ---------------- | --------------- | ------------------ |
| `@Bob Chen hi!`  | `Bob Chenhi!`   | `@Bob Chen hi!`    |
| `hey @Bob Chen`  | `heyBob Chen`   | `hey @Bob Chen`    |

The editor doc is correct — a `mention` node followed by a text node `" hi!"` — and the outbound text is correct. Only the rendered bubble is wrong.

### Root cause

`renderParagraphInline` coalesces consecutive text nodes into a run and flushes each run through `renderInlineMarkdown` (`MessageRichUserContent.tsx:166-172`). That helper parses with `markdownToMarkdownItTokens(text, false)` (`InlineMarkdown.tsx:33-38`), which runs markdown-it's **block** parser. A markdown paragraph strips its own leading and trailing whitespace, so every run that sits next to a chip loses the space on that boundary.

A chip splits a paragraph into separate runs, so the split is exactly where the whitespace gets eaten. Paragraphs with no chips are unaffected: they take the `onlyTextual` path (`MessageRichUserContent.tsx:88`), get joined into one string, and are parsed once, so interior spaces survive.

### Consequences

- Any message mixing a chip with surrounding text renders wrong, which is the entire point of mentions and commands.
- The accessible text is glued too. The chip carries `role="img"` with `aria-label="Bob Chen"`, so a screen reader reads the run and the chip with no separation.

## Goal

The sent bubble preserves whitespace exactly as composed, including on both sides of every chip.

## Acceptance criteria

- [ ] `@Bob Chen hi!` renders with a space between the chip and `hi!`.
- [ ] `hey @Bob Chen` renders with a space between `hey` and the chip.
- [ ] Consecutive chips separated by spaces keep their separators.
- [ ] Inline markdown inside a run still renders — bold, emphasis, inline code, links.
- [ ] The chip-free paragraph path is unchanged.
- [ ] Coverage compares the bubble's rendered text against the composed raw text for a paragraph mixing chips and text, so a regression fails on content rather than on markup shape.

## Approach notes

The run is handed to a block-level parse that trims by contract, so the fix belongs at the boundary rather than inside the markdown helper. Two options: capture each run's leading and trailing whitespace before parsing and re-emit it around the rendered fragment, or parse runs with an inline-only entry point that does not trim. A plain space in a JSX text node next to an element renders normally, so no `&nbsp;` is needed.

## Out of scope

- The chip not showing the `@` trigger character. That is driven by `resolveShowTriggerInChip` and is configuration, not a defect.
- `role="img"` on the chip element itself.
- #2154 — autocomplete send not clearing the input.

## Related

- `packages/ai-chat/src/chat/components-legacy/MessageRichUserContent.tsx:166-172` — run coalescing and flush.
- `packages/ai-chat/src/chat/components-legacy/MessageRichUserContent.tsx:88` — the chip-free path that stays correct.
- `packages/ai-chat/src/chat/components/helpers/InlineMarkdown/InlineMarkdown.tsx:33-38` — the trimming block parse.
- `examples/react/prompt-line-mentions-and-commands` — reproduction.

davidmenendez:
Closes #2155

adds logic to maintain the space around command and mention chips

#### Changelog

**New**

- `packages/ai-chat/tests/components/spec/messageRichUserContentChipSpacing_spec.tsx` test for chip spacing logic

**Changed**

- `packages/ai-chat/src/chat/components-legacy/MessageRichUserContent.tsx` adds logic to check for whitespace, remove it from the `textRun`, and create fragments for it so that it doesn't get stripped internally

#### Testing / Reviewing

`npm run test` and verify that the behavior laid out in the issue is resolved

### DIFF
--- a/packages/ai-chat-components/src/components/prompt-line/src/tiptap/render-token-chip.ts
+++ b/packages/ai-chat-components/src/components/prompt-line/src/tiptap/render-token-chip.ts
@@ -150,8 +150,6 @@ function createTokenContainer(
   dom.setAttribute('data-token-type', type);
   dom.setAttribute('data-token-context', context);
   dom.setAttribute('data-raw-value', value ?? label ?? '');
-  dom.setAttribute('role', 'img');
-  dom.setAttribute('aria-label', label || value || '');
   dom.className = 'cds-aichat--token';
   return dom;
 }
--- a/packages/ai-chat/src/chat/components-legacy/MessageRichUserContent.tsx
+++ b/packages/ai-chat/src/chat/components-legacy/MessageRichUserContent.tsx
@@ -167,9 +167,36 @@ function renderParagraphInline(
     if (!textRun) {
       return;
     }
-    out.push(
-      <React.Fragment key={key}>{renderInlineMarkdown(textRun)}</React.Fragment>
+    // The block-level markdown parser trims leading/trailing whitespace from
+    // each run it parses. When a chip splits a paragraph into separate runs,
+    // the space between the chip and the adjacent text gets eaten. Capture the
+    // boundary whitespace explicitly and emit it as plain text siblings so the
+    // rendered bubble matches what the user composed.
+    console.log(textRun);
+    const leading = textRun.match(/^\s+/)?.[0] ?? '';
+    const afterLeading = textRun.slice(leading.length);
+    const trailing = afterLeading.match(/\s+$/)?.[0] ?? '';
+    const trimmed = afterLeading.slice(
+      0,
+      afterLeading.length - trailing.length
     );
+    if (leading) {
+      out.push(
+        <React.Fragment key={`${key}-ws-pre`}>{leading}</React.Fragment>
+      );
+    }
+    if (trimmed) {
+      out.push(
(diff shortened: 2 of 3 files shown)


warrenmblood (changes requested):
I noticed in VoiceOver the chip text is announced twice because `aria-label` and `role="img"` are set. These are left over from when a different element was used for the token chips. These can be removed from `packages/ai-chat-components/src/components/prompt-line/src/tiptap/render-token-chip.ts` lines 153-154:
```
dom.setAttribute('role', 'img');
dom.setAttribute('aria-label', label || value || '');
```

Otherwise, looks good to me!

warrenmblood (approved):

Issue #2152, opened 2026-08-12
[Bug]: Runtime config updates recreate the prompt-line editor and truncate undo history


ethanwinters:
### Background

Release blocker. A host config update while the user is typing destroys and recreates the live Tiptap editor. That silently truncates undo history.

Reproduce on `examples/react/prompt-line-conversation-starters`. Focus the editor, type `hello world`, then press undo. The field goes to `"h"`. Press undo again and nothing happens. History is truncated at the first keystroke, so the user can never undo back to an empty field.

The chain, each step measured in the running example:

1. The example's action carries `disabled: inputHasText`, so the first keystroke rebuilds the host config.
2. The `extensions` array handed to `cds-aichat-prompt-line` gets a new identity. The inner extension object is new too, so the memo in `useInputExtensions` recomputed.
3. [prompt-line.ts](https://github.com/carbon-design-system/carbon-ai-chat/blob/main/packages/ai-chat-components/src/components/prompt-line/src/prompt-line.ts) sees `extensions` changed by identity and calls `setExtensions`.
4. [prompt-line-rich-runtime.ts](https://github.com/carbon-design-system/carbon-ai-chat/blob/main/packages/ai-chat-components/src/components/prompt-line/src/prompt-line-rich-runtime.ts) calls `_recreateEditor()` unconditionally.
5. The rebuilt editor installs a fresh `UndoRedo`, so history restarts from the reseeded content.

The element itself survives. Same node, still connected. So this is the extensions path, not a remount. The editor captured before the update throws "the editor view is not available" afterwards.

#### Scope

Only hosts that update config at runtime are affected. Check `examples/react/prompt-line-typeahead`, which has a static config. Typing there leaves the editor instance and the `extensions` array identity untouched. The framework does not churn on its own.

A second trigger surfaced while fixing this, and it is in scope too. Teardown ran synchronously in `disconnectedCallback`. So a reparent destroyed the editor and its undo history the same way, as did a framework remount that lasted one frame.

A reattach after a genuine teardown was worse. `firstUpdated` is a one-shot, so the element came back with no controller and silently ignored every prop. That second bug predates this issue.

#### Why this is a framework defect

The codebase already treats "a config update rebuilds the editor" as the thing to prevent. Each guard says so:

- `reconcileAppConfigReferences` is a structural-sharing pass. Its stated purpose is keeping value-equal sub-objects on their previous reference.
[...]

PR #2184, opened 2026-08-18
fix(prompt-line): keep the editor across config updates and DOM moves

ethanwinters:
Part of #2152

A runtime config update rebuilt an equivalent Tiptap extension set. `setExtensions` then recreated the live editor on any reference change, dropping undo history mid-typing.

Teardown ran synchronously in `disconnectedCallback`. So reparenting the element destroyed the editor too, as did a one-frame remount. A reattach after a real teardown came back inert.

Extension sets now compare by value. Teardown defers by a task. The `@carbon/ai-chat` half of #2152 lands in #2185, stacked on this one, so this PR alone does not close the issue.

- `src/tiptap/extension-equivalence.ts` — new. Built Tiptap instances cannot be compared directly. Each factory tags its output with the config it came from.
- `src/prompt-line-rich-runtime.ts` — the value comparison, the recreate withheld during IME composition, and live placeholder updates.
- `src/prompt-line.ts` — deferred teardown, the re-runnable mount path, and pending `ensureEditor()` callers surviving a remount.

#### Changelog

**New**

- Extension sets compare by value. An equivalent rebuild no-ops.
- Teardown defers by a task, so reparenting keeps the editor and its undo history.
- An element reattached after teardown mounts a working surface again.

**Changed**

- A recreate during IME composition waits for `compositionend`, then re-tests against the installed set. The rebuild runs a macrotask later, so ProseMirror commits the last composed segment first, and is skipped if the host detached in the meantime.
- Placeholder changes apply to the live editor. They no longer ride on a recreate.
- Keyboard-focus state survives a genuine recreate.

**Removed**

- `ValueSync`'s unread `lastTransactionIsHost` flag and `TypingIndicator.storage.reset()`.
- Barrel exports the package barrel never re-exported.

#### Testing / Reviewing

**Read it commit by commit.** Seven fix commits and a docs correction, then two behavior-neutral cleanups. Those two peel off with `reset --hard HEAD~2` if they stall the review.

**Gate** — `npm run build --workspace=@carbon/ai-chat-components`, then `npm run test --workspace=@carbon/ai-chat-components`. Build exits 0. Chromium and Firefox: 521 passed, 0 failed, 2 skipped. Plus 19 Jest tests for the React wrappers. `lint` and `lint:license` exit 0.

**WebKit is flaky here, and it pre-dates this branch.** WebKit's page intermittently dies on `workspace-shell.test.ts`, which this diff does not touch, dropping ~18 tests with zero assertion failures. It reproduces on the branch point with none of these commits applied, and the same commit gives green and red runs back to back, so it tracks machine load rather than code. That file passes 28/28 alone, and all 164 prompt-line tests pass on WebKit alone. Worth confirming against CI.

**No demo path here.** The churn that triggers the bug starts in `@carbon/ai-chat`. The issue's `prompt-line-conversation-starters` repro only goes green once #2185 lands.

The proof in this PR is `src/components/prompt-line/src/__tests__/prompt-line-editor-stability.test.ts`. It covers the issue's last four criteria. A same-frame remount keeps the editor. A pending `ensureEditor()` resolves across it. A reattach after teardown works. A genuine unmount still destroys.

**On the cleanup commits.** Every removed line pre-dates this branch. None had a runtime reader outside its own specs. The barrel names were never in a stable release. They were reachable only through undocumented `./es/*` deep imports in the two `1.9.0-rc` prereleases.

One exception, and it only bites in combination. `UndoRedo`'s barrel export was leaking the `@tiptap/extensions` module augmentation into `@carbon/ai-chat`'s TypeScript program. That augmentation declares `undo` and `redo` on Tiptap's `Commands`. Nothing on `main` reads it, so this PR builds green alone. #2185's spec calls `commands.undo()`, so it declares `@tiptap/extensions` itself.

**One untested path.** `clear stale IME state and allow rich-load retry` drops a memoized rejected `import()` so the retry can run. No test pins it. A seam exists — a Playwright launcher aborting the runtime chunk, or a settable importer — so treat this as a gap, not a limit.

**How the IME fix is proven.** Synthetic composition events never reach ProseMirror's own handlers, so no unit test can drive a real IME here. The new case dispatches into the outgoing editor on a microtask instead, standing in for PM's post-`compositionend` flush. Under the old synchronous rebuild that editor is already destroyed and the segment is lost, so the case does fail without the fix. The flush timing itself comes from reading `prosemirror-view`, not from a test.


ethanwinters:
Consolidated into #2185 — the same commits ship there as one PR.

Issue #1804, opened 2026-07-17
[Epic]: 2.0.0 Breaking public-API renames & reshape


ethanwinters:
### Value proposition

Public naming and API shape have drifted from the product — "prompt line", one canonical component, `upsertMessage`. 2.0 is the window to fix that. Grouping the breaking renames, reshapes, and removals lets us document and migrate them together, in one release.

### Expected outcomes

- [ ] Public config and type names match the product surface: `PromptLineConfig`, `Chat` / `cds-aichat`, all-caps `ID`.
- [ ] Messaging standardizes on `upsertMessage`. The legacy paths are deprecated, then removed.
- [ ] The writeable-element surface is final: one vocabulary, no overlapping slots.
- [ ] Event handlers can cancel and mutate events. Every message that goes out fires `pre:send` and `send`.
- [ ] The framework injects no messages of its own. The implementor owns message-list content.
- [ ] Every change here is written up in the shared 2.0 migration doc (#1539).

### Out of scope

- Non-breaking additive API — its own feature epic, or the post-2.0.0 bucket #1808.
- Theming and config tokens — the theming epic #1800 owns `customProperties` and theme.
- Response-type *rendering*. #1803 keeps that on 1.x; only the consumer-breaking half lives here, under #1913.

### Details

Parent epic: #1486. The sub-issue list on this epic is the child record, so it is not restated here.

Children that share a surface get a sub-epic. The rest stay flat.

| Sub-epic | Surface |
| --- | --- |
| #1529 | Items already deprecated on the 1.x line |
| #1889 | Framework auto-injected messages |
| #1913 | Response types |
| #1929 | The custom element as the one component; host owns launcher, float, and phone layout |
| #1930 | The writeable-element enum |
| #1931 | Event bus and send pipeline |

PR #2185, opened 2026-08-18
fix(prompt-line): keep the editor across config updates

### LINKED ISSUE #2152: [Bug]: Runtime config updates recreate the prompt-line editor and truncate undo history

ethanwinters:
Closes #2152

A runtime config update killed the live prompt-line editor and cut undo history short. `mergePublicConfig` deep-clones the host config on every update. Value-equal sub-objects come back with new identities. `setExtensions` then recreated the editor on any reference change. The user lost their undo stack mid-typing. DOM moves hit the same teardown. Reparenting the element destroyed the editor. A reattach after a real teardown came back inert.

The fix works at two layers. Extension sets now compare by value, so identity churn from above cannot recreate the editor. And `reconcileAppConfigReferences` recurses a second level into `public.input`. Value-equal siblings keep their previous references when one input field changes.

Complex files, worth reading closely:

- `prompt-line-rich-runtime.ts` — the rebuild/defer/absorb choice spans three methods. The two equivalence checks compare the requested set against a different partner on purpose: last-supplied first, then *installed*.
- `prompt-line.ts` — teardown defers a macrotask and cancels on reconnect. `_initializeSurface` replaces the one-shot `firstUpdated` mount.
- `extension-equivalence.ts` — new. Positional compare. Same instance wins. Starter triggers always compare equal, because their differences ride on live storage.

**The deep clone stays.** An earlier revision of #2152 prescribed dropping it, so the store would hold host references. That was built and then pulled back, because it quietly breaks hosts that change a config object they already handed over — allowed since 1.x. The editor fix does not need it, and holding by reference moves to 2.0 under #1804. The issue records this in its own "The deep clone stays" section; nothing here diverges from it.

#### How the twelve commits build up

Read oldest first. The element work is a chain, each link resting on the one before; the config work and the two refactors are independent strands that land where their proofs do. **Bold is load-bearing** — drop it and you break the fix, a proof the issue names, or the build. Proof file in brackets.

**1. Keep the editor when an equivalent set is rebuilt.** The core, and the commit the rest of the element work stands on. Each curated factory tags its output with the config that built it, and `setExtensions` compares those tags instead of array identity. An equivalent set now no-ops.

That flips a rule the rest of the element quietly leaned on. Settings used to reach the editor by riding the recreate. With recreates gone, each one has to be written through to the live instance, so this commit pays that debt for starter `items`/`isOn` and for the placeholder. It also fixes two defects the old recreate hid: a dropped content update, and a mis-read focus class. [`prompt-line-editor-stability.test.ts`, `extension-equivalence.test.ts`]

**2. Survive a detach, recover from a real teardown.** Commit 1 closed the config route to a destroyed editor. A DOM move still destroyed it by a different route, so the bug was only half fixed. Teardown now defers a task and cancels on reconnect.

Getting there means splitting the one-shot `firstUpdated` mount into `_initializeSurface` and `_teardownSurface`. Until that landed, an element reattached after a real teardown stayed inert for good. Re-running the mount also means tracking what the surface was seeded with, so the mount echo is suppressed once. The seed is a value plus a latch, and the latch is what makes it one-shot — matching on the value alone would suppress every later update back to it, so a host going `a` → `b` → `a` would silently keep `b`. [`prompt-line-editor-stability.test.ts`]

**3. Hold an extension change until the IME commits.** This one needs both of the first two. It only matters because commit 1 narrowed recreates down to genuine changes, and its bail on a detached host is only correct because commit 2 made teardown deferred.

Three problems, in order. A recreate during composition strands the in-flight candidate, so it waits. At the flush it must be re-tested against the *installed* set, or a host that changed the config and changed it back rebuilds for nothing. And the flush waits a macrotask, because ProseMirror commits the last composed segment a microtask after `compositionend`. Before this, a CJK user lost the character they had just typed. [`prompt-line-editor-stability.test.ts`]

4. **Correct claims the branch outdated.** Comments, JSDoc and test names, plus one assertion. It has to come after 1–3 because it strengthens the reattach test commit 2 wrote — that test asserted `view.dom.isConnected`, true of any mounted editor. It also renames the describe and four mechanism-named cases, and drops a duplicated local constant for the `PM_KEYBOARD_FOCUS_CLASS` commit 3 exports. The rest fixes comments that were wrong before this branch started. [`prompt-line-editor-stability.test.ts`]

**5. Reconcile input references a second level down.** The second layer, and the first commit in `@carbon/ai-chat`. Commits 1–4 make the element survive churn. This stops the churn at its source: `reconcileAppConfigReferences` now recurses into `public.input`, so a change to one input field leaves its value-equal siblings on their old references.

Either layer alone fixes the issue's repro. Together they cover different halves — commit 1 handles a memo that genuinely recomputes, this handles a memo that never should have. [`configReferenceSharing_spec.ts`, `mergePublicConfig_spec.ts`]

**6. Cover editor survival across a runtime config update.** The issue's regression box. It drives the whole path in one test — a React prop change through `mergePublicConfig`, `createAppConfig`, the reconciliation, the input hooks, into the live editor — which is why it sits after the element work rather than beside it.

All three of its cases pin layer 1, the value comparison. Each first shipped green against a build where the line it existed to pin had been reverted, and each fails without that line now. [`promptLineEditorStability_spec.tsx`]

7. Describe the new extension-comparison contract. Consumer docs for the rule commit 1 introduced. Curated configs compare by value. Their callbacks still compare by reference.

**8. Stop mutating the host's layout custom properties.** A separate defect in the same family, found while working on commit 5. `validateCustomProperties` deleted rejected Carbon tokens from the map it was handed. That map is `ChatAppEntry`'s change-detection baseline, so a value-identical update compared as changed and dispatched a config replace on every host render. Copy before validating.

It rides on this branch because its proof lives in `mergePublicConfig_spec.ts`, the file commit 5 creates. It is not needed for #2152's repro. [`styleUtils_spec.ts`, `mergePublicConfig_spec.ts`]

**9. Recompute starters when `disableDirectSend` changes.** Found by auditing the memo commit 5 edits, and still latent after it. `mergePublicConfig` clones `starters.items` on every update, so the memo recomputes on `items` identity whatever else is listed, and the missing dependency stays invisible — commit 5 does not change that, because a `disableDirectSend` flip makes the whole `starters` object unequal and the fresh clone is kept. What turns it live is anything that stops the churn: 2.0 holding config by reference, or a host supplying value-matched items. Then flipping only `disableDirectSend` would keep sending starters directly.

Only a hook-level spec can pin this, for the same reason. It pins the other direction too: satisfying the exhaustive-deps lint with `[starters]` would reintroduce #2152 with every test still green. [`useInputExtensions_spec.tsx`]

10. Move caret mapping next to the doc it walks. Refactor. No behavior change. `prompt-line.ts` carries no ProseMirror position math now. [`json-utils.test.ts`]

11. Write starter storage through one helper. Refactor, and last on purpose — it rewrites code commit 1 and the autocomplete controller both own. Commit 1 added a third writer to starter storage, and only one of the four copies guarded `items`, so which writer won depended on call order. One helper now. No new assertions — it ships under the existing suite, whose helper it rewires.

**12. Keep the editor when the starters list empties.** The last hole in commit 1's contract, found reviewing the branch. `buildCarbonExtensions` installed the starter trigger only when `items` was non-empty, so emptying the list dropped it from the built set — a length change the equivalence check has to reject, recreating the editor and resetting undo history. A host clearing its starters after the first message walked straight into it, and the contract commit 7 documents gave no hint.

The extension is installed whenever `starters` is configured now, so emptying the list is a storage write like any other swap. Emission stays gated on a non-empty list, so hosts that were never installing it see no new events. `resolvePromptLineMode` is untouched and still gates rich mode on a non-empty list — it asks whether Tiptap is needed at all, where the built set asks what is installed once it is. [`prompt-line-editor-stability.test.ts`, `extension-equivalence.test.ts`, `carbon-starter-trigger.test.ts`]

#### Changelog

**New**

- Extension sets compare by value. An equivalent rebuild keeps the editor and its undo history.
- Teardown defers by a task, so a same-frame detach keeps the editor.
- An element reattached after teardown mounts a working surface again.

**Changed**

- `reconcileAppConfigReferences` reconciles `public.input` a second level down.
- `validateCustomProperties` copies before validating. It no longer deletes bad Carbon tokens from the map it was handed.
- A `content` update that returns to the value the surface was seeded with now applies. The seed is skipped once, not for the life of the surface.
- A recreate during IME composition waits for `compositionend`. It re-tests against the installed set, a macrotask later.
- Placeholder changes apply to the live editor.
- Keyboard-focus state survives a recreate.
- The starters memo now depends on `isOn` and `disableDirectSend`. Neither was ever listed; the clone's churn hid both.
- Emptying `starters.items` keeps the editor. The starter trigger stays installed for an empty list, so the change is a storage write rather than a set-length change that would recreate the editor.
- `getEditor` guidance, the `tiptap.extensions` JSDoc and `PromptLine.md` now describe comparison by value. Host extensions and curated-config callbacks still compare by reference.

#### Testing / Reviewing

**The repro from the issue.** Start `examples/react/prompt-line-conversation-starters`. Focus the editor. Type `hello world`. Then undo. The field returns to empty in one step. Before, it stopped at `h`, and further undos did nothing.

Verified by hand on this branch. The editor is the same instance after typing. The reference captured before typing still works — it used to throw "the editor view is not available".

`prompt-line-typeahead` has a static config and is unaffected. Same editor instance across typing. Undo returns to empty. The autocomplete overlay still filters: `carbon` narrows to its two matching suggestions. Static-config hosts do take changes here: deferred teardown, live `setPlaceholder`, the content-seed guard. So check it rather than assume it.


warrenmblood (approved):
One non-blocker to consider. These changes now preserve editor history even after message send, so users can (crtl+z) to draft history of previous messages. If we don't want this behavior, we could default to rebuilding the editor after each message send.

PR #2192, opened 2026-08-18
feat(prompt-line): remove name from suggestion configs

### LINKED ISSUE #2139: Remove TriggerSuggestionConfig.name and AutocompleteConfig.name
### Background

`name` overrides the Tiptap schema-node name on a suggestion config. It exists to sidestep Tiptap's last-named-wins stacking caveat: each trigger is built as `Mention.extend({ name })`, so two mention-style triggers both registering as `mention` would collide and one would silently replace the other. Giving each a distinct name lets them coexist.

Setting it produces a suggestion surface that never shows any items.

The factory uses the same string for two jobs — the schema node name and the emitted trigger type (`tiptap/carbon-mention.ts:55`, `tiptap/carbon-autocomplete.ts:29`):

```ts
const name = config.name ?? build.defaultName;
...
onStart: (props) => emitTrigger(props.editor, name, props.query, props.range, ...)
```

The controller then resolves the config by matching that type against four hardcoded literals (`autocomplete-controller.ts:476-494`):

```ts
const config =
  trigger.type === 'mention' ? this._mention
    : trigger.type === 'command' ? this._command
      : trigger.type === 'autocomplete' ? this._autocomplete
        : undefined;
if (!config) {
  return [];
}
```

A custom name matches none of them, so the resolver returns an empty array every time. The schema half works — the node really does take the custom name and really does avoid the collision — but the very name that solves the collision is what breaks the lookup.

### Decision: remove it

Supporting multiple same-kind triggers means reworking how the controller routes a trigger to its config, which is more than this release warrants. Remove the field for now and bring it back with the routing work if hosts ask for it.

Removal is safe and cheap here:

- The surrounding config types are `@experimental`, so no deprecation cycle is owed.
- No consumer can have a working dependency on it. Setting it yields a dead surface, so there is no behavior to preserve.
- Hosts that never set it are unaffected — the factories already fall back to the default name.

### Public API / contract

Remove the field from both configs that expose it:

```ts
// packages/ai-chat-components/src/components/prompt-line/src/tiptap/types.ts

export interface TriggerSuggestionConfig extends Omit<...> {
-  /**
-   * Override the schema-node name. Threaded through `Mention.extend({ name })`
-   * inside the carbon factory to sidestep Tiptap's last-named-wins stacking
-   * caveat when multiple triggers coexist.
-   */
-  name?: string;
}

export interface AutocompleteConfig extends BaseSuggestionConfig {
-  name?: string;
}
```

Both are re-exported from `packages/ai-chat/src/types/config/InputConfig.ts`, so the public surface loses one optional field on each.

## Goal

`name` is gone from the public config types, the factories always use their default node name, and the constraint it was meant to solve is documented instead.

## Acceptance criteria

- [ ] `name` is removed from `TriggerSuggestionConfig` (`types.ts:152`) and `AutocompleteConfig` (`types.ts:188`).
- [ ] `carbon-mention.ts:55` and `carbon-autocomplete.ts:29` use the default name unconditionally.
- [ ] Plugin keys stay unique across the built-in triggers.
- [ ] Behavior is unchanged for any host that never set `name`.
- [ ] The docs state the limit that remains: one mention trigger and one command trigger per chat, because Tiptap resolves extensions by name.
- [ ] The generated API docs and symbol index no longer list the field.

## Out of scope

- Reworking the controller so a trigger routes to its config by something other than a hardcoded node name. That is the work that would let `name` come back, and it belongs with multi-trigger support rather than this removal.

## Related

- `packages/ai-chat-components/src/components/prompt-line/src/tiptap/types.ts:152,188` — the two declarations.
- `packages/ai-chat-components/src/components/prompt-line/src/tiptap/carbon-mention.ts:55` — name resolution.
- `packages/ai-chat-components/src/components/prompt-line/src/tiptap/carbon-autocomplete.ts:29` — same, for autocomplete.
- `packages/ai-chat-components/src/components/prompt-line/src/autocomplete-controller.ts:476-494` — the literal-matching resolver.
- #2158 — renaming a node to a custom type was the only route to a custom chip in the sent bubble. Removing `name` closes it explicitly rather than leaving it broken.

AlexanderMelox:
Closes #2139

Removes the `name` field from suggestion configs. Setting it always broke the suggestion surface: the controller matches configs against four hardcoded type literals, and a custom name matches none of them.

#### Changelog

**Removed**

- `name` from `TriggerSuggestionConfig` and `AutocompleteConfig`. Both types are `@experimental`, so no deprecation cycle.

**Changed**

- `carbon-mention.ts` and `carbon-autocomplete.ts` always use the default schema-node name.
- Plugin keys stay unique across mention/command/autocomplete.
- Docs now state the limit this leaves: one mention trigger and one command trigger per chat.

#### Testing / Reviewing

No behavior change for hosts that never set `name` — same default names as before.

- `npm run aiChat:start`, then in `demo/`: `npm start`
- Enable `@mentions` and autocomplete from the sidebar, confirm both suggestion pickers still show items and insert tokens/text as before.


packages/ai-chat/docs/api/markdown/carbonMention.md
```diff
@@ -7,8 +7,8 @@
 Tiptap extension factory for `@`-style mention triggers. Wraps
 `@tiptap/extension-mention` with Carbon-specific chip rendering, extended
 schema attributes (`value`, `data`), and direct
-`cds-aichat-trigger-change` dispatch. Pass distinct `name` values when
-composing multiple instances.
+`cds-aichat-trigger-change` dispatch. Each chat supports one mention
+trigger, because Tiptap resolves extensions by name.
```
ethanwinters: I think we can leave out ", because Tiptap resolves extensions by name."

packages/ai-chat/docs/api/markdown/TriggerSuggestionConfig.md
```diff
@@ -6,13 +6,15 @@
...
-`showTriggerInChip` default (whether selected items render as
-`/summarize` or a bare `summarize`, overridable per item) on top of
+trigger character, an optional `triggerPosition`, a custom-token renderer,
+an `onRemove` callback (the mirror of `onSelect`, fired when a token is
+deleted), and a `showTriggerInChip` default (whether selected items render
+as `/summarize` or a bare `summarize`, overridable per item) on top of
 BaseSuggestionConfig.
 
+Each chat supports one mention trigger and one command trigger, because
+Tiptap resolves extensions by name.
```
ethanwinters: Each chat supports one mention trigger and one command trigger.

packages/ai-chat/docs/api/symbol-index.json
```diff
@@ -23372,7 +23372,7 @@
       "kind": "TypeAlias",
       "parent": null,
       "category": "Config",
-      "summary": "Trigger-character-driven suggestion config consumed by\nInputConfig.mention and InputConfig.command. Adds the\ntrigger character, an optional `triggerPosition`, an optional schema-node\n`name` override, a custom-token renderer, an `onRemove` callback (the\nmirror of `onSelect`, fired when a token is deleted), and a\n`showTriggerInChip` default (whether selected items render as\n`/summarize` or a bare `summarize`, overridable per item) on top of\nBaseSuggestionConfig.",
+      "summary": "Trigger-character-driven suggestion config consumed by\nInputConfig.mention and InputConfig.command. Adds the\ntrigger character, an optional `triggerPosition`, a custom-token renderer,\nan `onRemove` callback (the mirror of `onSelect`, fired when a token is\ndeleted), and a `showTriggerInChip` default (whether selected items render\nas `/summarize` or a bare `summarize`, overridable per item) on top of\nBaseSuggestionConfig.\n\nEach chat supports one mention trigger and one command trigger, because\nTiptap resolves extensions by name.",
```
ethanwinters: This shouldn't have been touched. This is only touched by our ci/cd on a release.

packages/ai-chat/docs/api/markdown/carbonMention.md
ethanwinters: Oh, noticing where this came from. This whole folder shouldn't be touched except by ci/cd.

packages/ai-chat/docs/api/markdown/TriggerSuggestionConfig.md
ethanwinters: Oh, noticing where this came from. This whole folder shouldn't be touched except by ci/cd.

packages/ai-chat/src/types/config/InputConfig.ts
```diff
@@ -39,13 +39,15 @@ export type BaseSuggestionConfig = _BaseSuggestionConfig;
...
- * `showTriggerInChip` default (whether selected items render as
- * `/summarize` or a bare `summarize`, overridable per item) on top of
+ * trigger character, an optional `triggerPosition`, a custom-token renderer,
+ * an `onRemove` callback (the mirror of `onSelect`, fired when a token is
+ * deleted), and a `showTriggerInChip` default (whether selected items render
+ * as `/summarize` or a bare `summarize`, overridable per item) on top of
  * {@link BaseSuggestionConfig}.
  *
+ * Each chat supports one mention trigger and one command trigger, because
+ * Tiptap resolves extensions by name.
```
ethanwinters: Ok here is where we can leave out the ", because
 * Tiptap resolves extensions by name"

packages/ai-chat/src/types/utilities/inputUtils.ts
```diff
@@ -49,8 +49,8 @@ import type {
  * Tiptap extension factory for `@`-style mention triggers. Wraps
  * `@tiptap/extension-mention` with Carbon-specific chip rendering, extended
  * schema attributes (`value`, `data`), and direct
- * `cds-aichat-trigger-change` dispatch. Pass distinct `name` values when
- * composing multiple instances.
+ * `cds-aichat-trigger-change` dispatch. Each chat supports one mention
+ * trigger, because Tiptap resolves extensions by name.
```
ethanwinters: And here

ethanwinters (commented):
We shouldn't run the job to update docs/api unless we cut a release so its not a mixture of a bunch of different versions.

packages/ai-chat-components/src/components/prompt-line/src/tiptap/types.ts
```diff
@@ -179,11 +170,11 @@ export interface TriggerSuggestionConfig extends Omit<
...
  * Live autocomplete config. Selection inserts plain text (no token chip).
+ *
+ * Note: only one autocomplete config per editor is supported, because Tiptap
+ * resolves extensions by name.
  */
-export interface AutocompleteConfig extends BaseSuggestionConfig {
-  /** Override the suggestion plugin key name. Defaults to `"autocomplete"`. */
-  name?: string;
-}
+export type AutocompleteConfig = BaseSuggestionConfig;
```
AlexanderMelox, line 177: ℹ️  Had to move this to a type since leaving it as interface was causing build errors

davidmenendez (approved):

PR #2204, opened 2026-08-19
fix(shared): defer dynamic sheet rewrites out of the teardown window

### LINKED ISSUE #2201: Investigate WebKit browser crash in ai-chat-components local test runs
## Background

`npm run test --workspace=@carbon/ai-chat-components` reproducibly crashes the WebKit browser on local macOS runs. The failure is a browser-process crash, not a test failure — WebKit reports `0 failed` right up until it disconnects:

```
Chromium: 35/35 test files | 489 passed, 0 failed, 2 skipped
Firefox:  35/35 test files | 489 passed, 0 failed, 2 skipped
Webkit:   35/35 test files | 471 passed, 0 failed, 2 skipped
 ❌ Tests were interrupted because the browser disconnected.  (failed on Webkit)
 ❌ Error: page.goto: Page crashed  (failed on Webkit)
```

Roughly 18 tests never report on WebKit while the other two browsers are clean on identical code. It does not correlate with a particular spec file — the crash surfaces as `page.goto: Page crashed` during navigation, after most files have already run.

The cost is not the suite itself but `--nx-bail`: when `@carbon/ai-chat-components:test` fails, nx cancels every remaining test target, so `@carbon/ai-chat:test`, `@carbon/ai-chat-examples-demo:test`, and the example suites are silently *not run*. A local `npm run ci-check` therefore cannot come back clean, and each failure has to be hand-triaged to confirm it is only WebKit.

Observed on macOS 26.5.2 / arm64 / Node v24.14.0, Playwright 1.62.1. CI (Linux) is unaffected — the same commit and lockfile pass the `test` job there.

[web-test-runner.config.js](../../packages/ai-chat-components/web-test-runner.config.js) already carries tuning for resource-pressure flakes in this suite — `concurrency: 3`, `testsFinishTimeout: 300000`, `browserStartTimeout: 60000`, and a 10s Mocha timeout — with comments attributing earlier timeout flakes to CPU starvation from simultaneous lazy `import()`s. This crash may be the same root cause presenting as an OOM/renderer kill on WebKit rather than a timeout.

## Goal

Local `npm run test --workspace=@carbon/ai-chat-components` completes without crashing WebKit, so that `npm run ci-check` is a usable local gate on macOS.

## Acceptance criteria

- [ ] Root cause identified — renderer OOM, concurrency starvation, a Playwright/WebKit version interaction, or a specific component under test.
- [ ] `npm run test --workspace=@carbon/ai-chat-components` passes locally on macOS/arm64 across three consecutive runs.
- [ ] WebKit reports the same test count as Chromium and Firefox (currently 471 vs 489).
- [ ] If the fix is configuration, the reasoning is captured as a comment in `web-test-runner.config.js` alongside the existing concurrency notes.
- [ ] If WebKit turns out to be unsupportable for local runs, that is documented and the browser is gated (e.g. env-guarded) rather than left as a silent recurring failure.

## Out of scope

- CI behaviour — the Linux `test` job passes and is not affected.
- The Chromium/Firefox suites, which are clean.
- Broader test-runner migration.

## Investigation notes

Starting points, roughly in order of cheapness:

- Drop `concurrency` to 1 and re-run WebKit alone (`--browsers webkit`) to test the CPU-starvation theory.
- Check for a renderer OOM in Console.app / `~/Library/Logs/DiagnosticReports` at crash time.
- Bisect Playwright versions — 1.62.1 is current; the crash was first noticed 2026-08-18.
- Identify which ~18 tests are missing from the WebKit run by diffing reported test names against Chromium's.

## Related

Surfaced while validating the lockfile-maintenance backport onto `release/v1.19.0` (#2178). Not caused by it — that commit's lockfile diff contains no Playwright or WebKit change, and the crash reproduces independently.

ethanwinters:
Closes #2201

WebKit's renderer crashed partway through the `@carbon/ai-chat-components` suite. It took ~18 tests with it and still reported `0 failed`, so `ci-check` could never come back clean locally.

The cause was ours, not the runner. Four components clear their rules from `disconnectedCallback`. That put `replaceSync` inside WebKit's custom-element reaction queue during `Element.remove()`, where it walks the shared sheet's style scopes and hits a shadow-root host that is not valid mid-removal. All 31 local crash reports were identical: `SIGSEGV` at `0x1c` in `Style::Invalidator::invalidateHostAndSlottedStyleIfNeeded`.

- `src/components/shared/dynamic-css-var-sheet.ts` — the clear helpers flush on a microtask now. `setVarsForSelector` stays synchronous on purpose: `history-panel-item` measures right after a write.

#### Changelog

**New**

- First tests for `dynamic-css-var-sheet`.

**Changed**

- `clearSelector` and `clearVarsForSelector` flush on a microtask, not inline.

#### Testing / Reviewing

No visual change — the fix is on the teardown path. Proof is the suite plus the absence of crash logs.

```bash
npm run test --workspace=@carbon/ai-chat-components
```

- WebKit reports 511 passed / 0 failed, matching Chromium and Firefox. Ran three times in a row on macOS/arm64.
- No new files in `~/Library/Logs/DiagnosticReports/com.apple.WebKit.WebContent.Development-*.ips`.

To see the old crash: point both clear helpers back at `rewrite()`, rebuild, run the suite. Needs all three browsers — WebKit alone never reproduces it.

Two notes for the reviewer:

- Whether real Safari crashes on this path is still unverified. The path is a chat component leaving the DOM, which is ordinary consumer behaviour. If it does, this is a user-facing crash fix, not test cleanup.
- Any `chat-button` failures on a first run are a stale `es/` build, not this change. Nx replays a cached build and the tests import from `es/`. Rebuild and they clear.

### DIFF
--- a/packages/ai-chat-components/src/components/shared/__tests__/dynamic-css-var-sheet.test.ts
+++ b/packages/ai-chat-components/src/components/shared/__tests__/dynamic-css-var-sheet.test.ts
@@ -1,0 +1,111 @@
+/**
+ * @license
+ *
+ * Copyright IBM Corp. 2026
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { expect } from '@open-wc/testing';
+
+import {
+  setVarsForSelector,
+  clearSelector,
+  clearVarsForSelector,
+} from '../dynamic-css-var-sheet.js';
+
+/**
+ * Reads every rule the shared sheet has published onto the document. The module
+ * adopts its singleton on `document`, so a selector is "live" as soon as it
+ * appears here.
+ */
+function documentCssText(): string {
+  return Array.from(document.adoptedStyleSheets ?? [])
+    .map((sheet) => {
+      try {
+        return Array.from(sheet.cssRules)
+          .map((rule) => rule.cssText)
+          .join('\n');
+      } catch {
+        return '';
+      }
+    })
+    .join('\n');
+}
+
+describe('dynamic-css-var-sheet', function () {
+  // Unique per test so a leaked rule can never satisfy a later assertion.
+  let selector: string;
+  let counter = 0;
+
+  beforeEach(() => {
+    selector = `.dcvs-test-${++counter}`;
+  });
+
+  afterEach(async () => {
+    clearSelector(selector);
+    await Promise.resolve();
+  });
+
+  it('publishes a rule synchronously', () => {
+    setVarsForSelector(selector, { color: 'rgb(255, 0, 0)' });
+    expect(documentCssText()).to.contain(selector);
+  });
+
+  // Regression guard for #2201. Components call the clear helpers from
+  // `disconnectedCallback`, which WebKit runs while draining the custom-element
+  // reaction queue inside `Element.remove()`. Mutating an adopted sheet in that
+  // window segfaults the WebKit renderer, so the write must land after it.
+  it('defers clearSelector past the synchronous teardown window', async () => {
+    setVarsForSelector(selector, { color: 'rgb(255, 0, 0)' });
+    expect(documentCssText()).to.contain(selector);
+
+    clearSelector(selector);
+    expect(
+      documentCssText(),
+      'sheet must not be mutated inside disconnectedCallback'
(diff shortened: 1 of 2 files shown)


warrenmblood (approved):
LGTM

PR #2222, opened 2026-08-20
fix: markdown breaks

### LINKED ISSUE #2220: breaks not working in markdown
### Background

Two bugs reported by a consumer upgrading from v1.5.0 to v1.8.0.

### Goal

Hard and soft line breaks render correctly as `<br>` inside the markdown component, and `customRenderers.table` (and `codeBlock`) work as expected when the markdown component is used inside any of the chat container elements.

### Acceptance criteria

- [ ] A markdown string containing `  \n` (two trailing spaces + newline) renders a visible `<br>` in the output. Verified by the new `hardbreak` WTR spec in `markdown.test.ts`.
- [ ] A single `\n` inside a paragraph renders as `<br>` (breaks mode). Verified by the new `softbreak` WTR spec.
- [ ] Neither token type produces a `pluginFallback` named slot. Verified by the "does not route through plugin-fallback" WTR spec.
- [ ] Returning an `HTMLElement` from `customRenderers.table` inside a `ChatContainer` / `cds-aichat-container` / `cds-aichat-custom-element` renders the custom element in place of the default `cds-aichat-table`. Verified by the new "slot host stays in markdown element light DOM" WTR specs.
- [ ] Definition-of-done gate: `npm run test:web-components --workspace=@carbon/ai-chat-components` exits 0; `npm run build` exits 0 for both affected packages.

### Root causes

**Bug 1 — line breaks swallowed**

`hardbreak` tokens have `tag: 'br'` and were reaching `renderWithStaticTag`'s `default:` branch, which routed them through `renderFallback` (the plugin-fallback slot system). `softbreak` tokens (`tag: ''`) hit the same path. When the markdown-it instance was not yet initialized (first render), both types were silently dropped instead. Fixed by adding explicit `hardbreak` / `softbreak` cases in `renderTokenTree` before the tag dispatch, and adding both types to `NATIVELY_HANDLED_TOKEN_TYPES`.

**Bug 2 — `customRenderers.table` ignored**

All three chat containers (`ChatContainer.tsx`, `cds-aichat-container`, `cds-aichat-custom-element`) intercepted the `cds-aichat-markdown-plugin-host-mount` event for all descriptor kinds and relocated the returned `HTMLElement` host into the outer wrapper. A named `<slot>` only projects light-DOM children of its own host element, so moving the host to an ancestor broke the slot projection and the default `cds-aichat-table` rendered instead. Fixed by returning early (without `preventDefault()`) when `detail.element` is present, leaving those hosts in the markdown element's own light DOM where the slot projection can reach them.

### Out of scope

- Spacing adjustments for custom-renderer slot hosts (shadow CSS `> slot { display: block }` and `> *:not(:first-child)` already cover it).
- React wrapper (`Markdown` component) — it does not intercept `detail.element` hosts and was unaffected.

davidmenendez:
Closes #2220 

Here's a summary of every change made and why:

---

### Bug 1 — `\n` / hard line breaks swallowed

**[`markdown-renderer.ts`](packages/ai-chat-components/src/components/markdown/src/markdown-renderer.ts:262)** — Added explicit early returns for `hardbreak` and `softbreak` tokens before the `tag`-based dispatch:

```ts
if (token.type === 'hardbreak') { return html`<br />`; }
if (token.type === 'softbreak') { return html`<br />`; }
```

Both token types were previously falling through to `renderWithStaticTag`'s `default:` branch, which routed them through `renderFallback` — spinning up the full plugin-fallback slot machinery (emitting a named `<slot>`, dispatching mount events, creating light-DOM host elements) for what is just a `<br>`. When `markdownItInstance` was not yet set (first render), the fallback was skipped entirely and the break was silently dropped.

**[`plugin-fallback.ts`](packages/ai-chat-components/src/components/markdown/src/utils/plugin-fallback.ts:38)** — Added `'softbreak'` and `'hardbreak'` to `NATIVELY_HANDLED_TOKEN_TYPES` as a defence-in-depth guard so neither token can leak to `renderFallback` from any path.

---

### Bug 2 — `customRenderers.table` no longer applied

**[`ChatContainer.tsx`](packages/ai-chat/src/react/ChatContainer.tsx:241)**, **[`cds-aichat-container/index.ts`](packages/ai-chat/src/web-components/cds-aichat-container/index.ts:458)**, **[`cds-aichat-custom-element/index.ts`](packages/ai-chat/src/web-components/cds-aichat-custom-element/index.ts:253)**

All three containers' `handlePluginHostMount` handlers were calling `event.preventDefault()` unconditionally and then moving `detail.element` hosts (from `customRenderers.table`/`codeBlock`) into the outer wrapper. This pulled those hosts out of the markdown element's own light DOM, breaking the `<slot name="...">` projection — a named `<slot>` only projects children of its host element, not nodes from ancestor trees.

The fix: when `detail.element` is present (consumer-renderer path), return early without calling `preventDefault()`. The markdown element's own `reconcileCustomRendererHosts` `appendChild` path handles these hosts correctly. Only HTML-string `pluginFallback` hosts — which genuinely need relocation into page light DOM for consumer stylesheets (e.g. KaTeX) to reach them — go through the container's hoisting logic.

#### Testing / Reviewing

Here's how to test both fixes locally.

## Build first

The workspace resolves from built artifacts, so rebuild after the changes:

```bash
```

## Run the new automated tests

```bash
npm run test:web-components --workspace=@carbon/ai-chat-components
```

This runs the full WTR suite including the five new specs. Look for:

```
Finished running tests in ~75s, all tests passed! 🎉
```

## Verify Bug 1 (line breaks) visually

The quickest path is the existing Markdown storybook. Start it:

```bash
npm run storybook --workspace=@carbon/ai-chat-components
```

Then open [http://localhost:6006](http://localhost:6006), navigate to **Messaging → Markdown**, and paste this into the `markdown` control:

```
**Summary**  
Use the **Assets** application to add selected assets.

Second paragraph
with a soft break here
and another line.
```

**Before the fix:** the hard break (`  \n`) and soft breaks collapse — lines merge into one.  
**After the fix:** each `  \n` renders a visible line break (`<br>`), and single newlines also produce a `<br>` because `breaks: true` is active.

## Verify Bug 2 (`customRenderers.table`) visually

The `markdown-override` React example is purpose-built for this:

```bash
PORT=3001 npm start --workspace=@carbon/ai-chat-examples-react-markdown-override
```

Then open [http://localhost:3001](http://localhost:3001). Send a message that returns a markdown table (the example ships with one). Your custom table renderer should render instead of the default `cds-aichat-table`.

If you want to test against the exact markdown from the bug report, open `examples/react/markdown-override/src/customSendMessage.ts` and replace (or add) the response body with:

```
**Summary**  \nUse the **Assets** application to add selected assets to an existing active collection.\n\n**Result**\n\n| Outcome | Details |\n|---|---|\n| Membership | Selected assets are added to the specified collection |\n| Scope | Collections can contain assets, locations, and classification items |\n| Management | All members can be managed together as a group |
```

**Before the fix:** the table renders as the default `cds-aichat-table`; the custom renderer callback is ignored.  
**After the fix:** your custom renderer's output appears in place of the default table.

## Quick DOM check (no example app needed)

Open the browser console on any page running a `<cds-aichat-markdown>` element and run:

```js
// Should find a <br> for a hardbreak
const md = document.querySelector('cds-aichat-markdown');
md.markdown = '**line one**  \nline two';
await md.updateComplete;
console.assert(!!md.shadowRoot.querySelector('br'), 'hardbreak: br expected');

// customRenderers.table — host should be a direct child of the element
md.customRenderers = {
  table: () => {
    const div = document.createElement('div');
    div.id = 'my-custom-table';
    return div;
  }
};
md.markdown = '| h1 | h2 |\n|---|---|\n| a | b |\n\nTrailer';
await md.updateComplete;
const host = md.querySelector('[slot*="cds-aichat-markdown-renderer-table"]');
console.assert(host?.parentElement === md, 'table host: parent should be the markdown element');
console.assert(!!host?.querySelector('#my-custom-table'), 'table host: custom element should be inside');
```

All three assertions should pass silently.


packages/ai-chat/src/react/ChatContainer.tsx
```diff
@@ -238,22 +238,20 @@ function ChatContainer(
...
-        if (element.parentElement !== wrapper) {
-          wrapper.appendChild(element);
-        }
         return;
       }
+      // Plugin-fallback hosts carry an HTML string. Hoist them to page light
+      // DOM so consumer-loaded global stylesheets (e.g. KaTeX) can reach the
+      // rendered output — the markdown element's own light DOM sits inside this
+      // wrapper's shadow root, where global CSS does not apply.
+      event.preventDefault();
```
ethanwinters: **Blocker** — hoisting is disabled for every custom-renderer host to fix a case that only affects consumer-owned markdown elements, so the `config.markdown.customRenderers` path loses page-global CSS: a `@carbon/react` DataTable in `examples/react/markdown-override` goes from `th` padding `0px 16px` → `0px` and table width `608px` → `188.2px`. It also strands `examples/react/markdown-override/README.md:8`, which documents the behavior being removed. Narrow the guard to the case that actually breaks — reproduction and three options in the review summary.

packages/ai-chat/src/web-components/cds-aichat-container/index.ts
```diff
@@ -458,34 +458,28 @@ class ChatContainer extends FlattenedConfigElement {
...
-      // This applies to the live-element path too: appending here would land
-      // the node in this element's light DOM (inside the outer chat element's
-      // shadow root), where global CSS still wouldn't reach it.
       return;
     }
+    // Plugin-fallback hosts carry an HTML string. Hoist them to this element's
+    // outer light DOM so consumer-loaded global stylesheets (e.g. KaTeX) can
+    // reach the rendered output — the markdown element's own light DOM sits
+    // inside this element's shadow root, where global CSS does not apply.
     event.preventDefault();
```
ethanwinters: **Blocker** — same defect as in `ChatContainer.tsx`; whichever fix is chosen must land in all three containers together. See that comment for the evidence and the options.

packages/ai-chat/src/web-components/cds-aichat-custom-element/index.ts
```diff
@@ -250,28 +250,21 @@ class ChatCustomElement extends FlattenedConfigElement {
...
+      return;
+    }
     if (!this._pluginSlotNames.includes(detail.slotName)) {
       this._pluginSlotNames = [...this._pluginSlotNames, detail.slotName];
     }
     // cds-aichat-custom-element is always the outermost chat element in its
     // shadow chain; there is no further chat ancestor to defer to. Take
-    // hosting unconditionally.
+    // hosting unconditionally for plugin-fallback HTML-string hosts.
     event.preventDefault();
```
ethanwinters: **Blocker** — same defect as in `ChatContainer.tsx`; whichever fix is chosen must land in all three containers together. See that comment for the evidence and the options.

packages/ai-chat-components/src/components/markdown/__tests__/markdown.test.ts
```diff
@@ -2374,3 +2374,107 @@ describe('streaming table loading mode', () => {
...
+    // The slot's assigned nodes should include our host wrapper.
+    const assigned = (slotEl as HTMLSlotElement)?.assignedNodes({
+      flatten: true,
+    });
+    expect(
+      assigned?.length,
+      'slot should have at least one assigned node'
+    ).to.be.greaterThan(0);
+  });
+});
```
ethanwinters, line 2569: **Blocker** — both specs pass on unmodified `main`, so they don't cover the container change: `main` plus these specs gives 85 passed / 3 failed, and all three failures are line-break tests. The fixtures mount a bare `<cds-aichat-markdown>` with no container ancestor, so nothing intercepts the mount event — the entire mechanism under test. Rebuild them on `createRelocationHarness` (`markdown.test.ts:1503`), the only harness here that mirrors container behavior.

packages/ai-chat-components/src/components/markdown/src/markdown-renderer.ts
```diff
@@ -259,6 +259,21 @@ export function renderTokenTree(
...
+  if (token.type === 'hardbreak') {
+    return html`<br />`;
+  }
+
+  // Soft line break (single newline inside a paragraph). With `breaks: true`
+  // (which this component uses) markdown-it renders softbreak as <br>; match
+  // that to avoid swallowing the token or routing it through renderFallback.
+  if (token.type === 'softbreak') {
+    return html`<br />`;
+  }
```
ethanwinters: **Blocker** — these early returns sit above every delegation check, and break tokens are absent from `PLUGIN_DELEGABLE_TOKEN_TYPES` (`markdown-token-tree.ts:74-79`), so a plugin's `md.set({ breaks: false })` or `md.renderer.rules.softbreak` override — both honored on `main` via `renderFallback` — is now silently ignored. `markdownItPlugins` is this component's only public extension point, so there is no way back. Apply this, and add `'softbreak'` / `'hardbreak'` to `PLUGIN_DELEGABLE_TOKEN_TYPES` (both are leaf tokens, matching that set's invariant).

```suggestion
  if (token.type === 'hardbreak' || token.type === 'softbreak') {
    if (shouldDelegateToPluginRule(token, options.md) && options.md) {
      return renderFallback(token as Token, node, options.md, sanitize, options);
    }
    return token.type === 'softbreak' && options.md?.options.breaks === false
      ? html`${'\n'}`
      : html`<br />`;
  }
```

ethanwinters (commented):
Fix blockers before merge — the container change solves a real bug, but too broadly.

Bug 1 (line breaks) is a genuine defect, well fixed; one Blocker on it, because the hardcoded `<br />` closes the only plugin hatch consumers have. Bug 2 reproduces on `main` only when the consumer owns the markdown element, so disabling the hoist for every host fixes one path and regresses the other. Three Blockers total, one Important, plus the unowned trade-off below.

## The Bug 2 reproduction

It reproduces on unmodified `main` when a consumer renders markdown themselves inside chat light DOM (`renderUserDefinedResponse` and friends). The container hoists the host away and nothing in that subtree projects it back:

- Raw `<cds-aichat-markdown>` → `assignedNodes()` is `[]`, `flatten` yields the slot's own fallback, so the default `cds-aichat-table` renders and the host is orphaned at `cds-aichat-react < div < div < body`.
- Via the components React wrapper → `assignedNodes()` is `[slot]`, but that forwarder sits in the document tree, not a shadow tree, so it projects nothing and the region is blank.

Applying this diff flips `visibleCustomTables` 0 → 2 and `visibleDefaultTables` 1 → 0, reversibly. Since the report cites the `@carbon/ai-chat-components` Markdown component directly, that is very likely the reported case.

Through `config.markdown.customRenderers` it does **not** reproduce: seven configurations pass on `main`, including the nested `cds-aichat-custom-element` → `cds-aichat-container` chain, streaming, KaTeX alongside table overrides, 8-message history, and a full scroll sweep.

The stated root cause — "a named `<slot>` only projects light-DOM children of its own host element" — is not the mechanism. Hoisted `pluginFallback` / KaTeX hosts project correctly through the identical forwarder chain (`cds-aichat-container/index.ts:700-702` → `markdown.tsx:328`). The real cause is narrower: consumer-owned markdown elements have no forwarder in their subtree.

## Narrowing the fix

Hoisting buys global-CSS reach; not hoisting buys projection for consumer-owned elements. The two conflict, so pick deliberately. Best first:

1. **Make the markdown element self-heal.** Keep hoisting, then verify on the next frame that the shadow slot has assigned nodes; if not, re-adopt the host locally. Satisfies both cases with no consumer-visible contract change. Lives in `reconcileCustomRendererHosts` (`markdown.ts:902-920`), untouched by this PR.
2. **Discriminate at the container.** Take over hosting only when this container will render a forwarder that reaches the emitting element.
3. **Keep this change and own the trade-off** — update `markdown-override/README.md:8` and the `customRenderers` docs to say the returned element renders inside the chat's shadow root where global CSS will not reach it, and say how consumers should style it.

## Blocker — stale forwarder suppresses slot fallback

`packages/ai-chat-components/src/react/markdown.tsx:169-178` (outside this diff, so no line comment). With no `preventDefault()`, table/codeBlock slots never enter `delegatedPluginSlots` (`markdown.ts:916`), so the unmount dispatch (`markdown.ts:936-947`) never fires and the wrapper never prunes the name. The dead `<slot name=X slot=X>` counts as an assigned node and suppresses the slot's fallback (`table-helpers.ts:280`), so a renderer returning an element and later `null` leaves blank space where the default Carbon table belongs. Fix: add the same `detail.element` guard the containers now use.

## Important — the PR description's mechanism for Bug 1 is wrong

`softbreak`'s tag is `'br'`, not `''` (markdown-it pushes `state.push('softbreak', 'br', 0)`), so the `!tag` branches it cites are unreachable; and `markdownItInstance` is assigned before `renderAndDispatch` in the same block (`markdown.ts:598-615`, `:664`), so the "not yet set → break silently dropped" path cannot occur. The real defect: every break minted a `pluginFallback` descriptor, a host-mount event, and a light-DOM `<span slot><br></span>` on every render.

## Context worth carrying

The pre-#2119 slot-name collision produced the same user-visible symptom on components 1.8.0 (reproduced at v1.18.0 — message 1 `assignedCount: 2`, message 2 `assignedCount: 0`) and was fixed in 1.9.0-rc.0. A consumer on 1.8.0 may be hitting either cause.

## Test / verification gaps

- All three `handlePluginHostMount` handlers have zero coverage before and after (`ChatContainer.tsx` lines 231-239, 247-248, 254, 259-267 uncovered; neither web-component container appears in the Jest coverage report). That is why the original regression shipped.
- Nothing covers a `markdownItPlugins` entry changing break rendering.
- Untested, code reading only: `pluginFallback` hosts still hoist after this PR, so plugin/KaTeX output in the same consumer-owned configuration should still be broken.
- Pre-existing, worth its own issue: breaks inside a matched inline-HTML run are dropped — `<span>a  \\nb</span>` renders `ab`, because `utils/html-helpers.ts:348-361` `serializeInlineToken` returns `token.content ?? ''` and both break tokens carry `content: ''`.

Gates: `build` (components) 0 · `test:web-components` 554 passed / 0 failed on Chromium + Firefox + WebKit · `test:react` 19 · `test` (ai-chat) 1204 · lint / format / license / styles all 0. `build --workspace=@carbon/ai-chat` exits 1 on `typedoc` (`brace-expansion`/minimatch ESM) — environmental, not this diff.

1 Nit not listed (comment wording in the new specs).

davidmenendez:
@ethanwinters 

**All four reviewer Blockers are addressed. All five acceptance criteria are met.**

| | Status | Evidence |
|---|---|---|
| **Blocker A** — container hoisting regresses `config.markdown.customRenderers` CSS | ✅ PASS | Already resolved in `origin/main` (`325cb1e1`); containers are zero-diff in this PR |
| **Blocker B** — `customRenderers.table` specs pass on unmodified `main` | ✅ PASS | Both tests are now inside `createRelocationHarness`; they exercise the full container-intercept path |
| **Blocker C** — plugin delegation guard missing; `softbreak`/`hardbreak` absent from `PLUGIN_DELEGABLE_TOKEN_TYPES` | ✅ PASS | `shouldDelegateToPluginRule` guard in [`markdown-renderer.ts`](packages/ai-chat-components/src/components/markdown/src/markdown-renderer.ts:267); both types added to [`PLUGIN_DELEGABLE_TOKEN_TYPES`](packages/ai-chat-components/src/components/markdown/src/markdown-token-tree.ts:79) |
| **Blocker D** — stale `<slot>` forwarder suppresses fallback when renderer returns `null` | ✅ PASS | [`markdown.tsx`](packages/ai-chat-components/src/react/markdown.tsx:180) skips `detail.element` events in `handleMount`; second harness test asserts `cds-aichat-table` restores correctly |
| **AC 1** — hardbreak → `<br>` with WTR spec | ✅ PASS | [`markdown.test.ts:2470`](packages/ai-chat-components/src/components/markdown/__tests__/markdown.test.ts:2470) |
| **AC 2** — softbreak → `<br>` in breaks mode with WTR spec | ✅ PASS | [`markdown.test.ts:2482`](packages/ai-chat-components/src/components/markdown/__tests__/markdown.test.ts:2482) |
| **AC 3** — neither token produces `pluginFallback` slot with WTR spec | ✅ PASS | [`markdown.test.ts:2497`](packages/ai-chat-components/src/components/markdown/__tests__/markdown.test.ts:2497) |
| **AC 4** — `customRenderers.table` renders custom element inside container | ✅ PASS | First harness test: host hoisted, slot projects assigned element; second: `cds-aichat-table` restores on null |
| **AC 5** — `test:web-components` exits 0 | ✅ PASS | **555 passed, 0 failed** across Chromium, Firefox, WebKit |

packages/ai-chat-components/src/react/markdown.tsx
```diff
@@ -167,13 +167,21 @@ const Markdown = forwardRef<CDSAIChatMarkdown, MarkdownProps>(function Markdown(
...
+      if (!detail?.slotName) {
+        return;
+      }
+      // Consumer-renderer hosts (table/codeBlock) forward a live element that
+      // is adopted directly by the markdown element's own reconcile. They do
+      // not need a <slot> forwarder here — adding one would suppress the
+      // slot's fallback content if the callback later returns null.
+      if (detail.element) {
         return;
       }
```
ethanwinters, line 182: **Blocker** — this half is correct but incomplete alone: it skips the `<slot name=X slot=X>` forwarder for custom-renderer hosts while the containers still call `preventDefault()` and hoist those hosts into their own light DOM (`ChatContainer.tsx:241-254` on `main`). A named slot only projects children of its own host element, so the host is orphaned and the shadow slot falls back to the default Carbon table — the symptom #2220 Bug 2 reports, on every `customRenderers.table` / `codeBlock` returning an `HTMLElement` under `ChatContainer` (the chat's React path — `MarkdownWithDefaults` imports this wrapper). The container commits the PR body describes are what complete it; in `ChatContainer.tsx` that means moving the `detail.element` check above `preventDefault()` and dropping the hoist:

```ts
if (detail.element) {
  return;
}
event.preventDefault();
```

packages/ai-chat-components/src/components/markdown/__tests__/markdown.test.ts
```diff
@@ -1773,6 +1773,97 @@ HTTP: http://example.com
...
+
+    // ── customRenderers.table — container topology ──────────────────────────
+    // These tests use createRelocationHarness so the mount event is intercepted
+    // by a container-like ancestor (event.preventDefault() called, host hoisted
+    // to outer light DOM). The standalone fixture path was never broken; the
+    // regression only surfaces when a container ancestor owns the slot host.
+
+    const tableMarkdown = `| h1 | h2 |\n| --- | --- |\n| a | b |\n\nTrailer`;
+
+    it('customRenderers.table: host is hoisted to container light DOM', async () => {
```
ethanwinters: **Blocker** — this passes under `createRelocationHarness()` only because the harness appends a forwarder for every mount, including ones carrying `detail.element`. That models the wrapper's *old* behavior rather than the one this PR ships, so the case green-lights the regression flagged on `react/markdown.tsx` instead of catching it. The harness is outside the diff so I cannot suggest it here, but its mount listener needs the same guard before it appends the forwarder:

```ts
if (!detail.element && !forwarderOwners.has(detail.slotName)) {
```

With that in place this case fails until the container commits land beside it, which is what makes it a proof rather than a restatement of the current behavior.

ethanwinters (changes requested):
Rework — Bug 1 is sound and better than the issue asked for; Bug 2 ships its React half without the container half, which regresses `customRenderers` on the React path.

Both blockers are inline. A third has no diff line to hang on: the description says `ChatContainer.tsx`, `cds-aichat-container/index.ts` and `cds-aichat-custom-element/index.ts` return early without `preventDefault()`, but none of the three is in the diff — `gh pr diff --name-only` lists five files, all under `ai-chat-components`. #2220's fourth acceptance criterion names all three containers, so those commits need to land here.

Two things worth settling as you add them. First, the relocation you are removing exists so page-level CSS reaches the host — the comment above it says exactly that — so a consumer styling their custom table from a global stylesheet loses that reach. I assume that is the intended trade for working slot projection, but it is a behavior change and belongs in the PR body. Second, the three containers are not copy-paste identical: `cds-aichat-container` registers the slot name and checks `hasOuterChatHandler` before its `detail.element` branch, so the guard sits in a different place there than in `ChatContainer.tsx`.

Bug 1 is the strong half. Routing `softbreak`/`hardbreak` through `shouldDelegateToPluginRule` and adding them to `PLUGIN_DELEGABLE_TOKEN_TYPES` keeps plugin rule overrides working, which is better than dropping them.

**Would you take one more fix while you are in here?** (Tracked as #2270 either way.) `serializeInlineToken` (`src/components/markdown/src/utils/html-helpers.ts:348-361`) returns `token.content ?? ''` for `softbreak` and `hardbreak` — the empty string — and both types are in `INLINE_HTML_ALLOWED_TOKEN_TYPES`. So a break inside a merged inline-HTML run such as `<span>one\ntwo</span>` is still silently deleted after this PR, which makes "breaks are fixed" only partly true. It is outside the diff so I cannot suggest it inline, but the edit is:

```ts
function serializeInlineToken(tokenTree: TokenTree): string {
  const token = tokenTree.token;
  const type = token.type ?? '';

  if (type === 'softbreak' || type === 'hardbreak') {
    return '<br />';
  }

  if (
    type === 'text' ||
    type === 'code_inline' ||
    type === 'entity' ||
    type === 'html_inline'
  ) {
    return token.content ?? '';
  }
```

Not `'<br />\n'` — the trailing newline adds a stray whitespace text node. The merged node keeps `type: 'html_container'` and still routes through `sanitizeHtmlContent` then `unsafeHTML`, and DOMPurify's default allow-list keeps `<br>`, so it holds under both `sanitize-html` and `remove-html`. Entirely your call. Take it here and #2270 closes with this PR; leave it and I will land it there instead of growing this one further. #2270 is written assuming you leave it, so nothing is blocked either way.

One smaller thing: the branch is 24 commits behind `main`, which has since changed `markdown/src/utils/table-helpers.ts`, so the markdown suite should run against current `main` before merge.

No findings dropped.

ethanwinters:
Correction to one detail in my last review. I said the merged node keeps `type: 'html_container'`. It keeps `html_inline` — `combineConsecutiveHtmlInline` spreads `...startNode.token` and overrides only `content`, and `html_container` is set only by `combineSplitHtmlBlocks`, a different function. I've fixed the same error in #2270's body.

The suggested edit is unaffected. The `<br />` early return works either way, since both types route through `sanitizeHtmlContent` and then `unsafeHTML`.

The open question is still the one from that review: are you taking the `serializeInlineToken` fix here, or leaving it for #2270? Either works — #2270 is written assuming you leave it. I'd just like to know whether to schedule it.

davidmenendez:
> The open question is still the one from that review: are you taking the `serializeInlineToken` fix here, or leaving it for #2270? Either works — #2270 is written assuming you leave it. I'd just like to know whether to schedule it.

i'll leave it 👍

davidmenendez:
Every blocker from both reviews is addressed. Here's the full mapping:

---

**Review 1 (Aug 20) — all 4 blockers resolved:**

| Blocker | What was asked | Current state |
|---|---|---|
| **A** — container hoisting regresses global CSS for `config.markdown.customRenderers` | Narrow the fix | ✅ All three containers now return early for `detail.element` before `preventDefault()` |
| **B** — harness specs pass on unmodified `main` (no container ancestry) | Rebuild tests on `createRelocationHarness` | ✅ Both `customRenderers.table` tests use the harness; harness now mirrors real container behavior |
| **C** — `softbreak`/`hardbreak` bypass plugin delegation | Add `shouldDelegateToPluginRule` guard + add both to `PLUGIN_DELEGABLE_TOKEN_TYPES` | ✅ Guard in `markdown-renderer.ts:267`, both types in `markdown-token-tree.ts:79` |
| **D** — stale forwarder in `markdown.tsx` suppresses fallback when renderer returns null | Add `detail.element` guard to `handleMount` | ✅ `markdown.tsx:180` returns early for `detail.element` |

**Review 2 (Sep 2) — both blockers resolved:**

| Blocker | What was asked | Current state |
|---|---|---|
| **`markdown.tsx:182`** — containers still hoist while wrapper skips forwarder | Add container commits | ✅ All three containers have the guard |
| **`markdown.test.ts`** — harness still appends forwarder for `detail.element`, masking regression | Guard the harness the same way | ✅ Harness returns early for `detail.element`, no `preventDefault`, no hoist, no forwarder |

ethanwinters (approved):

PR #2243, opened 2026-08-26
fix(prompt-line): storybook follow-ups

### LINKED ISSUE #2065: [Task]: Clean up prompt-line storybook follow-ups from #2022
## Background

Review of #2022 (`feat(prompt-line): add storybook`) turned up a set of non-blocking cleanups, deferred here so they don't hold up the story work.

Two bugs were found and fixed in #2022 itself (commit `541578fd`), but **shipped without test coverage**, so their regression tests are tracked here:

1. **Escape failed to dismiss the conversation-starters overlay.** `dismiss()` fired a no-op ProseMirror transaction, which `carbonStarterTrigger.onTransaction` saw while the editor was still focused and empty, re-emitting the `starter` detail it had just cleared. Verified live: the trigger-change trace after one Escape press was `['starter', 'starter']`, overlay still open.
2. **`_handleClickOutside` couldn't see clicks on an anchor inside a shadow root.** `event.target` is retargeted to the shadow host at document level, so `anchorElement.contains(target)` never matched — which is the `@carbon/ai-chat` integration, where the prompt line lives inside a shadow root. Fixed by switching to `event.composedPath()`.

Neither has a test. The existing `dismiss() clears trigger and items` (`autocomplete-controller.test.ts:185`) uses a mock editor, so it structurally cannot catch either — bug 1 needs a real Tiptap editor wired to the controller, bug 2 needs a real shadow root.

## Goal

Both fixed bugs have regression coverage, the prompt-line stories follow the package's story conventions, no dead code remains, and the Storybook index contains only real stories.

### Regression tests for the bugs fixed in #2022

- [ ] A WTR case asserts Escape does not re-open the starter overlay: with a real Tiptap editor focused and empty and the starter trigger active, `controller.dismiss()` must leave the trigger `null` and emit no further `starter` detail. Model the editor setup on `tiptap/__tests__/carbon-starter-trigger.test.ts`, which already builds a real `Editor`.
- [ ] `autocomplete/__tests__/autocomplete.test.ts` covers `anchorElement` outside-click handling across all five cases below. A throwaway version of this was run against both sides of `541578fd`: case B failed pre-fix and passed post-fix in Chromium, Firefox **and** WebKit, while A/C/D/E passed on both — so B is the one that actually guards the fix.

  | Case | Expected |
  | --- | --- |
  | A — click on a light-DOM anchor | no dismiss |
  | **B — click on an anchor inside a shadow root** | **no dismiss** ← guards the fix |
  | C — genuine outside click | dismisses |
  | D — click inside the autocomplete itself | no dismiss |
  | E — no `anchorElement` set, outside click | dismisses |

  Case B needs a real shadow root, and the click must be dispatched with `{ bubbles: true, composed: true }` so it carries a genuine `composedPath()`:

  ```ts
  const host = document.createElement('div');
  document.body.appendChild(host);
  const root = host.attachShadow({ mode: 'open' });
  const anchor = document.createElement('div');
  anchor.innerHTML = '<span id="deep">editor text</span>';
  root.appendChild(anchor);
  el.anchorElement = anchor;

  root.querySelector('#deep')!.dispatchEvent(
    new MouseEvent('click', { bubbles: true, composed: true })
  );
  // expect: no cds-aichat-autocomplete-dismiss fired
  ```

- [ ] A Jest spec under `src/react/__tests__/` covers the mousedown-preventDefault focus retention on the React autocomplete hosts (`CustomElementHost` and `CustomReactNodePortal`).

### Cleanups

- [ ] `forceClear` is removed from `maybeEmit` in `packages/ai-chat-components/src/components/prompt-line/src/tiptap/carbon-starter-trigger.ts` — deleting `onBlur` left the parameter and its `!forceClear &&` guard unreachable.
- [ ] `_handleClickOutside` drops the double `.filter(isNode)` scan and the `!` assertion. `composedPath()` already contains every ancestor, so membership is the whole check:

  ```ts
  const path = event.composedPath();
  if (path.includes(this)) return;
  if (this.anchorElement && path.includes(this.anchorElement)) return;
  this._dismiss();
  ```

  Drop the `// Check if click is on autocomplete itself` / `// Check if click is on anchor element` comments, which restate the code; keep the `composedPath()` comment above them, which explains a non-obvious why.
- [ ] `preview-prompt-line--*` contains only the 6 intended stories plus Overview. The 12 `*API` / `*ArgTypes` exports in `__stories__/prompt-line-react.stories.jsx` are all `type: story` in the static build (`tags: ['test', 'manifest']`), so Chromatic snapshots them. Make the `*API` objects module-local consts and add `parameters: { chromatic: { disable: true } }` to the `*ArgTypes` exports that must stay exported for `<ArgTypes of={…}>`.
- [ ] `Expanded` composes from `Default` rather than duplicating ~50 lines of its render to hardcode one attribute. Add an `expanded` boolean control to the shared `argTypes`, keep it live on both stories, and set `args: { ...Default.args, expanded: true }` so the Chromatic baseline still covers the expanded layout.
- [ ] `CommandsAndMentions` disables the `attached` control it ignores, matching the other non-Typeahead stories.
- [ ] The `enableSendButton` control description no longer says "(Typeahead only)" — `CommandsAndMentions` uses it too.
- [ ] `useChatAutocomplete.tsx` uses one shared mousedown handler instead of a memoized callback on `CustomElementHost` and an inline arrow on `CustomReactNodePortal`; widen the type to `(e: Event | React.MouseEvent) => void` so both surfaces can share it. Drop the redundant `promptLineRef` dep.
- [ ] The `setConfigs` comment about the `isOn` toggle re-evaluating while unfocused is corrected — gating `onTransaction` on `editor.isFocused` made that no longer true.
- [ ] `references/component-authoring.md` documents nested sub-component directories (`prompt-line/autocomplete/`), which #2022 introduced alongside the existing flat shape.

## Out of scope

- The two fixes themselves — both landed in #2022 (`541578fd`). Only their missing tests are tracked here.
- Bringing the prompt-line Overview MDX in line with the `overview-lit.template.mdx` hero-Canvas shape. No shipped Overview follows it yet; that's a separate repo-wide pass.
- The autocomplete directory move itself. `packages/ai-chat-components/src/components/autocomplete/` is absent from every release tag through `@carbon/ai-chat-components@1.8.0`, so relocating it under `prompt-line/` breaks no published consumer.

## Related

- #2022 — the PR this review covers; both fixed bugs and all cleanups originate there
- #1347 — the story work #2022 closes

warrenmblood:
Closes #2065

Adds regression tests for two bugs that were fixed in #2022:
1) Escape key failed to dismiss the Prompt Line autocomplete overlay when using conversation starters
2) `_handleClickOutside` didn't detect clicks on an anchor inside a shadow root

Also fixes a series of small cleanup items (listed in #2065) related to the Prompt Line storybook.

#### Changelog

**New**

- `autocomplete.test.ts`: Regression test. Clicking an autocomplete anchor element inside a shadow root should not trigger the autocomplete to close.
- `autocomplete-controller.test.ts`: Regression test. Conversation starters should not re-open immediately after pressing Escape on an empty editor.

**Changed**

- `component-authoring.md`: Documents nested sub-component directory pattern.
- `prompt-line.stories.js`: Removes duplicate code by composing "Expanded" prompt line story from "Default".  In "Commands and mentions" story, disables `disableDirectSend` control as it has no effect on mentions/commands and makes `attached` control functional.
- `prompt-line-react.stories.jsx`: Applies the same fixes as the WC stories. Also makes the *API objects module-local consts and adds `parameters: { chromatic: { disable: true } }` to the `*ArgTypes` exports that must stay exported for `<ArgTypes of={…}>`. This stops Chromatic from taking blank snapshots of these exports.
- `autocomplete.ts`: Simplifies checks in `_handleClickOutside()`.
- `autocomplete-controller.ts`: Updates outdated comment.
- `carbon-starter-trigger.ts`: Removes `forceClear` parameter from `maybeEmit()` (dead code).
- `useChatAutocomplete.tsx`: Use `handleContainerMousedown` as mousedown handler for `CustomElementHost` and `CustomReactNodePortal`.

#### Testing / Reviewing
- Ensure both new regression tests pass. Already verified that they fail prior to #2022 commit.
- Go to Storybook deploy preview -> Prompt Line -> Commands and mentions. Verify toggling the `attached` control works.

### DIFF
--- a/packages/ai-chat-components/references/component-authoring.md
+++ b/packages/ai-chat-components/references/component-authoring.md
@@ -20,6 +20,34 @@ component-name/
 
 Shared pieces: [../src/components/shared/](../src/components/shared/); design tokens / utilities: [../src/globals/](../src/globals/); test helpers: [../src/testing/](../src/testing/); ambient types: [../src/typings/](../src/typings/).
 
+### Nested sub-component directories
+
+When a component grows a logically independent child element that can be used standalone (e.g. [`prompt-line/autocomplete/`](../src/components/prompt-line/autocomplete/)), that child lives as a **sibling directory** next to `src/` and `__stories__/` rather than as another file under `src/`:
+
+```
+component-name/
+  index.ts              # re-exports parent elements AND the sub-component's public API
+  src/                  # parent element implementation (unchanged)
+  __stories__/          # parent stories (unchanged)
+  __tests__/            # parent tests (unchanged)
+  sub-component/
+    index.ts            # sub-component's own public entry
+    src/
+      sub-component.ts
+      sub-component.scss
+    __stories__/        # stories specific to the sub-component
+    __tests__/
+      sub-component.test.ts
+```
+
+Rules for this pattern:
+
+- **Use it when** the child element has meaningful standalone use and its own story set. A helper element that only ever exists inside the parent belongs in `src/` instead.
+- **Both `index.ts` files matter.** The parent's `index.ts` re-exports the sub-component so consumers have one import path; the sub-component's own `index.ts` exists so it can also be imported directly.
+- **Types may flow upward.** If the sub-component's types are shared with the parent (e.g. `SuggestionItem` defined in `src/tiptap/types.ts` and re-used by `autocomplete/src/autocomplete.ts`), import from the parent rather than duplicating them.
+- **Stories are scoped to the sub-component directory.** The sub-component's `__stories__/` produces its own Storybook group; the parent's `__stories__/` does not need to duplicate those stories.
+- **Tests mirror the same split.** `__tests__/` directories exist at both levels; tests for the sub-component's element live under `sub-component/__tests__/`, not alongside the parent's tests.
+
 **Experimental components** use a `preview-*` prefix and a `Preview/` story title. Their APIs may change without a deprecation window and are not recommended for production; they graduate to the main directory (and `Components/` group) when stable.
 
 ## Generated artifacts — never hand-edit
--- a/packages/ai-chat-components/references/component-authoring.md
+++ b/packages/ai-chat-components/references/component-authoring.md
@@ -44,7 +44,7 @@ Rules for this pattern:
 
 - **Use it when** the child element has meaningful standalone use and its own story set. A helper element that only ever exists inside the parent belongs in `src/` instead.
 - **Both `index.ts` files matter.** The parent's `index.ts` re-exports the sub-component so consumers have one import path; the sub-component's own `index.ts` exists so it can also be imported directly.
-- **Types may flow upward.** If the sub-component's types are shared with the parent (e.g. `SuggestionItem` defined in `src/tiptap/types.ts` and re-used by `autocomplete/src/autocomplete.ts`), import from the parent rather than duplicating them.
+- **Types may flow upward.** If the sub-component's types are shared with the parent (e.g. `SuggestionItem` defined in `src/components/prompt-line/src/tiptap/types.ts` and re-used by `autocomplete/src/autocomplete.ts`), import from the parent rather than duplicating them.
 - **Stories are scoped to the sub-component directory.** The sub-component's `__stories__/` produces its own Storybook group; the parent's `__stories__/` does not need to duplicate those stories.
 - **Tests mirror the same split.** `__tests__/` directories exist at both levels; tests for the sub-component's element live under `sub-component/__tests__/`, not alongside the parent's tests.
 
--- a/packages/ai-chat-components/src/components/prompt-line/__stories__/prompt-line-react.stories.jsx
+++ b/packages/ai-chat-components/src/components/prompt-line/__stories__/prompt-line-react.stories.jsx
@@ -823,7 +823,7 @@ const _disableMeta = Object.fromEntries(
   Object.keys(WCMeta.argTypes).map((k) => [k, { table: { disable: true } }])
 );
 
-export const PromptLineAPI = {
+const PromptLineAPI = {
   tags: ['!dev'],
   name: 'PromptLine',
   argTypes: {
@@ -875,7 +875,7 @@ export const PromptLineAPI = {
   ),
 };
 
-export const PromptLineShellAPI = {
+const PromptLineShellAPI = {
   tags: ['!dev'],
   name: 'PromptLineShell',
   argTypes: {
@@ -919,7 +919,7 @@ export const PromptLineShellAPI = {
   ),
 };
 
-export const CDSAIChatInputSendControlAPI = {
+const CDSAIChatInputSendControlAPI = {
   tags: ['!dev'],
   name: 'CDSAIChatInputSendControl',
   argTypes: {
@@ -978,7 +978,7 @@ export const CDSAIChatInputSendControlAPI = {
   ),
 };
 
-export const CDSAIChatAutocompleteAPI = {
+const CDSAIChatAutocompleteAPI = {
   tags: ['!dev'],
   name: 'CDSAIChatAutocomplete',
   argTypes: {
@@ -1025,7 +1025,7 @@ export const CDSAIChatAutocompleteAPI = {
   ),
 };
 
-export const CDSAIChatFileUploadsAPI = {
+const CDSAIChatFileUploadsAPI = {
   tags: ['!dev'],
   name: 'CDSAIChatFileUploads',
   argTypes: {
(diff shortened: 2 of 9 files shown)


packages/ai-chat-components/src/components/prompt-line/__stories__/prompt-line-react.stories.jsx
```diff
@@ -282,60 +292,8 @@ export const Default = {
...
-            disabled={disabled}
-            hasValidInput={hasValidInput}
-            onSend={() => action('cds-aichat-input-send')()}
-          />
-        </PromptLineShell>
-      </Wrapper>
-    );
-  },
+  args: { ...Default.args, expanded: true },
+  render: Default.render,
```
AlexanderMelox, line 300: Did we mean to remove argTypes too here? Should we include it on the new expanded?

packages/ai-chat-components/src/components/prompt-line/__stories__/prompt-line.stories.js
```diff
@@ -930,75 +975,18 @@ export const Default = {
...
-            @cds-aichat-input-send=${() => action('cds-aichat-input-send')()}
-            ${ref((el) => {
-              sendControlEl = el ?? null;
-            })}></cds-aichat-input-send-control>
-        </cds-aichat-prompt-line-shell>
-      </div>
-    `;
-  },
+  args: { ...Default.args, expanded: true },
+  render: Default.render,
```
AlexanderMelox, line 983: Same here, was argTypes supposed to be removed?

packages/ai-chat-components/src/components/prompt-line/autocomplete/__tests__/autocomplete.test.ts
```diff
@@ -460,6 +460,40 @@ describe('cds-aichat-autocomplete', () => {
...
+      root
+        .querySelector('#deep')!
+        .dispatchEvent(
+          new MouseEvent('click', { bubbles: true, composed: true })
+        );
+
+      let dismissEventFired = false;
+      el.addEventListener('cds-aichat-autocomplete-dismiss', () => {
+        dismissEventFired = true;
+      });
```
AlexanderMelox:
```suggestion
      let dismissEventFired = false;
      el.addEventListener('cds-aichat-autocomplete-dismiss', () => {
        dismissEventFired = true;
      });
      root
        .querySelector('#deep')!
        .dispatchEvent(
          new MouseEvent('click', { bubbles: true, composed: true })
        );
```

We should move the dismissed event before calling the #deep click. If `path.includes(this.anchorElement)` regressed for descendants, that dispatch would silently call `_dismiss()`, the later click on anchor itself would still return early, and the test would stay green.

packages/ai-chat-components/references/component-authoring.md
```diff
@@ -20,6 +20,34 @@ component-name/
...
+    __stories__/        # stories specific to the sub-component
+    __tests__/
+      sub-component.test.ts
+```
+
+Rules for this pattern:
+
+- **Use it when** the child element has meaningful standalone use and its own story set. A helper element that only ever exists inside the parent belongs in `src/` instead.
+- **Both `index.ts` files matter.** The parent's `index.ts` re-exports the sub-component so consumers have one import path; the sub-component's own `index.ts` exists so it can also be imported directly.
+- **Types may flow upward.** If the sub-component's types are shared with the parent (e.g. `SuggestionItem` defined in `src/tiptap/types.ts` and re-used by `autocomplete/src/autocomplete.ts`), import from the parent rather than duplicating them.
```
AlexanderMelox:
```suggestion
- **Types may flow upward.** If the sub-component's types are shared with the parent (e.g. `SuggestionItem` defined in `src/components/prompt-line/src/tiptap/types.ts` and re-used by `autocomplete/src/autocomplete.ts`), import from the parent rather than duplicating them.
```

component-authoring.md:47 — `src/tiptap/types.ts` resolves relative to `references/`, where it doesn't exist, so validate:agents now emits the only warning in the repo (File reference may be outdated) — base is clean. An agent following the doc from that location finds nothing. Use the path from the package root: `src/components/prompt-line/src/tiptap/types.ts`.

AlexanderMelox (changes requested):

packages/ai-chat-components/src/components/prompt-line/__stories__/prompt-line-react.stories.jsx
warrenmblood: yes- this story doesn't have an autocomplete overlay so argTypes that disable the autocomplete controls should be included here.

AlexanderMelox (approved):

PR #2244, opened 2026-08-26
fix: add hover select to autocomplete

### LINKED ISSUE #2156: [Bug]: Custom suggestion lists ignore arrow keys and hover desyncs the active option
## Background

Two defects in the suggestion list's interaction model. Both reproduce on `examples/react/prompt-line-conversation-starters`; the second also reproduces on `examples/react/prompt-line-typeahead`.

### 1. Arrow keys do nothing when a host supplies `renderCustomList`

The list announces "Use up and down arrows to move, Enter to pick, Escape to close", but ArrowUp, ArrowDown, Enter, and Escape have no effect.

Keys pressed in the editor are re-dispatched as a synthetic `keydown` onto `this._listElement` (`autocomplete-controller.ts:425-449`). Which element that is depends on the path:

- Built-in list: `ref={setListElement}` hands over the `cds-aichat-autocomplete` element itself (`useChatAutocomplete.tsx:237`), whose own listener (`autocomplete.ts:206`) handles the key. This works.
- Custom list: `setListElement` receives the `CustomReactNodePortal` wrapper (`useChatAutocomplete.tsx:227-231`), which is an **ancestor** of the host's autocomplete element. The synthetic event bubbles up from the wrapper, so the element nested inside never sees it.

Measured on the starters example:

| Dispatch target                            | Active option moves |
| ------------------------------------------ | ------------------- |
| wrapper `div` — what the controller holds   | no                  |
| `cds-aichat-autocomplete` — nested inside   | yes                 |

A real ArrowDown on the built-in list (`prompt-line-typeahead`) does move the active option, confirming the split.

Tab leaves the widget and closes the list, which is correct combobox behavior on its own. Combined with dead arrow keys it means there is no keyboard path to a custom suggestion list at all — WCAG 2.1 A, 2.1.1 Keyboard.

### 2. Hover shades a second row without moving the active option

Hovering a non-active row shades it while the active row stays shaded, so two rows read as selected at once. `aria-activedescendant` does not follow the pointer.

Measured with option 1 active and option 3 hovered:

| option | `aria-selected` | background                     |
| ------ | --------------- | ------------------------------ |
| 1      | `true`          | `$background-selected` (0.2 α)  |
| 3      | `false`         | `$background-hover` (0.12 α)    |

This one is not custom-list specific — it reproduces on the built-in list too.

The result is a mis-selection hazard: with the pointer resting on option 3, pressing Enter acts on option 1. The row the user is pointing at is not the row that gets picked.

## Goal

Keyboard and pointer both drive a single active option, for built-in and custom lists alike.

## Acceptance criteria

- [ ] ArrowUp, ArrowDown, Enter, and Escape work when a host supplies `renderCustomList`, for both the `HTMLElement` and React-node return shapes.
- [ ] Keys reach the rendered list regardless of how deeply the host nests it inside the returned tree.
- [ ] Hovering a row makes it the active option: `aria-activedescendant` follows the pointer and exactly one row is shaded.
- [ ] Enter picks the row the pointer is on.
- [ ] Built-in list keyboard behavior is unchanged.
- [ ] Coverage drives real keyboard input against both the built-in and custom list shapes.

## Approach notes

For the keys: dispatching an event on a wrapper can never reach a descendant, so registering the wrapper is the defect. Either resolve the real list element before registering it, or forward the key by calling an API on the list instead of dispatching a bubbling event.

For hover: pointer entry should set the same focused index the arrow keys set, so there is one source of truth for the active option rather than a CSS `:hover` rule running in parallel with it.

## Out of scope

- Tab closing the list. That is correct combobox behavior and should stay.
- #2147 — grouped listbox ownership.
- #2150, #2152 — unrelated prompt-line defects.

## Related

- `packages/ai-chat-components/src/components/prompt-line/src/autocomplete-controller.ts:425-449` — key re-dispatch onto `_listElement`.
- `packages/ai-chat-components/src/react/hooks/useChatAutocomplete.tsx:227-231` — custom list registers the portal wrapper.
- `packages/ai-chat-components/src/react/hooks/useChatAutocomplete.tsx:237` — built-in list registers the element itself.
- `packages/ai-chat-components/src/components/prompt-line/autocomplete/src/autocomplete.ts:206,286` — the element's own keydown listener.
- `examples/react/prompt-line-conversation-starters`, `examples/react/prompt-line-typeahead` — reproductions.

davidmenendez:
Closes #2156 

adds a new method and mouse event handler for setting an item as active in the autocomplete component

#### Changelog

**Changed**

- `packages/ai-chat-components/src/components/prompt-line/autocomplete/src/autocomplete.ts` adds `_handleItemMouseEnter` to handle focusing suggestions on hover
- `packages/ai-chat-components/src/components/prompt-line/autocomplete/__tests__/autocomplete.test.ts` adds testing

#### Testing / Reviewing

run the storybook and go to the `preview-prompt-line-autocomplete--default` story. hover over one of the items, move the mouse away, and verify that the last item that was hovered over has the active class on it

### DIFF
--- a/packages/ai-chat-components/src/components/prompt-line/autocomplete/__tests__/autocomplete.test.ts
+++ b/packages/ai-chat-components/src/components/prompt-line/autocomplete/__tests__/autocomplete.test.ts
@@ -463,6 +463,33 @@ describe('cds-aichat-autocomplete', () => {
   });
 
   describe('keyboard navigation', () => {
+    it('responds to a synthetic ArrowDown dispatched directly on the element (controller forwarding path)', async () => {
+      // The autocomplete-controller dispatches a synthetic KeyboardEvent directly
+      // on the registered list element. Verify the element handles it identically
+      // to a user-initiated keydown so the custom-list code path works.
+      const el = await fixture<AutocompleteElement>(html`
+        <cds-aichat-autocomplete
+          .items="${mockItems}"></cds-aichat-autocomplete>
+      `);
+
+      el.dispatchEvent(
+        new KeyboardEvent('keydown', {
+          key: 'ArrowDown',
+          bubbles: true,
+          cancelable: true,
+        })
+      );
+      await el.updateComplete;
+
+      const options = el.shadowRoot?.querySelectorAll('li[role="option"]');
+      expect(
+        (options?.[1] as HTMLElement)?.getAttribute('aria-selected')
+      ).to.equal('true');
+      expect(
+        (options?.[0] as HTMLElement)?.getAttribute('aria-selected')
+      ).to.equal('false');
+    });
+
     it('should move focus down with ArrowDown', async () => {
       const el = await fixture<AutocompleteElement>(html`
         <cds-aichat-autocomplete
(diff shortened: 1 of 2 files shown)


packages/ai-chat-components/src/components/prompt-line/autocomplete/__tests__/autocomplete.test.ts
```diff
@@ -610,6 +637,120 @@ describe('cds-aichat-autocomplete', () => {
...
+      // Hover over the second item
+      (options?.[1] as HTMLElement)?.dispatchEvent(
+        new MouseEvent('mouseenter', { bubbles: true })
+      );
+      await el.updateComplete;
+
+      const updatedOptions =
+        el.shadowRoot?.querySelectorAll('li[role="option"]');
+      expect(
+        (updatedOptions?.[1] as HTMLElement)?.getAttribute('aria-selected')
```
warrenmblood: Right now we set `aria-selected` true when a suggestion item is hovered (or "active") when `aria-activedescendant` is already indicating the same thing using the more appropriate aria attribute. Can we remove setting `aria-selected` from `autocomplete.ts`? Accordingly, we could remove this test, then in the `aria-activedescendant` test (starting line 669) also check that the previously active item has `aria-activedescendant = false` when a new item receives hover.

warrenmblood (changes requested):

packages/ai-chat-components/src/components/prompt-line/autocomplete/__tests__/autocomplete.test.ts
davidmenendez: good catch, the hover shouldn't indicate selected state 👍

warrenmblood (approved):
LGTM!

PR #2250, opened 2026-08-28
feat(examples): demonstrate a rejected upload in file-upload examples

### LINKED ISSUE #2108: Demonstrate a rejected upload in the prompt-line-file-upload examples
### Background

- `mockOnFileUpload` in both `prompt-line-file-upload` examples always succeeds after a 1s delay. Neither flavor shows a failed upload.
- `onFileUpload` is documented to throw on failure, but no example shows what that produces, where the error appears, or what the user does next.
- This goes in the existing examples, not a new one — rejecting an upload is part of implementing `onFileUpload`, not a separate concern.
- Parent epic: #2103

### Goal

Both `prompt-line-file-upload` examples demonstrate a rejected upload, and their READMEs describe it.

### Acceptance criteria

- [ ] `mockOnFileUpload` rejects on a fixed, reproducible condition in both flavors. It throws after the simulated delay, so the chip visibly goes from uploading to errored.
- [ ] The welcome message states the rejection condition, so it's discoverable without reading source.
- [ ] React and Web Components use the same condition and the same rejection message.
- [ ] The rejection message reads like a real server error, not example scaffolding.
- [ ] Each README documents the failure path on its own, without referencing the sibling example.
- [ ] `npm run verify:example-readmes` exits 0.
- [ ] Touched files keep accurate top-of-file comments; the rejection has a comment explaining why.
- [ ] `examples/**` definition-of-done gate passes for both examples ([definition-of-done.md](https://github.com/carbon-design-system/carbon-ai-chat/blob/main/references/definition-of-done.md)).

### Public API / contract

No public API change.

### Out of scope

- `accept`, `maxFileSizeBytes`, `maxFiles` — these reject before `onFileUpload` runs.
- A separate error-handling example.
- Playwright tests for these examples.

### Related

- Parent epic: #2103
- A working version of the mock handler existed on the `prompt-line-review-fixes` branch (React only): reject filenames starting with `a`, plus a matching welcome-message line. It was reverted, not committed, but is the shape this issue wants.

AlexanderMelox:
Closes #2108

Screenshot 2026-08-28 at 9 35 38 AM

Both `prompt-line-file-upload` examples only showed the example on how it works but not what happens if the file upload fails. `onFileUpload`'s contract says "throw to reject", but nothing demonstrated what that produces — a host wiring a real endpoint had a template for success and none for failure.

Files whose name starts with "a" now fail. The check sits after the existing delay/abort race, not before it: throwing earlier would skip the in-flight state, and the uploading → errored transition is the point.

#### Changelog

**New**

- Upload failure path in both `prompt-line-file-upload` examples. Names starting with "a" throw from the mock `onFileUpload`.

**Changed**

- Welcome message and both example READMEs state the condition.

#### Testing / Reviewing

Not reachable from the demo site — run the example.

```bash
npm run start --workspace=@carbon/ai-chat-examples-react-prompt-line-file-upload
```

1. Welcome message states the "a" condition.
2. Attach `apple.txt`. Chip goes uploading → errored. Banner shows the rejection message plus the recovery step.
3. Type some text. Send stays disabled.
4. Remove the file. Banner clears, send re-enables.
5. Attach `avocado.txt` and `budget.csv` together. One errors, one succeeds, no extra handling.
6. Repeat on `@carbon/ai-chat-examples-web-components-prompt-line-file-upload`. Same condition, same message.

### DIFF
--- a/examples/react/prompt-line-file-upload/README.md
+++ b/examples/react/prompt-line-file-upload/README.md
@@ -8,6 +8,7 @@
 - Simulating a 1-second upload with `AbortSignal` support and returning an `ExternalFileReference` wrapped in `StructuredData`.
 - The chat rendering each uploaded file as a chip in the user's own message bubble, from the `name` and `mime_type` on the returned reference.
 - Echoing attached file metadata back as a text message via `instance.messaging.addMessage` and `MessageResponseTypes.TEXT`, showing what the server receives.
+- Demonstrating the upload failure path: attaching a file whose name starts with "a" makes `onFileUpload` throw after the simulated delay, and the chat shows the attachment chip moving from uploading to errored with a description of what went wrong.
 - Documenting the optional `accept`, `maxFileSizeBytes`, and `maxFiles` config knobs (commented in source).
 
 ## When to use this pattern
--- a/examples/react/prompt-line-file-upload/src/customSendMessage.ts
+++ b/examples/react/prompt-line-file-upload/src/customSendMessage.ts
@@ -36,7 +36,9 @@ const WELCOME_TEXT = `Welcome! This example demonstrates file uploads in Carbon 
 
 Attach a file using the paperclip button in the input area, then send a message. The mock server will echo back the file metadata.
 
-You can attach multiple files at once before sending.`;
+You can attach multiple files at once before sending.
+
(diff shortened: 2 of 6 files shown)


warrenmblood (approved):
LGTM

PR #2260, opened 2026-09-01
fix(prompt-line): preserve keyboard focus state across rich editor upgrade

### LINKED ISSUE #2187: [Bug]: Focus origin is lost across the textarea to rich upgrade
### Background

`<cds-aichat-prompt-line>` suppresses the focus ring for mouse-driven focus and shows it for keyboard focus. Each controller tracks the origin in a private `_focusFromMouse`, set by `pointerdown`/`mousedown`/`touchstart` on the host and consumed by the next focus event.

The textarea→rich upgrade loses that bit. `_swapToRich` deliberately carries everything else across:

```ts
const value = previous.getValue();
const selection = previous.getSelection();
const hadFocus = previous.hasFocus();
…
if (hadFocus) { rich.focus(); }
```

But `RichController.focus()` is:

```ts
focus(): void {
  this._focusFromMouse = true;
  this._editor?.commands.focus();
}
```

So the restored focus emits `cds-aichat-prompt-focus` with `keyboard: false`, and `prompt-line-shell` clears its ring. A keyboard user who tabs into the input and triggers the upgrade — by typing, or by a host calling `ensureEditor()` — lands in the rich editor with no visible focus indicator.

`RichController._recreateEditor` proves the bit is worth preserving; it restores the origin explicitly across a recreate. `_swapToRich` cannot do the same, because `PromptLineController` exposes only `focus()` / `blur()` / `hasFocus()`, and `TextareaController` consumes its `_focusFromMouse` inside `_onFocus` rather than retaining it.

Behind that sits a duplication. The mount blocks and `_setMouseFlag` are byte-identical between `prompt-line-controller.ts` and `prompt-line-rich-runtime.ts` — one policy implemented twice. What diverged is the visual half: the rich controller adds a `keyboard-focus` class backed by `outline: revert`, while the textarea sets `outline: none` unconditionally with nothing to revert it. So the two surfaces disagree about what a keyboard user sees, and a fix applied to one controller silently misses the other.

Not a regression from #2152 — both halves predate it.

### Goal

Focus origin survives the textarea→rich upgrade, so a keyboard user keeps the ring and a mouse user still gets none. The policy lives in one place that both controllers read.

### Public API / contract

`cds-aichat-prompt-focus` already carries `{ keyboard: boolean }`, and that detail is observable by hosts and by `prompt-line-shell`. No signature changes; the fix corrects when `keyboard: true` is reported.

- **Derivation**: `keyboard` is `true` unless a pointer/mouse/touch event preceded this focus on the same surface. Today a focus restored by `_swapToRich` always reports `false` regardless of what the user did — that is the bug. After the fix, an upgrade that preserves focus reports the origin of the interaction that first gave the surface focus.
- **No-op case**: an upgrade with no prior focus emits nothing, unchanged.
- **Preconditions / Failure / Timing / Repeat calls / Events / Ownership**: unchanged. No new events; ordering unchanged.
- **Announcement**: the focus ring is the accessible half of this. A keyboard user losing the indicator while retaining focus is a WCAG 2.4.7 failure.
- **Defaults**: unchanged.

**Settle before implementing**: whether the textarea gets the same `outline: revert` treatment the rich editor has, or the inner class goes away and the shell's outline becomes the only ring. Today they differ, and the answer decides whether this is one rule or two.

### Acceptance criteria

- [ ] A keyboard user focused in the textarea still has the focus ring after the surface upgrades to rich. Covered in `packages/ai-chat-components/src/components/prompt-line/src/__tests__/prompt-line.test.ts`.
- [ ] A mouse user focused in the textarea has no ring after the same upgrade. Same spec.
- [ ] Both surfaces report the same `keyboard` value in `cds-aichat-prompt-focus` for the same input sequence — pointer-then-focus, and focus alone. Same spec.
- [ ] The focus-origin policy exists once: one module owns "was this focus mouse-driven", and neither controller carries its own copy of the listener set. Verifiable by a single definition of the pointer/mouse/touch handler in the package.
- [ ] The inner-ring decision above is recorded in the code that implements it, so the next reader does not have to rediscover which surface draws what.
- [ ] The gate in [definition-of-done.md](https://github.com/carbon-design-system/carbon-ai-chat/blob/main/references/definition-of-done.md) passes: `npm run build && npm run ci-check`, build first.

### Out of scope

- Focus restoration across a DOM move, which is #2186.
- The shell's own focus-ring styling beyond the decision named above.
- Any other duplication between the two controllers.

### Related

- #2186 — focus across a reparent, the other half of the prompt-line focus story.

warrenmblood:
Closes #2187

Fixes bug where the prompt line keyboard-focus state wasn't preserved across the upgrade from lite to rich editor. The keyboard-focus state determines whether the blue focus outline should appear around the prompt line editor/text area. When focus is not keyboard-triggered (i.e. via Mouse), the outline is not visible.

#### Changelog

**New**
- `prompt-line.test.ts`: Tests the keyboard-focus flag is preserved across the upgrade from lite editor to rich.
- `prompt-line-editor-stability.test.ts`: Test that both keyboard-focus and non-keyboard-focus (mouse) states are preserved across a rebuild.

**Changed**
- `TextareaController` and `RichController` previously held their own logic for managing the keyboard-focus state (both tracked via a `_focusFromMouse`). These changes remove this duplication by moving the mouse-focus tracking to an abstract class, `MouseFocusController`, which is implemented by both `TextareaController` and `RichController`.
- Re-organized files to a cleaner split:
  - `prompt-line-rich-runtime.ts`: Serves same purpose as before. Implements class `RichController`, the rich editing-surface controller.
  - `prompt-line-textarea-runtime.ts`: Implements the class `TextareaController`, the default editing-surface controller.
  - `prompt-line-types.ts`: Shared types for the prompt-line. Previously lived inside `packages/ai-chat-components/src/components/prompt-line/src/prompt-line-controller.ts` along with the `TextareaController` implementation.
  - `prompt-line-mouse-focus.ts`: Shared mouse-focus tracking used by `TextareaController` and `RichController`.
- `PromptLineController.focus()` accepts `keyboardFocus`, which when true specifies that a focus ring should be visible around the prompt line text area. When a lite editor is upgraded to rich in `_swapToRich()`, the keyboard-focus state is preserved by passing the previous value into `PromptLineController.focus()`.
- Updated imports according to file changes.

**Removed**

- #2187 references logic that applies `outline: 'revert'` to the rich editor in `ensureKeyboardFocusRule()`, which disagrees with how the lite editor handles keyboard focus. This is effectively dead code that has no visual impact. The actual outline styles exist in `prompt-line-shell.scss` and are driven by the class `cds-aichat--input-text-area--keyboard-focus`. This PR removes this dead code and a corresponding test (`prompt-line-editor-stability.test.ts`).

#### Testing / Reviewing

1. Verify new test passes
2. Go to demo deploy preview -> Chat Configuration -> Input -> Enable expanded layout
3. Verify no regressions in focusing input via mouse (no focus outline) or via keyboard (blue focus outline appears)
4. Set "Enable autocomplete suggestions" to true to switch prompt line to rich mode
5. Repeat Step 3 to verify no regressions in rich editor

### DIFF
--- a/packages/ai-chat-components/src/components/prompt-line/src/__tests__/prompt-line-editor-stability.test.ts
+++ b/packages/ai-chat-components/src/components/prompt-line/src/__tests__/prompt-line-editor-stability.test.ts
@@ -13,12 +13,11 @@
  * its undo history mid-typing.
  */
 
-import { expect, fixture, html, nextFrame } from '@open-wc/testing';
+import { expect, fixture, html, nextFrame, oneEvent } from '@open-wc/testing';
 import { Extension } from '@tiptap/core';
 
 import '../prompt-line.js';
 import type PromptLineElement from '../prompt-line.js';
-import { PM_KEYBOARD_FOCUS_CLASS } from '../prompt-line-rich-runtime.js';
 import { buildCarbonExtensions } from '../tiptap/build-extensions.js';
 import {
   readStarterStorage,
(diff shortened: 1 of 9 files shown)


davidmenendez (approved):
LGTM!

Issue #1722, opened 2026-07-13
Update ai-chat, ai-chat-components and demo to babel 8, and demo to webpack-dev-server 6


ethanwinters:
### Background

Sub-issue of #1932. First scoped as "examples, ai-chat and ai-chat-components to babel 8, webpack-dev-server 6, webpack-cli 7". Two things have narrowed it since:

- The examples move to Vite under a sibling child of #1932. Vite transforms with Oxc. So the 84 per-example babel and webpack deps get deleted, not upgraded. Nothing babel-related survives in `examples/` for this issue.
- `demo/package.json` already declares `webpack-cli: ~7.2.0`. That half of the webpack bump is done.

What remains: babel 8 across the two packages and `demo/`, plus webpack-dev-server 5 → 6 for `demo/` alone. `demo/` stays on webpack on purpose, as the repo's webpack regression surface.

Babel 8 is a real major here, not a version bump, and one dependency makes it awkward.

`babel-preset-carbon` is loaded on every lint run. `eslint-config-carbon` sets `parser: '@babel/eslint-parser'` and passes `presets: ['babel-preset-carbon']`, but only lists that preset in its own devDependencies. So the root `package.json` has to declare it. Its 0.9.0 release requires `@babel/core ^8`, which would nest a second babel major under a preset eslint loads constantly. Meanwhile `eslint-config-carbon@3.10.0` pins the preset at `^0.5.0`, and moving that config forward needs eslint 9.

That makes the babel-preset bump, the eslint-config bump, and #1014 one upgrade, not three. Sequence them together or the tree ends up with two babel majors.

### Goal

Both packages and `demo/` build and test on babel 8. `demo/` runs on webpack-dev-server 6.

### Acceptance criteria

- [ ] `@carbon/ai-chat` builds on babel 8. Its rollup build reads `.babelrc` through `@rollup/plugin-babel` (`packages/ai-chat/tasks/rollup.aichat.js:184`). Proof: `npm run build --workspace=@carbon/ai-chat` exits 0, and the emitted `dist/es/` entry points still load.
- [ ] Both packages' jest suites run on babel 8 through `babel-jest`. Proof: `npm run test` exits 0 for each workspace. No test skipped and no assertion loosened to get there.
[...]

PR #2261, opened 2026-09-02
chore(deps): remove unused babel-preset-carbon

ethanwinters:
Closes #

Drops `babel-preset-carbon`, a root devDependency nothing loads.

It has been declared since the initial `ai-chat` setup (#41) and referenced nowhere since. All three babel configs — `demo/.babelrc`, `packages/ai-chat/.babelrc`, `packages/ai-chat-components/babel.config.jest.cjs` — list `@babel/preset-*` directly.

Surfaced by the 0.9.0 renovate bump (#2253). That release moves the preset to `@babel/core ^8`, so npm nested a private 101-package Babel 8 tree under a preset no config references — about 1730 lines of lockfile for nothing. Removing is cheaper than carrying it. Babel 8 stays tracked under #1722.

#### Changelog

**Removed**

- `babel-preset-carbon` root devDependency.

#### Testing / Reviewing

No behavior to exercise — nothing imported the preset. Confirm it was dead and the tree still resolves:

1. `git grep babel-preset-carbon` — only hits are removed by this PR.
2. `npm ci` — resolves clean.
3. `npm run build && npm run ci-check`.

Run on this branch: build succeeded for 89 projects; `ci-check` passed for 7 — `ai-chat-components` 564 Chromium / 564 Firefox, demo Playwright 36.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,7 +29,6 @@
         "@types/prop-types": "^15.7.15",
         "@types/react": "^19.2.2",
         "@types/react-dom": "^19.2.1",
-        "babel-preset-carbon": "^0.8.0",
         "chromatic": "^17.0.0",
         "commander": "^15.0.0",
         "concurrently": "^10.0.0",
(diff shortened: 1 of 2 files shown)


ethanwinters:
Closing — the premise is wrong. `babel-preset-carbon` is not unused.

`eslint-config-carbon` configures the babel parser with it:

```js
// node_modules/eslint-config-carbon/base.js:11-16
parser: '@babel/eslint-parser',
parserOptions: {
  babelOptions: { presets: ['babel-preset-carbon'] },
  requireConfigFile: false,
},
```

Same in its `vanilla.js`. It lists `babel-preset-carbon` only in its own devDependencies (`^0.5.0`), so it never installs it for consumers — the root `package.json` declaration is what satisfies it. Removing it produces 72 `Parsing error: Cannot find module 'babel-preset-carbon'` failures, which is the lint job on this PR.

Worth recording why the local gate went green: `npm install --package-lock-only` updates the lockfile without touching `node_modules`, so build, test, and lint all ran against a tree that still had the package installed. Only `npm ci` surfaces it.

The underlying tension is real but different from what this PR assumed: `babel-preset-carbon@0.9.0` requires `@babel/core ^8`, while `eslint-config-carbon@3.10.0` wants `babel-preset-carbon@^0.5.0`. Those move together, so #2253 and #2254 are one upgrade rather than two independent bumps.

PR #2295, opened 2026-09-05
fix(markdown): render line breaks inside merged inline-HTML runs and in InlineMarkdown

### LINKED ISSUE #2270: Line breaks are dropped inside merged inline-HTML runs
### Background

`serializeInlineToken` (`packages/ai-chat-components/src/components/markdown/src/utils/html-helpers.ts:346-361`) returns `token.content ?? ''` for both break types — the empty string, because break tokens carry no content:

```ts
if (
  type === 'text' ||
  type === 'code_inline' ||
  type === 'softbreak' ||
  type === 'hardbreak' ||
  type === 'entity' ||
  type === 'html_inline'
) {
  return token.content ?? '';
}
```

Both `softbreak` and `hardbreak` are in `INLINE_HTML_ALLOWED_TOKEN_TYPES` (`html-helpers.ts:32-41`), so they are eligible to be merged into an inline-HTML run. When markdown like `<span>one\ntwo</span>` collapses into a single `html_inline` node, the break between `one` and `two` is serialized to `''` and silently deleted.

This path never reaches the token renderer: the merged node is serialized to an HTML string and handed to `unsafeHTML`. So the main-path break fix in #2222, which returns `<br>` from `renderTokenTree` for both token types, cannot reach it. The two are independent and either can land first.

Not new: no commit has ever touched `softbreak`, `hardbreak`, or `<br>` in this file, and `markdown.test.ts` has no coverage for a break inside an inline-HTML run.

### Goal

A line break inside a merged inline-HTML run renders as a `<br>`, matching how the same break renders outside one.

### Acceptance criteria

- [ ] A soft break inside a merged inline-HTML run (`<span>one\ntwo</span>`) renders one `<br>` and text content `onetwo`. Proof: new describe in `packages/ai-chat-components/src/components/markdown/__tests__/markdown.test.ts`. Fails today.
- [ ] A hard break in the same position renders one `<br>`, covering `hardbreak` separately from `softbreak`. Proof: same spec.
- [ ] Both hold with `sanitize-html` set. Proof: same spec. DOMPurify's default allow-list keeps `<br>`, so this is a check that the merged node still routes through `sanitizeHtmlContent` rather than an assumption about it.
- [ ] Both hold with `remove-html` set. Proof: same spec.
- [ ] The fix introduces no stray whitespace text node — the run's child nodes are exactly the text before, the `<br>`, and the text after. Proof: same spec asserting `childNodes` rather than `textContent`, which is what distinguishes `'<br />'` from `'<br />\n'`.
- [ ] Existing inline-HTML merge behavior is unchanged. Proof: the existing `describe('markdown between HTML tags')` cases in the same spec pass **unchanged** — no assertion loosened, no case skipped.
- [ ] `npm run build --workspace=@carbon/ai-chat-components` and `npm run test --workspace=@carbon/ai-chat-components` exit 0, plus `npm run lint` and `npm run lint:license`. Gate rows, their ordering rule and the watcher precondition: [definition-of-done.md](https://github.com/carbon-design-system/carbon-ai-chat/blob/main/references/definition-of-done.md).

### Public API / contract

No signature changes. Behavioral, on rendered markdown output.

| Lock | Answer |
| --- | --- |
| Preconditions | A line break inside an inline-HTML run that `combineConsecutiveHtmlInline` collapses into one `html_inline` node. Breaks outside such a run are #2222's path and unaffected. |
| No-op case | Inline HTML with no line break in it serializes exactly as today. |
| Failure | Never throws. |
| Events | None. |
| Timing | Synchronous within the element's own render. |
| Repeat calls | Pure function of its token tree. |
| Defaults | Unchanged. `breaks: true` stays the parser default. |
| Derivation | N/A. |
| Announcement | A break that was previously deleted now reaches the accessibility tree as the line structure the author wrote. Focus unaffected. |
| Ownership | Framework. Consumers do nothing. |

**Proposed fix** — return `'<br />'` for both break types ahead of the shared content pass-through:

```ts
if (type === 'softbreak' || type === 'hardbreak') {
  return '<br />';
}
```

Not `'<br />\n'` — the trailing newline adds a stray whitespace text node, which the fifth criterion above pins. The merged node keeps the opening tag's `type: 'html_inline'`, because the merge spreads `...startNode.token` and overrides only `content`. So it still routes through `sanitizeHtmlContent` then `unsafeHTML`.

### Out of scope

- **The main-path break fix** — #2222, closing #2220. Landed or landing separately; this issue assumes it and does not re-prove it.
- **`InlineMarkdown` disagreeing with the block renderer on soft breaks** — #2272.
- **The fire-once plugin-fallback mount handshake** — #2271.
- **Consolidating the container half of the plugin-host protocol** — #2273.
- **First-time test coverage for `conversational_search`**, which has none in either package. A real gap, and not this issue's job.

### Related

- #2220 / #2222 — the main-path fix. Independent of this one: that change is in `renderTokenTree`, this one is in the inline-HTML serializer, and neither blocks the other. If #2222 absorbs this fix, close this issue with it.
- #2272 — `InlineMarkdown` disagrees with the block renderer on soft breaks.
- #2271 — plugin-fallback mount handshake is fire-once with no replay.
- #2273 — consolidate the plugin-host container protocol behind one shared helper.

### LINKED ISSUE #2271: Make the plugin-fallback host handshake survive a late subscriber
### Background

Markdown that a plugin produces — a KaTeX formula, a footnote, a custom container — can end up rendering nowhere. No error, no placeholder, no fallback. The content is simply missing.

That output does not live in the markdown component. `<cds-aichat-markdown>` hands anything it cannot render natively to the chat container, which builds the host element in the page's own DOM. That way a stylesheet the consumer loaded on the page, such as a KaTeX stylesheet, can reach it. The handoff runs in four steps:

1. The renderer emits an empty `<slot name=X>` into the markdown shadow root.
2. The markdown element fires a `cds-aichat-markdown-plugin-host-mount` event.
3. The chat container answers with `preventDefault()` and builds the host in its own DOM.
4. The React wrapper renders a forwarder that projects that host back through the shadow boundaries into the slot.

The protocol has three events, split by how the content changes hands. Plugin fallback passes an HTML **string**, so later chunks need a second event, `plugin-host-update`, to reach the host. `customRenderers` passes a **live element** that the markdown element keeps mutating, so it needs no update at all. `plugin-host-unmount` tears a host down.

Step 2 fires **once per slot name**, and never again. Later passes fire `plugin-host-update`, which the wrapper does not listen for. Nothing replays the mount.

Two listeners have to hear that single event, and they start listening at different times:

- The **container** subscribes when the chat mounts, so it is effectively always there. It decides step 3.
- The **React wrapper** subscribes in a `useEffect` that runs after the markdown element's first render. It decides step 4.

A slot created in the gap between the two is stranded for good. Step 3 runs and step 4 never does. The element also skips its own fallback host, because the cancelled event told it the container took over. The content now has a home on the page and no route back into the slot.

The two callers of this path fail differently. A `customRenderers` host (`table` or `codeBlock`) degrades quietly but visibly: its slot carries fallback content, so a lost host leaves the default Carbon table or code snippet on screen. Plugin fallback emits its slot empty, so a lost host shows nothing at all.

This surfaced while diagnosing #2270, where every `<br>` traveled this same path and line breaks vanished depending on where the message sat. Taking breaks off the path is #2220, in review as #2222. Once that lands, the exposure drops to consumers who use markdown-it plugins that introduce unknown tokens, or who set `customRenderers`. Until it lands, every line break in every streamed message still travels this path.

**Instrumentation never reproduced the interleaving.** A one-shot event with two independently timed subscribers is the defect, whatever the odds of hitting it today.

### Goal

Plugin and custom-renderer output reaches its slot no matter when each listener subscribed, and a host that nothing claims fails visibly instead of silently.

### Acceptance criteria

- [ ] A markdown element whose wrapper subscribes *after* the mount event still renders the plugin output. Proof: a new case in `packages/ai-chat-components/src/components/markdown/__tests__/markdown.test.ts` that delays the forwarder subscription past the first reconcile.
- [ ] The same holds for a `customRenderers` host, `table` or `codeBlock`. Proof: same spec.
- [ ] Repeated reconciles still produce exactly one host and one forwarder per slot name. Proof: same spec, asserting both counts after a run of streaming ticks.
- [ ] The existing plugin and custom-renderer cases pass unchanged, including the per-element slot names from #2119. Proof: the existing `describe('slot names across markdown elements')` block.
- [ ] The three chat containers behave identically on a mount, an update, and an unmount. Proof: drive each with the same event sequence and compare the resulting host DOM. If the shared helper from #2273 has landed first, this reduces to that helper's own spec.
- [ ] Gate: `npm run build --workspace=@carbon/ai-chat-components` and `npm run test --workspace=@carbon/ai-chat-components` exit 0, plus `npm run lint` and `npm run lint:license`. If the containers in `@carbon/ai-chat` change too, the gate becomes `npm run build` and `npm run ci-check`. Gate rows and their ordering rule: [definition-of-done.md](https://github.com/carbon-design-system/carbon-ai-chat/blob/main/references/definition-of-done.md).

### Public API / contract

No signature changes expected. The contract here is the event protocol between the markdown element and its hosting chat container, which anyone writing their own container can observe.

| Lock | Answer |
| --- | --- |
| Preconditions | Applies when `markdownItPlugins` introduce a token the renderer does not own, or when `customRenderers` is set. |
| No-op case | Markdown with no plugin tokens and no custom renderers dispatches nothing, as today. |
| Failure | An unclaimed host has to surface rather than disappear. How is open: leave the element's own host in place, or warn in dev mode. |
| Events | Unchanged. Mount still fires once per live host, update still carries later chunks, unmount still tears down. The chosen approach adds no event and re-fires none. |
| Timing | Projection must not depend on which listener subscribed first. That is the whole issue. |
| Repeat calls | Idempotent per slot name. No duplicate hosts, no duplicate forwarders. |
| Defaults | Unchanged. |
| Derivation | N/A. |
| Announcement | Unchanged, except that content which used to vanish now reaches the accessibility tree. |
| Ownership | Framework, on both sides of the handshake. |

**Approach.** The React wrapper seeds its slot names from a getter on the markdown element, instead of relying on having caught the event. This touches the element and the wrapper only. It leans on no container behavior, and it leaves the event contract untouched — which matters, because `@carbon/ai-chat-components` ships on its own and anyone can implement the hosting half against these events.

Two alternatives were rejected. Re-dispatching mount when a listener subscribes rests on every host being idempotent per slot name; the three in this repo are, and nothing enforces it. Dispatching on every reconcile changes what every host observes, and no one can check a third-party host for idempotence.

The fix has to cover the `customRenderers` path as well as plugin fallback.

### Out of scope

- **Taking breaks off this path** — #2220, in review as #2222. That lands first and independently, and it is what makes this a follow-up rather than a release blocker.
- **Reworking how plugin output is styled or where hosts live.** Page light DOM is deliberate, so consumer stylesheets reach plugin output.
- **Test coverage for `plugin-host-update`.** No test in this repo listens to that event. A real gap, but the chosen approach does not touch that path.

### Related

Depends on: #2220 (its fix rewrites the wrapper mount handler this builds on).

- #2220 — breaks not working in markdown. Its fix, #2222, handles `softbreak` and `hardbreak` natively so breaks stop minting slots here. It also rewrites the wrapper's mount handler that a fix here has to build on.
- #2270 — a line break inside a merged inline-HTML run is serialized away. A separate defect, in a different file.
- #2273 — consolidate the plugin-host container protocol behind one shared helper, replacing three near-verbatim copies plus a test mock. Not a blocker in either direction: the chosen approach is element- and wrapper-side. It mattered to the two rejected alternatives, which rest on every host staying idempotent per slot name — true today by coincidence of three separate implementations, not by construction.
- #2119 — namespaced custom-renderer slot names per element. The prior fix in this machinery.

### LINKED ISSUE #2272: Render a soft break as a line break in InlineMarkdown
### Background

`@carbon/ai-chat` renders markdown two ways. The block renderer is `<cds-aichat-markdown>` from `@carbon/ai-chat-components`, whose parser is configured `breaks: true` — a single `\n` inside a paragraph is a line break. The other is `renderInlineMarkdown` (`packages/ai-chat/src/chat/components/helpers/InlineMarkdown/InlineMarkdown.tsx`), a small React token walker used for user message bubbles that carry chips.

The two disagree on soft breaks. `InlineMarkdown` renders `hardbreak` as `<br />` but `softbreak` as a space:

```tsx
case 'softbreak':
  pushChild(<Fragment key={nextKey()}> </Fragment>);
  break;

case 'hardbreak':
  pushChild(<br key={nextKey()} />);
  break;
```

That contradicts the `breaks: true` config of the shared parser it calls tokens out of — under that config a `softbreak` *is* a line break, which is exactly why the block renderer emits `<br>` for it.

It shows up in one component, on one keystroke. `MessageRichUserContent` encodes a user's <kbd>Shift</kbd>+<kbd>Enter</kbd> as a single `'\n'` (`joinTextualInline`, `packages/ai-chat/src/chat/components-legacy/MessageRichUserContent.tsx:138-147`), then routes the paragraph one of two ways depending on whether it contains chips:

- no chips → `<MarkdownWithDefaults text={flat} …>` (line 65) — the block renderer, which renders a `<br>`;
- chips → `renderInlineMarkdown(joined)` (line 89) — a space.

So the same message typed the same way renders as two lines or one, decided by whether the user happened to mention someone in it.

Found while diagnosing #2270. The fix that makes the block renderer's `<br>` reliable rather than position-dependent is #2220, in review as #2222. That sharpens this divergence into something consistently reproducible — before it, both paths were unreliable in different ways, so the disagreement was masked.

### Goal

A soft break renders as a line break everywhere. The same user message renders identically whether or not it contains chips.

### Acceptance criteria

- [ ] `renderInlineMarkdown('one\ntwo')` emits one `<br>` and text content `onetwo`. Proof: new case in `packages/ai-chat/tests/utils/spec/inlineMarkdown_spec.tsx`.
- [ ] `renderInlineMarkdown('one  \ntwo')` — a two-space hard break — still emits one `<br>`, unchanged. Proof: same spec.
- [ ] A user message containing a single <kbd>Shift</kbd>+<kbd>Enter</kbd> renders the same number of visual lines with and without a chip in it. Proof: new case in `packages/ai-chat/tests/components/spec/messageRichUserContentChipSpacing_spec.tsx`, which already exercises the chips/no-chips fork; both branches given the same input text. Consecutive breaks are out of scope — see the amendment below.
- [ ] Existing `renderInlineMarkdown` cases — bold, em, inline code, strike, link, HTML stripping — pass unchanged. Proof: existing cases in `inlineMarkdown_spec.tsx`.
- [ ] `npm run build` then `npm run ci-check` exit 0. This is a cross-package change: the assertion spans `@carbon/ai-chat` behavior that depends on `@carbon/ai-chat-components` output, and workspace deps resolve through built artifacts. Gate rows, their ordering rule and the watcher precondition: [definition-of-done.md](https://github.com/carbon-design-system/carbon-ai-chat/blob/main/references/definition-of-done.md).

### Amendment — 2026-09-08

Criterion 3 is narrowed to a **single** <kbd>Shift</kbd>+<kbd>Enter</kbd>. Full chip/no-chip parity is not reachable from this issue.

Two consecutive breaks produce `\n\n`, which markdown-it parses as two blocks. `renderInlineMarkdown` reads only the first `inline` token (`InlineMarkdown.tsx:44`), so the chip path drops everything after the blank line while the chip-free path renders both paragraphs. That is the same divergence this issue exists to close, but the cause is the walker's block handling rather than its break handling, and the fix is the refactor in #1873 — which deletes `renderInlineMarkdown` outright. Evidence and measurements are on #1873.

Every other criterion stands and is delivered by #2299.

### Public API / contract

No signature changes. Behavioral, on rendered user-message content.

| Lock | Answer |
| --- | --- |
| Preconditions | None. Applies wherever `renderInlineMarkdown` is called. |
| No-op case | Text with no soft break renders exactly as today. |
| Failure | Never throws. |
| Events | None. |
| Timing | Synchronous, within the React render. |
| Repeat calls | Pure function of its input. |
| Defaults | Unchanged. |
| Derivation | N/A — no state or enum a host reads. |
| Announcement | A soft break becomes a real line break in the accessibility tree instead of a space, matching what the user typed. Focus unaffected. |
| Ownership | Framework. |

**Proposed fix** — collapse the two cases in `InlineMarkdown.tsx`:

```tsx
case 'softbreak':
case 'hardbreak':
  pushChild(<br key={nextKey()} />);
  break;
```

One line of behavior change, but it touches `@carbon/ai-chat` and is verified against `@carbon/ai-chat-components` output, so it carries a cross-package gate and belongs in its own PR rather than riding along with #2270.

### Out of scope

- **The block renderer's handling of soft breaks** — #2220, in review as #2222. This issue assumes that fix has landed and makes the inline walker agree with it.
- **Consecutive soft breaks, and multi-block input generally.** The walker renders only the first block. Tracked on #1873, which replaces `renderInlineMarkdown` with an inline mode on the element. See the amendment above.
- Wider reconciliation of the two markdown paths. `renderInlineMarkdown` deliberately handles a small inline subset; only the break disagreement is in scope here.

### Related

- #2220 — breaks not working in markdown, in review as #2222. Blocks this in practice: the "renders identically with and without chips" criterion is only verifiable once the block renderer's `<br>` is deterministic.
- #2270 — a line break inside a merged inline-HTML run is serialized away. A separate defect, in a different file.

### LINKED ISSUE #2273: Consolidate the plugin-host container protocol behind one shared helper
### Background

`<cds-aichat-markdown>` offers plugin-fallback and custom-renderer hosts to an ancestor chat container over three composed events — `cds-aichat-markdown-plugin-host-mount`, `-update`, and `-unmount`. The container half of that protocol is implemented **three times, independently, with no shared code**:

- `packages/ai-chat/src/react/ChatContainer.tsx`
- `packages/ai-chat/src/web-components/cds-aichat-container/index.ts`
- `packages/ai-chat/src/web-components/cds-aichat-custom-element/index.ts`

The duplication is near-verbatim. `ChatContainer`'s `handleUpdate` and `cds-aichat-container`'s `handlePluginHostUpdate` are line-for-line identical. The mount handlers differ only in where they append the host (`wrapper.appendChild` vs `this.appendChild`) and where they park it (a closure `Map` vs `this._pluginHosts`) — the branching on `detail.element` vs `detail.html`, the inline-vs-block tag choice, the `marginBlockStart = '1rem'` spacing patch, and the `innerHTML` diffing are copied.

A fourth copy lives in tests: `createRelocationHarness()` in `packages/ai-chat-components/src/components/markdown/__tests__/markdown.test.ts`, whose own docstring says it "Mirrors `ChatContainer.tsx` and `cds-aichat-container`." Nothing enforces the mirror.

Nothing is broken today — the copies currently agree. The cost is structural:

- Every protocol change is a four-site change, and the fourth site is a test mock that can silently stop resembling the thing it mocks.
- A divergence between containers produces content that renders under one and vanishes under another, with no error — the same silent-failure mode as #2271, since a plugin-fallback host is emitted into an empty `<slot>`.
- A third party writing its own container has no reference implementation, only three examples to reverse-engineer.

There is a second half to the same duplication. Each copy decides which of two payload shapes it received by testing whether `detail.element` is set — the shape is the discriminant. Consolidating the handlers without fixing that would bake the shape test into the shared helper, so this issue publishes a real discriminant first: a `kind` field on the mount detail, and the exported union behind it.

Found while diagnosing #2270 and scoping #2271.

The dependency between the packages runs one way — `@carbon/ai-chat` depends on `@carbon/ai-chat-components`, never the reverse — so the shared helper belongs in `@carbon/ai-chat-components`, the package that already defines the element half of the protocol. That is the only placement reachable by all three containers **and** the test harness.

### Goal

One module owns the container half of the plugin-host protocol. All three chat containers and the test harness consume it, so a protocol change is a one-site change and divergence between containers is not expressible.

### Acceptance criteria

- [ ] Given one mount event, all three containers produce the same host DOM — same tag for the inline and block cases, same `slot` attribute, same spacing. Proof: new spec driving each container with an identical event and comparing the resulting host. Nothing enforces this today.
- [ ] A custom-renderer mount (carrying a live `element`) and a plugin-fallback mount (carrying an `html` string) are each handled identically across all three. Proof: same spec, both detail shapes.
- [ ] Update and unmount agree across all three: an update rewrites host content in place, an unmount removes the host and drops its slot name. Proof: same spec.
- [ ] `createRelocationHarness()` consumes the shared helper instead of reimplementing it. Proof: the existing cases in `describe('slot names across markdown elements')` pass unchanged.
- [ ] No behavior change anywhere. Proof: the existing plugin, `customRenderers`, and relocation cases in `markdown.test.ts` pass **unchanged** — no assertion loosened, no case skipped, no snapshot regenerated.
- [ ] Hosting is covered by a test that loads the real handlers. Proof: a new spec under `packages/ai-chat/tests/` drives each of the three containers with one mount event of each shape and asserts the resulting host DOM — tag, `slot` attribute, spacing, content. It fails when a handler's branch is removed. `grep -rn "plugin-host" packages/ai-chat/tests` is empty today, so this spec does not exist yet.
- [ ] The mount detail carries a `kind` discriminant — `'pluginFallback'` where the detail carries an HTML string, `'customRenderer'` where it carries a live element — and the discriminated union behind it is exported from `packages/ai-chat-components/src/components/markdown/index.ts`. Proof: a fixture importing the union from the package entry, not by deep import, compiles under `npx tsc --noEmit`; and new cases in `markdown.test.ts` assert the field on both dispatch paths. Neither dispatch site sets it today, so both cases fail before the change.
- [ ] The shared helper narrows on `kind`, not on payload shape. Proof: no `if (detail.element)` branch remains in `cds-aichat-custom-element/index.ts`, `cds-aichat-container/index.ts` or `ChatContainer.tsx`, and the helper switches on `detail.kind`.
- [ ] A wrong branch fails to compile. Proof: read `detail.html` inside a `kind === 'customRenderer'` branch of the helper; `npm run build --workspace=@carbon/ai-chat` fails type-checking. Revert.
- [ ] The six `?? ''` fallbacks the shape test needed are gone — two per handler. Proof: `grep -rn "detail.html ?? ''" packages/ai-chat/src` returns nothing.
- [ ] A mount detail with no `kind` still hosts. Proof: the same spec dispatches the legacy shape — `{ slotName, html, isInline }`, no `kind` — at each container and asserts the host is created and carries the HTML. The components dependency is a caret range, so an older build can still emit the pre-`kind` shape.
- [ ] `cds-aichat-container` still declines the offer when an outer chat element is present. Proof: the spec nests a container inside `cds-aichat-custom-element`; the container's handler returns without `preventDefault()` and appends nothing.
- [ ] `npm run build` then `npm run ci-check` exit 0. Cross-package: the helper lands in `@carbon/ai-chat-components` and all three consumers live in `@carbon/ai-chat`. Gate rows, their ordering rule and the watcher precondition: [definition-of-done.md](https://github.com/carbon-design-system/carbon-ai-chat/blob/main/references/definition-of-done.md).
- [ ] Ensure existing tests run smoothly/not broken

### Public API / contract

The three events are already consumer-observable — anyone implementing their own chat container listens for them. This issue does not change their names, order, or payloads; that is the point of it being a refactor.

Whichever chat element is outermost claims the mount event. `cds-aichat-container` defers when it detects an outer `cds-aichat-custom-element` or `cds-aichat-react` in the composed path, and the outer element claims instead. The event still bubbles to the page — nothing calls `stopPropagation()` — but it arrives already `defaultPrevented`.

The mount detail gains a `kind` discriminant, and the discriminated union behind it is newly exported from `@carbon/ai-chat-components`. Both are additive: every listener reads named properties off a bare `Event` cast, so a new field cannot break one. The union is not re-exported from `@carbon/ai-chat`: a consumer never names the type, and that package re-declares a cross-package type by hand with its own docs, which this does not earn.

One question it does have to settle:

| Lock | Answer |
| --- | --- |
| Ownership | Framework. The helper owns host creation, relocation, spacing, content diffing, and teardown; the container owns only where hosts are appended. |
| Preconditions | None beyond today's — the container must have an element to append to. |
| No-op case | A mount event with no `slotName` is ignored, as today in all three copies. |
| Failure | Never throws. Unchanged. |
| Events | Unchanged — same three events, same payloads, same order. |
| Timing | Unchanged, synchronous in the listener. |
| Repeat calls | Idempotent per slot name — a repeated mount reuses the existing host, as today. |
| Defaults | Unchanged, including the `marginBlockStart: 1rem` block-host spacing patch. |
| Derivation | N/A. |
| Announcement | Unchanged. |
| Exported or internal? | **Open — settle before implementation.** See below. |

**Open question: does the helper ship as public API?** Exporting it from `@carbon/ai-chat-components` gives anyone writing their own chat container a correct implementation instead of three examples to reverse-engineer — which is the situation that produced this issue. It also makes the protocol public surface, with the semver weight and JSDoc obligations that carry. Keeping it `@internal` consolidates the code without taking on that commitment, and leaves third-party container authors exactly where they are today. Deliberately unresolved here; whoever picks this up settles it first, and it does not change the acceptance criteria either way.

### Out of scope

- **Changing the protocol's behavior** — #2271. This lands as a pure refactor with existing tests green and unmodified; the fire-once repair rides on that issue. Order between the two is a judgment call, not a hard dependency — see Related.
- **`reconcileCustomRendererHosts`** (`packages/ai-chat-components/src/components/markdown/src/markdown.ts`), 165 lines doing four jobs. Real, but one method in one place with no divergence risk, so it fails differently from this and is not the same fix.
- **Removing `<br>` from the fallback path** — #2220, in review as #2222.
- **The `-update` and `-unmount` detail types**, and the mount listener in the `@carbon/ai-chat-components` React Markdown wrapper. They keep their inline detail types.
- **Documenting the protocol's events in `custom-elements.json`** — #2281. This issue exports the mount detail union; the `@fires` tags that put it in the manifest, and the guard that enforces them, stay there.

### Related

- #2281 — gates `custom-elements.json` against the events elements actually dispatch. **Not a blocker in either direction.** It originally owned the `kind` discriminant, which moved here: this issue's gate is already a full monorepo build, so a cross-package type change costs nothing extra here, while #2281 kept its gate to a components-only build on purpose. #2281 documents the union it finds rather than inventing one, and the two can land in either order.
- #2271 — the fire-once handshake repair. Not a blocker in either direction: its candidate approaches are element- and wrapper-side, and all three containers already dedupe by slot name. This issue matters to it because two of those approaches rest on every container being idempotent per slot name — true today across three separate implementations by coincidence rather than by construction, and unenforced. Landing this first makes that property structural instead of incidental.
- #2270 — a line break inside a merged inline-HTML run is serialized away; where the duplication was found.
- #2119 — namespaced custom-renderer slot names per element; the prior fix in this machinery, applied to the element side only.

ethanwinters:
Closes #2270
Closes #2272
Closes #2271
Closes #2273

Five changes to the line-break and plugin-host story. One per commit, each green and revertable alone.

The first two are two-line fixes with their proofs. Commits 3-5 are the ones to read closely:

- `packages/ai-chat-components/src/components/markdown/src/markdown.ts` — `delegatedPluginSlots` becomes a kinded `Map`, and a new public `delegatedPluginSlotNames` getter reads only one of the two kinds. The kind split is the load-bearing part; see the getter's JSDoc for why.
- `packages/ai-chat-components/src/react/markdown.tsx` — six lines inside the existing subscribe effect. Order matters: attach both listeners, *then* seed.

#### Commit map

1. **`fix(markdown): render line breaks inside merged inline-HTML runs`** — `serializeInlineToken` returned `token.content ?? ''` for `softbreak` and `hardbreak`. Break tokens carry no content. So a break inside a merged inline-HTML run — `<span>one\ntwo</span>` — was deleted, not rendered. Returns `'<br />'` now, not `'<br />\n'`; the trailing newline adds a stray whitespace text node that a spec pins.
2. **`fix(markdown): treat a soft break as a line break in InlineMarkdown`** — `renderInlineMarkdown` rendered `softbreak` as a space. `hardbreak` already produced `<br />`. Under `breaks: true` a soft break *is* a line break, so the two paths disagreed. The same message rendered as one line or two, depending only on whether it held a chip. The `softbreak` case now falls through to `hardbreak`.
3. **`fix(markdown): seed plugin slot forwarders from the element`** — the late-subscriber fix. Detail below.
4. **`refactor(markdown): consolidate the plugin-host container protocol`** — the container half of the protocol was written five times. One helper now, plus a real discriminant on the mount detail. Detail below.
5. **`fix(markdown): mint a plugin-fallback slot name once per token`** — the defect commit 3 turned up, root-caused and fixed. Detail below.

#### The late-subscriber fix

`cds-aichat-markdown-plugin-host-mount` fires once per live host and is never replayed for one — later passes fire `plugin-host-update` instead. Two listeners depend on that single event and start listening at different times:

| Listener | Attaches | Job |
| --- | --- | --- |
| Chat container | chat mount | `preventDefault()`, hoist the host into page light DOM |
| React `Markdown` wrapper | the element's first passive-effect flush | render the `<slot name=X slot=X>` forwarder |

That forwarder is the only thing crossing `<cds-aichat-markdown>`'s own shadow boundary. That holds on **every** container topology; the web-component containers supply outer hops only.

Miss the one event and the slot is stranded for good. The container owns the host. The element skips its own local fallback. No forwarder ever appears, so plugin-fallback output shows nothing at all.

The wrapper now seeds from a getter on the element instead of relying on having caught the event. No event added, none re-fired, no container changed.

#### The consolidation (#2273)

The container half of the plugin-host protocol existed five times: `ChatContainer.tsx`, `cds-aichat-container`, `cds-aichat-custom-element`, and two harnesses in `markdown.test.ts` whose docstrings claim to mirror them. `handleUpdate` was line-for-line identical in all three containers. Nothing enforced the mirror, and a divergence between containers renders content under one surface and silently drops it under another — a plugin-fallback host is emitted into an empty `<slot>`, so a lost host looks like nothing at all.

`createMarkdownPluginHostController` in `@carbon/ai-chat-components` is the single implementation. A caller supplies the element hosts are appended to, plus the two policies the framework cannot infer: where forwarded slot names go, and whether an outer chat element will host instead. Net −430 lines across the three containers.

Each copy also decided which of two payloads it had received by testing whether `detail.element` was set — the shape was the discriminant. The mount detail now carries `kind`, and the union behind it is exported. The union members declare only their own fields, so reading `detail.html` inside a `customRenderer` branch is `TS2339` rather than a silent `undefined`. Details from before the field existed still host: one shim fills `kind` in at the protocol boundary, and everything downstream switches on it.

**One deliberate behavior change.** Both web-component containers tracked `customRenderer` slot names and rendered an outer forwarder for them; the React wrapper and `ChatContainer` did not. Those forwarders are inert — a `customRenderer` host is never delegated, so the element hosts it in its own light DOM, the outer forwarder gathers nothing, and its `slot=` half terminates against a `<slot name=X>` the wrapper declines to render. The controller keeps the wrapper's rule. That closes a hazard for third-party containers, where a claimed live-element mount plus a forwarder would hold the named slot occupied and suppress the fallback a `null`-returning callback depends on. It also changes the observable contents of `_pluginSlotNames`, which on `cds-aichat-container` is not declared `private`.

#### The slot-counter fix (commit 5)

Found while testing commit 3; root cause is not what it first looked like.

`renderFallback` does not run while `renderMarkdownTree` walks the tree. It runs inside a lit-html `repeat()` directive callback, so it evaluates when lit-html **commits** the template. `render()` returns the stored `renderedContent`, and any later Lit update inside the render throttle re-commits that same `TemplateResult`, re-running the repeat callback against the same counter closure. One render result committed twice advanced the counter twice and pushed two descriptors for one token, so a delegating container built and tore down a spurious host per commit.

Two corrections to how this was first described: it is **not** streaming-specific — a bare `requestUpdate()` on a settled element reproduces it, and a stream only hits it often because every chunk is another commit — and `table`/`codeBlock` names were never affected, because `slotNameFor` derives those from `token.map` and the parent's `currentIndex`, which is idempotent under re-commit.

The index is now memoized per tree node rather than counting evaluations. First evaluation still assigns 0, 1, 2 in walk order, so a render committed once is byte-identical to before.

#### Changelog

**New**

- `CDSAIChatMarkdown.delegatedPluginSlotNames` — read-only getter. Lists the plugin-fallback slot names an outer listener has claimed. For anyone implementing the hosting half of the protocol: a late subscriber can recover the names it missed.
- `MarkdownPluginHostMountDetail` and `MarkdownPluginHostMountDetailInput`, exported from `@carbon/ai-chat-components/es/components/markdown/index.js`. The `cds-aichat-markdown-plugin-host-mount` detail, discriminated on a new `kind` field, plus the shape as it arrives on the wire from a build predating that field. Additive: every listener reads named properties off a bare `Event` cast, so a new field cannot break one.

**Changed**

- A single `\n` inside inline HTML renders as `<br>` instead of vanishing.
- `InlineMarkdown` renders a soft break as `<br />`, matching the block renderer. A message with a chip and one without now render the same.
- Plugin-fallback output reaches its slot regardless of when the forwarder's listener subscribed.
- `cds-aichat-container` and `cds-aichat-custom-element` no longer render an outer `<slot>` forwarder for a `customRenderers` slot name. Those forwarders were inert; see above.
- A plugin-fallback token produces one slot host for as long as it exists, rather than one per lit-html commit. Anyone counting `cds-aichat-markdown-plugin-host-mount` events sees fewer; anyone keyed by slot name sees no difference, because the extra host was always retired immediately.

**Removed**

- A plugin overriding `md.renderer.rules.softbreak` no longer takes effect. `softbreak` was never in `PLUGIN_DELEGABLE_TOKEN_TYPES`, so it only ever reached output through the unknown-token fallback — never a supported extension point.

#### Testing / Reviewing

Build first — workspace deps resolve through built artifacts:

```bash
```

Automated:

```bash
npm run test --workspace=@carbon/ai-chat-components   # 581 WTR × 3 browsers, 24 Jest
npm run test --workspace=@carbon/ai-chat
npm run lint && npm run lint:license
```

Manual — commit 1. Paste into the Storybook markdown story: `<span>**Summary**\nUse the Assets application.</span>`. Before, the newline disappears and the words merge; after, the lines are separate.

Manual — commit 2. Type a message with Shift+Enter, then mention someone so the paragraph routes through the chip path. Before, the break collapses to a space; after, it matches the same text without the chip.

Commit 3 is covered by `describe('plugin-host handshake with a late subscriber')` in `markdown.test.ts`, plus the new React-wrapper spec. Reproducing it by hand is awkward. You need a `markdownItPlugins` plugin emitting an unknown token, inside a chat container, on the first paint. The tests model that split directly.

Commit 4 is covered by `packages/ai-chat/tests/web-components/spec/pluginHostParity_spec.tsx` — the first test anywhere to load these handlers, since `grep -rn "plugin-host" packages/ai-chat/tests` was empty. It drives all three containers with one event sequence and compares the resulting host DOM. Ten of its eleven cases were written against the old handlers and pass unchanged; the forwarder case failed for both web-component containers before the change and passes after. Deleting a branch from any one handler turns the spec red — worth re-running that yourself if you want to trust it.

Commit 5 is covered by `describe('plugin-fallback slot names across render commits')`. All four cases fail before it in all three browsers.

#### Reviewer notes

Three things worth knowing, none of them blocking:

- **Criterion 2 on #2271 was re-aimed.** It asserted a *timing* failure for `customRenderers` hosts. #2222 removed that timing. All three containers and the wrapper now ignore mount events carrying `detail.element`, so such a host is never delegated. Re-aimed at the live question instead: does the seed leak a `customRenderers` slot name into the forwarder list? It doesn't, and a test fails if that changes.
- **Criterion 5 on #2271 is descoped to #2273.** Three-container parity is #2273's whole purpose. This commit changes no container. That leaves the invariant unproven until #2273 lands; recorded on both issues.
- **Criterion 5 on #2271 is met here, one commit later than planned.** Commit 3 descoped three-container parity to #2273 because it changed no container. `pluginHostParity_spec.tsx` is that proof.
- **Four of #2273's criteria needed correcting**, all recorded on the issue: its "package entry" does not exist (`@carbon/ai-chat-components` publishes no `"."` export, so the markdown barrel is the entry); its wrong-branch compile check names `@carbon/ai-chat`, but the helper lives in the components package, so the error surfaces in that build; and its "an unmount drops its slot name" half is provable across two containers, not three, because `ChatContainer` keeps no slot-name list and renders no forwarder.
- **The refactor is behavior-preserving except where it says otherwise.** The `customRenderer` forwarder change is the one exception and it has its own failing-first test. Everything else is pinned by cases written against the old handlers and left untouched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/ai-chat-components/src/components/markdown/__tests__/markdown.test.ts
+++ b/packages/ai-chat-components/src/components/markdown/__tests__/markdown.test.ts
@@ -9,6 +9,11 @@
 
 import { expect, fixture, html, waitUntil } from '@open-wc/testing';
 import CDSAIChatMarkdownElement from '../src/markdown.js';
+import {
+  createMarkdownPluginHostController,
+  resolveMarkdownPluginHostMountDetail,
+} from '../src/utils/plugin-host-container.js';
+import type { MarkdownPluginHostMountDetailInput } from '../src/utils/plugin-host-container.js';
 const MARKDOWN_ELEMENT_TAG = 'cds-aichat-markdown';
 
 const TEXT = `Carbon <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 32 32" onclick="window.open('https://carbondesignsystem.com', '_blank')"><defs><style>.cls-1{fill:none;}</style></defs><title>If you click on this icon, it will go to https://carbondesignsystem.com. This is here to test "shouldSanitizeHTML". If true, the click shouldn't work!</title><path d="M13.5,30.8149a1.0011,1.0011,0,0,1-.4927-.13l-8.5-4.815A1,1,0,0,1,4,25V15a1,1,0,0,1,.5073-.87l8.5-4.815a1.0013,1.0013,0,0,1,.9854,0l8.5,4.815A1,1,0,0,1,23,15V25a1,1,0,0,1-.5073.87l-8.5,4.815A1.0011,1.0011,0,0,1,13.5,30.8149ZM6,24.417l7.5,4.2485L21,24.417V15.583l-7.5-4.2485L6,15.583Z"/><path d="M28,17H26V7.583L18.5,3.3345,10.4927,7.87,9.5073,6.13l8.5-4.815a1.0013,1.0013,0,0,1,.9854,0l8.5,4.815A1,1,0,0,1,28,7Z"/><rect class="cls-1" width="32" height="32" transform="translate(32 32) rotate(180)"/></svg> is a **chemical element** with the *atomic number* 6 and symbol **C**. \`C + O₂ → CO₂\` represents one of carbon's most fundamental reactions.
@@ -1539,11 +1544,15 @@ HTTP: http://example.com
     }
 
     /**
-     * Stands in for a chat container: accepts every host-mount offer, relocates
-     * the host into its own light DOM, and renders one `<slot name=X slot=X>`
-     * forwarder per slot name — deduplicated by name, exactly as the real
-     * containers dedupe `_pluginSlotNames`. Mirrors `ChatContainer.tsx` and
-     * `cds-aichat-container`.
+     * A whole delegating topology in one object: the shipped container
+     * controller does the hosting, and a second listener adds the React
+     * `Markdown` wrapper's forwarder on top. It cannot drift from the
+     * containers, because it runs their code.
+     *
+     * The two halves belong to different layers. Hosting is a container's job;
+     * the `<slot name=X slot=X>` forwarder into the markdown element's own
+     * light DOM is the wrapper's, and is the only hop that crosses that
+     * element's shadow boundary.
      */
     async function createRelocationHarness() {
       const harness = await fixture<HTMLElement>(
@@ -1555,47 +1564,37 @@ HTTP: http://example.com
         host: HTMLElement;
       }> = [];
       const forwarderOwners = new Map<string, Element>();
-      const pluginHosts = new Map<string, HTMLElement>();
 
+      // The container half is the shipped one, so this harness cannot drift
+      // from the surfaces it stands in for.
+      const controller = createMarkdownPluginHostController(harness);
+      controller.connect();
+
+      // The forwarder half, plus bookkeeping. Registered after the controller
+      // so a plugin-fallback host already exists by the time this runs. This
+      // is the React `Markdown` wrapper's job, not a container's — no chat
+      // container renders a forwarder into the markdown element's own light
+      // DOM, and that hop is the one that crosses its shadow boundary.
       harness.addEventListener(
         'cds-aichat-markdown-plugin-host-mount',
         (event) => {
-          const detail = (
-            event as CustomEvent<{
-              slotName: string;
-              html?: string;
-              element?: HTMLElement;
-              isInline: boolean;
-            }>
-          ).detail;
+          const detail = resolveMarkdownPluginHostMountDetail(
+            (event as CustomEvent<MarkdownPluginHostMountDetailInput>).detail
+          );
           const owner = event.composedPath()[0] as Element;
 
-          // Custom-renderer hosts (table/codeBlock) carry a live element that
-          // the markdown element manages directly. Do not preventDefault or
-          // hoist — the markdown element will call appendChild itself, keeping
-          // the shadow slot occupied. A stale forwarder or a missing host
-          // causes the slot to fall back to the default component, which
-          // triggers heavy async work (e.g. CodeMirror) that hangs updateComplete.
-          if (detail.element) {
+          if (detail.kind === 'customRenderer') {
+            // A live element the markdown element manages itself. Never
+            // hoisted, never forwarded — a stale forwarder or a missing host
+            // makes the slot fall back to the default component, which
+            // triggers heavy async work (e.g. CodeMirror) that hangs
+            // updateComplete.
             mounted.push({
               owner,
               slotName: detail.slotName,
               host: detail.element,
             });
             return;
-          }
-
-          event.preventDefault();
-
-          // Plugin fallbacks forward an HTML string; the container hosts it.
-          const host =
-            pluginHosts.get(detail.slotName) ??
-            document.createElement(detail.isInline ? 'span' : 'div');
-          host.innerHTML = detail.html ?? '';
-          pluginHosts.set(detail.slotName, host);
-          host.setAttribute('slot', detail.slotName);
-          if (host.parentElement !== harness) {
-            harness.appendChild(host);
           }
 
           if (!forwarderOwners.has(detail.slotName)) {
@@ -1613,7 +1612,10 @@ HTTP: http://example.com
             forwarder.setAttribute('slot', detail.slotName);
             owner.appendChild(forwarder);
           }
-          mounted.push({ owner, slotName: detail.slotName, host });
+          const host = controller.hosts.get(detail.slotName);
+          if (host) {
+            mounted.push({ owner, slotName: detail.slotName, host });
+          }
         }
       );
       harness.addEventListener(
@@ -1622,8 +1624,6 @@ HTTP: http://example.com
           const { slotName } = (event as CustomEvent<{ slotName: string }>)
             .detail;
           forwarderOwners.delete(slotName);
-          pluginHosts.get(slotName)?.remove();
-          pluginHosts.delete(slotName);
         }
       );
 
@@ -1641,7 +1641,13 @@ HTTP: http://example.com
         return el;
       }
 
-      return { harness, mounted, forwarderOwners, pluginHosts, addMarkdown };
+      return {
+        harness,
+        mounted,
+        forwarderOwners,
+        pluginHosts: controller.hosts,
+        addMarkdown,
+      };
     }
 
     /** Returns a `codeBlock` renderer stamping its result with `owner`. */
@@ -1878,6 +1884,86 @@ HTTP: http://example.com
     });
   });
 
+  // ── The mount detail's `kind` discriminant (#2273) ────────────────────
+  // Both dispatch sites offer a host over the same event, and a listener used
+  // to tell them apart by testing whether `detail.element` was set — the shape
+  // was the discriminant. `kind` publishes what the element already knows.
+  describe('plugin-host mount detail kind', () => {
+    const KIND_HARNESS_TAG = 'cds-test-markdown-kind-host';
+
+    if (!customElements.get(KIND_HARNESS_TAG)) {
+      customElements.define(
+        KIND_HARNESS_TAG,
+        class extends HTMLElement {
+          connectedCallback() {
+            if (!this.shadowRoot) {
+              this.attachShadow({ mode: 'open' });
+            }
+          }
+        }
+      );
+    }
+
+    /** Records every mount detail without claiming anything. */
+    async function createDetailRecorder() {
+      const harness = await fixture<HTMLElement>(
+        html`<cds-test-markdown-kind-host></cds-test-markdown-kind-host>`
+      );
+      const details: Array<Record<string, unknown>> = [];
+      harness.addEventListener(
+        'cds-aichat-markdown-plugin-host-mount',
+        (event) => {
+          details.push((event as CustomEvent<Record<string, unknown>>).detail);
+        }
+      );
+      async function addMarkdown(
+        markdown: string,
+        props: Partial<MarkdownElementInstance> = {}
+      ) {
+        const el = document.createElement(
+          MARKDOWN_ELEMENT_TAG
+        ) as MarkdownElementInstance;
+        Object.assign(el, props, { markdown });
+        harness.shadowRoot?.appendChild(el);
+        await el.updateComplete;
+        return el;
+      }
+      return { harness, details, addMarkdown };
+    }
+
+    it('marks a plugin-fallback offer as pluginFallback', async () => {
+      const { details, addMarkdown } = await createDetailRecorder();
+
+      await addMarkdown('Hi :tag: there', {
+        markdownItPlugins: [tagPlugin],
+      } as Partial<MarkdownElementInstance>);
+
+      expect(
+        details.length,
+        'the plugin token should offer a host'
+      ).to.be.at.least(1);
+      expect(details[0].kind).to.equal('pluginFallback');
+      expect(details[0].html).to.be.a('string');
+      expect(details[0].element).to.equal(undefined);
+    });
+
+    it('marks a customRenderers offer as customRenderer', async () => {
+      const { details, addMarkdown } = await createDetailRecorder();
+
+      await addMarkdown('```json\n{ "a": 1 }\n```', {
+        customRenderers: { codeBlock: () => document.createElement('div') },
+      } as Partial<MarkdownElementInstance>);
+
+      expect(
+        details.length,
+        'the code block should offer a host'
+      ).to.be.at.least(1);
+      expect(details[0].kind).to.equal('customRenderer');
+      expect(details[0].element).to.be.instanceOf(HTMLElement);
+      expect(details[0].html).to.equal(undefined);
+    });
+  });
+
   // ── Late-subscriber handshake (#2271) ──────────────────────────────────
   // `createRelocationHarness` above does the container's job and the React
   // wrapper's job in one listener, so it can never miss its own event. The
@@ -1930,58 +2016,11 @@ HTTP: http://example.com
       const harness = await fixture<HTMLElement>(
         html`<cds-test-markdown-late-subscriber-host></cds-test-markdown-late-subscriber-host>`
       );
-      const pluginHosts = new Map<string, HTMLElement>();
 
-      harness.addEventListener(
-        'cds-aichat-markdown-plugin-host-mount',
-        (event) => {
-          const detail = (
-            event as CustomEvent<{
-              slotName: string;
-              html?: string;
-              element?: HTMLElement;
-              isInline: boolean;
-            }>
-          ).detail;
-          // Custom-renderer hosts carry a live element the markdown element
-          // manages itself — never hoisted, never claimed. Matches all three
-          // real containers.
-          if (detail.element) {
-            return;
-          }
-          event.preventDefault();
-          const host =
-            pluginHosts.get(detail.slotName) ??
-            document.createElement(detail.isInline ? 'span' : 'div');
-          host.innerHTML = detail.html ?? '';
-          host.setAttribute('slot', detail.slotName);
-          pluginHosts.set(detail.slotName, host);
-          if (host.parentElement !== harness) {
-            harness.appendChild(host);
-          }
-        }
-      );
-      harness.addEventListener(
-        'cds-aichat-markdown-plugin-host-update',
-        (event) => {
-          const detail = (
-            event as CustomEvent<{ slotName: string; html: string }>
-          ).detail;
-          const host = pluginHosts.get(detail.slotName);
-          if (host && host.innerHTML !== detail.html) {
-            host.innerHTML = detail.html;
-          }
-        }
-      );
-      harness.addEventListener(
-        'cds-aichat-markdown-plugin-host-unmount',
-        (event) => {
-          const { slotName } = (event as CustomEvent<{ slotName: string }>)
-            .detail;
-          pluginHosts.get(slotName)?.remove();
-          pluginHosts.delete(slotName);
-        }
-      );
+      // The container half, shipped implementation. It is attached here, at
+      // harness construction, the way a chat container attaches at chat mount.
+      const controller = createMarkdownPluginHostController(harness);
+      controller.connect();
 
       async function addMarkdown(
         markdown: string,
@@ -2023,10 +2062,10 @@ HTTP: http://example.com
         el.addEventListener(
           'cds-aichat-markdown-plugin-host-mount',
           (event) => {
-            const detail = (
-              event as CustomEvent<{ slotName: string; element?: HTMLElement }>
-            ).detail;
-            if (detail.element) {
+            const detail = resolveMarkdownPluginHostMountDetail(
+              (event as CustomEvent<MarkdownPluginHostMountDetailInput>).detail
+            );
+            if (detail.kind !== 'pluginFallback') {
               return;
             }
             addForwarder(detail.slotName);
@@ -2047,7 +2086,12 @@ HTTP: http://example.com
         return { forwarders, announced };
       }
 
-      return { harness, pluginHosts, addMarkdown, subscribeForwarder };
+      return {
+        harness,
+        pluginHosts: controller.hosts,
+        addMarkdown,
+        subscribeForwarder,
+      };
     }
 
     it('projects plugin output when the forwarder subscribes after the first reconcile', async () => {
--- a/packages/ai-chat-components/src/components/markdown/__tests__/markdown.test.ts
+++ b/packages/ai-chat-components/src/components/markdown/__tests__/markdown.test.ts
@@ -1499,30 +1499,31 @@ HTTP: http://example.com
   // weren't, two messages whose code block both started on line 0 minted the
   // same name: the first element's slot gathered both hosts and the second
   // gathered none. See issue #2099.
+  // Minimal inline plugin whose output goes through the plugin-fallback slot
+  // path (the same page-level hoisting, a separate slot-name mint site). Shared
+  // by the two describes below; both need a plugin-fallback descriptor.
+  function tagPlugin(md: any) {
+    md.inline.ruler.before(
+      'text',
+      'cds_test_tag',
+      (state: any, silent: boolean) => {
+        if (!state.src.slice(state.pos).startsWith(':tag:')) {
+          return false;
+        }
+        if (!silent) {
+          state.push('cds_test_tag', '', 0);
+        }
+        state.pos += ':tag:'.length;
+        return true;
+      }
+    );
+    md.renderer.rules.cds_test_tag = () =>
+      `<span class="cds-test-tag">tag</span>`;
+  }
+
   describe('slot names across markdown elements', () => {
     const HARNESS_TAG = 'cds-test-markdown-relocation-host';
     const codeMarkdown = '```json\n{ "name": "Alice" }\n```';
-
-    // Minimal inline plugin whose output goes through the plugin-fallback slot
-    // path (the same page-level hoisting, a separate slot-name mint site).
-    function tagPlugin(md: any) {
-      md.inline.ruler.before(
-        'text',
-        'cds_test_tag',
-        (state: any, silent: boolean) => {
-          if (!state.src.slice(state.pos).startsWith(':tag:')) {
-            return false;
-          }
-          if (!silent) {
-            state.push('cds_test_tag', '', 0);
-          }
-          state.pos += ':tag:'.length;
-          return true;
-        }
-      );
-      md.renderer.rules.cds_test_tag = () =>
-        `<span class="cds-test-tag">tag</span>`;
-    }
 
     if (!customElements.get(HARNESS_TAG)) {
       customElements.define(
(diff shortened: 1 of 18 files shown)


ethanwinters:
Closing in favor of #2299, which re-lands this work off a clean base.

This branch mixed the line-break fixes with plugin-host changes that have since been reworked, and it predates the finding that consumer CSS stopped reaching custom-renderer content — the regression that now leads the replacement PR.

PR #2304, opened 2026-09-08
fix(demo): drop the experimental loader that breaks playwright

ethanwinters:
`npm run ci-check` fails on `main` before any demo spec runs.

On Node 24, an off-thread `--experimental-loader` breaks Playwright's own sync `module.registerHooks` transform. The CommonJS entry point goes through the ESM loader. `defaultLoad` gives it a null source. `validateLoad` throws.

The loader was already dead weight. It patched around a missing `es/globals/*` export in `@carbon/web-components`. Version 2.62.0 ships that export. So: deleted, not ported.

`node-polyfill.js` stays. Its premise may have expired too. Untested.

#### Changelog

**Changed**

- Demo test script no longer adds a custom ESM loader
- `demo/AGENTS.md` says why `--experimental-loader` is off-limits here

**Removed**

- `demo/tests/alias-loader.js`

#### Testing / Reviewing

- Run `npm run test --workspace=@carbon/ai-chat-examples-demo`. 44 pass.
- Run `npm run ci-check`. Green.
- For the old failure, put the flag back on the parent commit. It dies at startup.
- Exit 143 from `frameworks-vite` was lerna's SIGTERM to siblings. That one passes alone.


AlexanderMelox (approved):
LGTM

Issue #2103, opened 2026-08-10
Surface file-upload failures so users can see and recover from them


ethanwinters:
### Value proposition

When a file upload fails, you can't tell what went wrong or do anything about it.

A file rejected by `accept` or `maxFileSizeBytes` disappears with no visible trace — no chip, no message, nothing. A file rejected by your `onFileUpload` handler does show an error, but that error's title is hardcoded English, it's announced politely even though it blocks you from sending, and it never says what to do next. Recovering means removing the attachment and attaching it again, and nothing on screen tells you that.

Hosts can't paper over any of this. `UploadConfig` gives you a callback and four limits, and no control over the copy the chat wraps around your error.

### Expected outcomes

- [ ] Every rejection path — `accept`, `maxFileSizeBytes`, `maxFiles`, and a thrown `onFileUpload` — produces a visible result, not just an announcement.
- [ ] No user-visible upload string is hardcoded in a component; all route through the language pack.
- [ ] A blocking upload error is announced assertively and exactly once, and the errored chip carries its state programmatically.
- [ ] A failed upload tells the user how to recover from it.
- [ ] The `prompt-line-file-upload` examples demonstrate the failure path in both flavors.

### Decisions

**Recovery is remove-and-re-attach.** There is no retry control and no retry API. A user who hits a failed upload removes the attachment and attaches it again — the same gesture that clears the error and re-enables sending. This was scoped down deliberately rather than dropped; see #2107 for what it replaced. It leaves one obligation on the remaining children: the chat has to *say* so, which is why the recovery string sits in #2105.

No ADR is needed. With retry gone, no child adds public API beyond additive language-pack keys.

### Out of scope

- The human-agent (service desk) upload path. It has its own status pipeline through `HumanAgentServiceImpl` and its own `FileUpload` state; this epic covers the assistant path behind `UploadConfig` only.
- Drag-and-drop attachment, which the chat does not support today.
- Upload progress reporting. `onFileUpload` resolves once; there is no progress channel, and adding one is separate work.
- Which surface owns the visible error copy. The message renders in both the chip and the prompt line today, and that stays as it is pending a design decision.
- A retry control, and any public API for reading or re-running a failed upload. See the decision above.

### Details

The sub-issue list on this epic is the child record, so it is not restated here.

Reproductions behind every child come from driving the React `prompt-line-file-upload` example with a handler that rejects any file whose name starts with `a`. Each child cites the specific one it fixes.

[...]

PR #2305, opened 2026-09-09
feat(upload): announce failures once and expose the chip's error state

### LINKED ISSUE #2106: Announce upload failures once, assertively, and expose the chip's error state
### Background

A failed upload blocks the user from sending, yet it's announced politely and three times over. The errored chip carries no programmatic error state, and with two attachments a screen-reader user can't tell the two remove buttons apart.

Three defects, all on the same failure:

**Announced politely, three times.** `AnnounceOnMount` wraps its children in an `aria-live="polite"` region *and* calls the announcer with the same text, so the message lands twice. The `file-uploads` component adds a third, generic `"There was an error uploading the file."` Reproduction: reject an upload in the React `prompt-line-file-upload` example and read the live regions — two carry the composed title and description, a third carries the generic sentence. The package's own guidance is that a blocking error announces assertively; this one blocks sending and announces politely.

**The chip has no programmatic error state.** The host element carries no `aria-invalid`, no `role`, and its accessible name is just the filename. Carbon renders the message as a plain `div.cds--form-requirement__title` that nothing references — no `aria-describedby`, and the invalid icon is `aria-hidden`. A user who tabs back to the chip later is told the filename and nothing else.

**Both remove buttons are named identically.** With `ok.txt` and `a-reject.txt` attached, the accessibility tree holds two buttons both named `"Remove file"`. Nothing distinguishes them, so removing the failed one is guesswork. WCAG 2.1 AA 4.1.2.

Parent epic: #2103

### Goal

A failed upload announces once, assertively, and the errored chip carries its state and its file's identity programmatically.

### Acceptance criteria

- [ ] A failed upload produces exactly one announcement. Demo: reject an upload in the example, read the live regions, confirm one carries the message and no other repeats it.
- [ ] That announcement is assertive. Proof: the announce call passes `assertive: true`; spec under `packages/ai-chat/tests/`.
- [ ] The errored chip sets `aria-invalid` on a host element that exposes it, and its error text is referenced by `aria-describedby`. Proof: web-test-runner spec in `packages/ai-chat-components/` asserting both on `cds-aichat-file-upload-item`.
- [ ] Each remove button's accessible name includes its file name. Proof: the same spec asserts two chips yield two distinct button names. The label comes from the language pack with the file name interpolated, not string concatenation in the component.
- [ ] A successful upload still announces success exactly once, unchanged. Spec: existing `file-uploads` announcement coverage still passes.
- [ ] Verified with VoiceOver and NVDA on the failure path, per `packages/ai-chat-components/AGENTS.md`.
- [ ] Definition-of-done gate for both `packages/ai-chat/` and `packages/ai-chat-components/` passes: https://github.com/carbon-design-system/carbon-ai-chat/blob/main/references/definition-of-done.md

### Public API / contract

```typescript
// Two public surfaces move: one language-pack key added, and the
// cds-aichat-file-upload-item element gains a labelling property.

interface LanguagePack {
  // ...
  /** Accessible name for the button that removes a named attachment. */
  fileSharing_removeNamedFile: string; // "Remove {filename}"
}

// The element's existing `removeFileLabel` stays for the unnamed case. Add a
// property carrying the per-file label so the element does no interpolation:
//
//   /** Accessible name for the remove button, with the file name already
//     * interpolated by the caller. Falls back to `removeFileLabel`. */
//   removeFileNamedLabel?: string;

// Preconditions: none. A chip with no resolvable name falls back to the
//   existing generic label rather than rendering an empty name.

// No-op case: a read-only chip renders no remove button, so neither label
//   applies and nothing changes for attachments on sent messages.

// Failure: unchanged. This issue does not alter when an upload errors, only
//   how that error is exposed.

// Events: unchanged. `cds-aichat-file-remove` still fires with the same detail.

// Timing: the announcement fires on the uploading -> error transition, in the
//   same update that sets the chip's error state. Not before.

// Repeat calls: two files failing in the same frame announce once each rather
//   than coalescing. Unlike "added" and "uploading", which are counted and
//   batched, a failure carries a file-specific reason that a count would lose.

// Derivation: politeness follows blocking-ness, and that is the rule this
//   issue establishes —
//     assertive  a failure that blocks sending (a rejected onFileUpload)
//     assertive  a client-side rejection (already assertive today)
//     polite     a non-blocking status change (added, uploading, removed,
//                uploaded successfully)

// Announcement: one assertive announcement carrying the localized title and the
//   host's message. Focus does not move — moving focus on an async failure
//   would yank the caret out of the input mid-typing.

// Ownership: the framework announces. The host supplies only the message text,
//   by throwing from onFileUpload.
```

### Out of scope

- The hardcoded title in the announced text — sibling issue. This issue changes politeness and count, not the words.
- The visible copy. The message renders in both the chip and the prompt line, and that duplication stays pending a design decision. Only the announcements are deduplicated here.
- Making client-side rejections visible — sibling issue. Their politeness is already correct.
- Reworking `AnnounceOnMount` generally. It has the same double-announce shape wherever it's used with `announceOnce`; fixing the pattern repo-wide is separate.

### Related

Parent epic: #2103

AlexanderMelox:
Closes #2106

This PR fixes 3 issues
1. Issue where a failed upload was announced three times and was exposed by an aria politely. 
2. The errored file upload chip or tile exposed no error state
3. The two attachments gave two buttons the same name "Remove file"

#### Changelog

**New**

- `fileSharing_removeNamedFile` language-pack key — `"Remove {filename}"`.
- `removeFileNamedLabel` prop on `cds-aichat-file-upload-item` web component.
- `getRemoveFileLabel` and `getFileUploadFailureText` on `cds-aichat-file-uploads`.
  Formatters, so the consumer owns the wording.

**Changed**

- Upload failures announce once, assertively. Was three times, politely.
- The announcement carries that file's own reason. Two files failing together give one
  announcement holding both. The title is stated once.
- An errored chip exposes `role="group"` and `aria-invalid`. Its `aria-describedby`
  points at the reason.
- Each remove button is named after its file.
- `fileSharing_uploadFailed` no longer reaches the upload path. The human-agent flow
  still uses it. Set `getFileUploadFailureText` to change the announced text.

#### Testing / Reviewing

Build first, then run the React upload example:

```bash
npm run aiChat:build
npm run start --workspace=@carbon/ai-chat-examples-react-prompt-line-file-upload
```

Attach two good files. Each remove button is named after its own file. The two
names differ.

Now attach `fail.txt`. The example rejects any name holding "fail". Confirm:

- One live region holds the message. It is `aria-live="assertive"`.
- The input-area chip has `role="group"` and `aria-invalid="true"`.
- Its remove button's `aria-describedby` resolves to the reason.
- A good file still announces success once, politely.
- Chips on sent messages are untouched. No `role`, no `aria-label`.

Then attach `also-fail.txt` too. Remove the first one. The other file's reason
is announced.

Automated:

```bash
npm test --workspace=@carbon/ai-chat-components
npm test --workspace=@carbon/ai-chat
```

Still unverified with VoiceOver and NVDA. The risk sits in `AnnounceOnMount`.
It flips `aria-live` from `off` to `polite` as the text changes. It also calls
the announcer. The error may be spoken twice.


packages/ai-chat-components/src/components/file-uploads/src/file-upload-item.ts
```diff
@@ -212,6 +225,86 @@ class FileUploadItemElement extends LitElement {
...
+
+    // Shadow DOM hides the chip's implicit semantics, and `aria-invalid` on a
+    // role-less host is dropped — so the host states a role and names itself.
+    this.setAttribute('role', 'group');
+    const name = this._resolved?.name;
+    if (name) {
+      this.setAttribute('aria-label', name);
+    } else {
+      this.removeAttribute('aria-label');
+    }
```
ethanwinters, line 258: **What this block does:** `_syncErrorState` runs on every chip this component renders — and it renders in two places. In the input area it's a pending attachment you can remove; in the message list it's a read-only chip on an already-sent message (`MessageFileAttachments.tsx` mounts it with `readOnly`). The `invalid` local on line 248 accounts for that split, but only `aria-invalid` below uses it.

**Important** — `role="group"` and `aria-label` are set above the `readOnly` check, so every attachment on every *sent* message now announces a named grouping boundary and then repeats the filename Carbon already renders as text. Those chips sit in `ul role="list" > li` in `MessageFileAttachments.tsx:44-52`, where `listitem` is the role that belongs; the input-area container has no `role="list"`, so the group buys no position or count there either.

The package rule at `packages/ai-chat-components/AGENTS.md:26` also asks for NVDA + VoiceOver on a component that announces or has dynamic state, and the PR says that hasn't happened yet. Could you run VoiceOver on the failure path (attach `a-reject.txt` in `examples/react/prompt-line-file-upload`) and confirm how the group reads? This suggestion is the narrow version in the meantime — it leaves sent-message chips exactly as they were and keeps the role where the `aria-invalid` rationale actually applies.

```suggestion
    // Shadow DOM hides the chip's implicit semantics, and `aria-invalid` on a
    // role-less host is dropped — so an input-area chip states a role and names
    // itself. A read-only chip is already a list item in the message list, so it
    // is left as Carbon renders it.
    if (this.readOnly) {
      this.removeAttribute('role');
      this.removeAttribute('aria-label');
    } else {
      this.setAttribute('role', 'group');
      const name = this._resolved?.name;
      if (name) {
        this.setAttribute('aria-label', name);
      } else {
        this.removeAttribute('aria-label');
      }
    }
```

packages/ai-chat-components/src/components/file-uploads/src/file-upload-item.ts
```diff
@@ -212,6 +225,86 @@ class FileUploadItemElement extends LitElement {
...
+      }
+    }
+
+    const removeButton = innerRoot.querySelector('button.cds--file-close');
+    if (removeButton) {
+      if (invalid) {
+        removeButton.setAttribute('aria-invalid', 'true');
+      } else {
+        removeButton.removeAttribute('aria-invalid');
+      }
```
ethanwinters, line 301: **What this block does:** having reached into Carbon's shadow root, this wires up the close button — `aria-invalid` to mark it, `aria-describedby` to point it at the failure reason. The `aria-describedby` half is the part that carries the reason to the user, and it works; this is only about the `aria-invalid` half above it.

**Important** — ARIA 1.2 lists `aria-invalid` under *Inherited States and Properties* for `button` as "deprecated on this role" (same for `group`), so it is not dropped today but is on a removal path. More immediately, it asserts that the *remove control* is invalid rather than the attachment — a user focusing the button hears the button reported as the invalid thing. The `aria-describedby` below already gives them the reason, which is the part that carries.

```suggestion
    if (removeButton) {
```

Two assertions pin the attribute this removes, so the companion suggestions on `file-upload-item.test.ts:355` and `:387` have to be applied with it or the suite goes red.

packages/ai-chat-components/src/components/file-uploads/src/file-uploads.ts
```diff
@@ -213,6 +255,14 @@ class FileUploadsElement extends LitElement {
...
     }
 
+    if (failures.length > 0) {
+      this._announcer.announce(
+        this.getFileUploadFailureText({
+          messages: failures.filter(Boolean),
+        }),
+        'assertive'
+      );
+    }
```
ethanwinters, line 265: **What this block does:** `_announceTransitions` walks this frame's uploads against `_snapshots` (last frame's), collects what changed, and speaks it. The loop above classifies each upload by the *edge* it crossed — newly present, healthy→errored, reason revised — and pushes onto `failures`; this block turns that into one assertive announcement.

**Important** — `hasUploadError` in `Input.tsx:611` is a *state* predicate but this announces only on *edges*, so a frame where the visible reason changes without crossing an edge is announced by neither. Two files fail; `AppShell.tsx:457-462` `.find()`s file A so the screen states A's reason; the user removes A; `inputError` recomputes to B's reason, the wrapper is still silenced, and B's snapshot is unchanged so no branch here fires — they hear "File removed." and never B's reason.

Diffing the whole failure set per frame also fixes the second problem: `ChatActionsImpl.ts:748-795` dispatches per file, so two files failing milliseconds apart currently make two formatter calls that the announcer joins, repeating the title and recovery sentence. Worth a look before you hit apply — this trades that repetition for re-stating A's reason when B fails later, which I think is the better read but is a judgement call.

The three `failures.push` branches above are dead once this lands. They need removing by hand: the `const failures` declaration sits outside the diff, so it can't ride along in a suggestion.

```suggestion
    // Announce the whole current failure set whenever it differs from last
    // frame's, so removing one of two failed files still speaks the reason that
    // is now on screen.
    const failureMessages = this.uploads
      .filter((upload) => upload.isError)
      .map((upload) => upload.errorMessage ?? '')
      .filter(Boolean);
    const previousMessages = [...previous.values()]
      .filter((snapshot) => snapshot.isError)
      .map((snapshot) => snapshot.errorMessage ?? '')
      .filter(Boolean);
    const unchanged =
      failureMessages.length === previousMessages.length &&
      failureMessages.every(
        (message, index) => message === previousMessages[index]
      );
    if (failureMessages.length > 0 && !unchanged) {
      this._announcer.announce(
        this.getFileUploadFailureText({ messages: failureMessages }),
        'assertive'
      );
    }
```

packages/ai-chat/src/chat/languages/en.json
```diff
@@ -190,6 +190,7 @@
   "fileSharing_ariaAnnounceSuccess": "The file was uploaded successfully.",
   "fileSharing_fileIcon": "File icon",
   "fileSharing_removeButtonTitle": "Remove file",
+  "fileSharing_removeNamedFile": "Remove {filename}",
```
ethanwinters, line 193: **Why this one line matters more than it looks:** `LanguagePack` is declared as `typeof enLanguagePack`, so every key in this file is public API. A host overrides copy by passing `strings` into `PublicConfig`, keyed by exactly these names — which means renaming one after release is a breaking change for consumers, not a refactor.

**Important** — this file's convention for an interpolated variant of an existing label is base + `With<Thing>`: `messages_searchResultsOpenDocumentWithLabel:65`, `codeSnippet_ariaLabelReadOnlyWithLanguage:223`, `codeSnippet_ariaLabelEditableWithLanguage:225`. `LanguagePack = typeof enLanguagePack` makes the key public API, so a rename after release is a breaking change — worth getting right now.

The name came straight from #2106's contract, so this needs fixing in both places: apply here, apply the companion suggestion on `Input.tsx:302`, and amend the issue.

```suggestion
  "fileSharing_removeButtonTitleWithName": "Remove {filename}",
```

packages/ai-chat/src/chat/components/input/Input.tsx
```diff
@@ -292,6 +292,39 @@ function Input(props: InputProps, ref: Ref<InputFunctions>) {
...
   );
 
+  // Interpolated here rather than in the component so a translation can order
+  // the name however its language needs.
+  const getRemoveFileLabel = useMemo(
+    () =>
+      ({ name }: { name?: string }) =>
+        name
+          ? intl.formatMessage(
+              { id: 'fileSharing_removeNamedFile' },
```
ethanwinters, line 302: **Context:** this is the only reader of the new language-pack key — `getRemoveFileLabel` interpolates the filename here so each remove button gets a distinct accessible name.

**Important** — companion to the `en.json:193` rename. Both suggestions have to be applied together or `formatMessage` resolves nothing.

```suggestion
              { id: 'fileSharing_removeButtonTitleWithName' },
```

packages/ai-chat/tests/instance/spec/uploadFailureAnnounce_spec.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    await waitFor(() => {
+      const carrying = liveRegions().filter((region) =>
+        region.text.includes('5 MB limit')
+      );
+      // A blocking failure interrupts, and no second region repeats it — the
+      // defect was the same text landing in three regions, two of them polite.
+      expect(carrying).toHaveLength(1);
+      expect(carrying[0].live).toBe('assertive');
+    });
```
ethanwinters, line 87: **What this test is for:** it's the proof for AC 1 of #2106 — "A failed upload produces exactly one announcement" — which was the headline defect, since the old code announced three times. So this assertion is load-bearing: if it can pass while a duplicate exists, the regression it guards against can come back unnoticed.

**Important** — the "exactly once" count sits inside `waitFor`, which returns at the first poll where one region carries the text, so a duplicate landing after that poll is never observed. AC 1 is the criterion this spec exists to prove, and its proof is timing-dependent. Wait for the text, settle past the announcer's 250 ms debounce, then assert the count outside `waitFor`.

I checked this rather than assuming it: with a second announcement injected 300 ms after the first, the spec as written **passes**, and with this change it fails on `Expected length: 1, Received length: 2`. A duplicate that lands promptly is still caught either way — it's the late one that slips through.

```suggestion
    await waitFor(() =>
      expect(
        liveRegions().filter((region) => region.text.includes('5 MB limit'))
      ).not.toHaveLength(0)
    );

    // Settle past the announcer's 250 ms debounce so a second write to a
    // different region is observed rather than raced.
    await new Promise((resolve) => setTimeout(resolve, 400));

    // A blocking failure interrupts, and no second region repeats it — the
    // defect was the same text landing in three regions, two of them polite.
    const carrying = liveRegions().filter((region) =>
      region.text.includes('5 MB limit')
    );
    expect(carrying).toHaveLength(1);
    expect(carrying[0].live).toBe('assertive');
```

packages/ai-chat-components/src/components/file-uploads/__tests__/file-upload-item.test.ts
```diff
@@ -280,4 +294,132 @@ describe('file-upload-item', () => {
...
+      await settle(el);
+
+      // aria-invalid needs a role to be exposed; on a role-less host it is
+      // dropped.
+      expect(el.getAttribute('role')).to.equal('group');
+      expect(el.getAttribute('aria-invalid')).to.equal('true');
+
+      const root = innerRoot(el);
+      const button = root.querySelector('button.cds--file-close')!;
+      expect(button.getAttribute('aria-invalid')).to.equal('true');
```
ethanwinters, line 355: **Context:** this is the errored-chip test, and it asserts on two different nodes — the host element and Carbon's inner close button. Only the button's `aria-invalid` assertion is affected here.

**Important** — companion to removing the remove button's `aria-invalid`. The `aria-describedby` assertions below are the ones that prove the outcome; this line pinned the attribute that says the *button* is invalid.

```suggestion
      const button = root.querySelector('button.cds--file-close')!;
```

packages/ai-chat-components/src/components/file-uploads/__tests__/file-upload-item.test.ts
```diff
@@ -280,4 +294,132 @@ describe('file-upload-item', () => {
...
+        makeUpload('a', FileStatusValue.EDIT, { isError: true })
+      );
+      await settle(el);
+
+      // Still invalid — the state is real even when the host said nothing about
+      // why — but pointing at Carbon's empty requirement node would describe the
+      // button with silence.
+      expect(el.getAttribute('aria-invalid')).to.equal('true');
+      const button = innerRoot(el).querySelector('button.cds--file-close')!;
+      expect(button.getAttribute('aria-invalid')).to.equal('true');
```
ethanwinters, line 387: **Context:** this is the no-reason-given case — the chip is invalid but the host threw an `Error` with no message, so there's nothing to describe it with.

**Important** — companion to removing the remove button's `aria-invalid`. The host's `aria-invalid` on line 385 still proves the chip carries the state.

```suggestion
      const button = innerRoot(el).querySelector('button.cds--file-close')!;
```

ethanwinters (changes requested):
Three real defects here, all in the same area: an announcement that goes missing, one that doubles, and a test that can't catch either. Details below, and eight of the fixes are suggestions you can apply from this page.

Worth saying first that the hard part landed. I checked the `aria-describedby` wiring against Carbon's published `cds-file-uploader-item` template: the IDREF stays inside a single shadow root, lit-html reuses the requirement `div` so your imperative `id` survives Carbon's re-render, and `iconDescription` really is what names the close button. That mechanism is sound.

## The three defects

**1. Removing one of two failed files leaves the new error unannounced.** `hasUploadError` in `Input.tsx:611` is a *state* predicate — "is any upload errored" — and it silences the `AnnounceOnMount` wrapper for as long as that's true. But `<FileUploads>` announces on *edges*: a new id, healthy→errored, or a changed `errorMessage`. So any frame where the visible error text changes without crossing one of those edges is announced by neither side. Concretely: two files fail; `AppShell.tsx:457-462` `.find()`s the first one, so the prompt line shows A's reason; the user removes A; `inputError` recomputes to B's reason and the screen updates; but B's snapshot hasn't changed, so no branch fires, and the wrapper is still muted. The user hears "File removed." and is never told why sending is still blocked.

**2. Two files failing at different times repeat the title and recovery sentence.** The `failures` array only states the title once when both failures land in the same Lit update. `ChatActionsImpl.ts:748-795` awaits `onFileUpload` per file and dispatches from each `catch`, so two files that fail a few hundred milliseconds apart produce two separate frames, two formatter calls, and two full compositions that `AriaAnnouncerManager` then joins inside its 250 ms window — "File upload error. <reason A>. Remove the attachment and try again. File upload error. <reason B>. Remove the attachment and try again." The batching in the PR description holds only for the synthetic case the unit test drives, where the whole `uploads` array is replaced at once.

**3. The "exactly once" test can pass while a duplicate exists.** The count assertion sits inside `waitFor`, which returns at the first poll where exactly one region carries the text — so a second announcement arriving after that poll is never seen. I verified this rather than assuming it: with a duplicate injected 300 ms after the first, the spec as written **passes**. With the suggested change it fails on `Expected length: 1, Received length: 2`.

Defects 1 and 2 share a fix, which is the suggestion on `file-uploads.ts:258`.

## Needs a hand — can't be a suggestion

**`AnnounceOnMount.tsx:70` — the once-latch burns on a message that was never spoken.** `this.onceAnnounced = true` sits outside the branch that actually announces, so mounting with `live={false}` marks the component as having announced when it stayed silent. That matters because you deliberately keep the wrapper mounted rather than swapping it out, and `componentDidMount` only ever runs once: after an upload failure, the max-char-count branch at `Input.tsx:595` reuses the very same instance (same element type at the same position, so React reconciles instead of remounting), and its message never reaches `ariaAnnouncer`. Whether the user hears anything then depends entirely on whether flipping `aria-live` from `off` back to `polite` in the same commit as the text change still speaks — which is exactly the behavior no one has verified yet. Set the latch only where the announcement fires, and add a `componentDidUpdate` that announces when `live` goes true with an unannounced `announceOnce`. No button for this one: the latch line falls outside the diff.

**`packages/ai-chat/src/types/config/UploadConfig.ts:58-62` — a host's override silently stops working.** `fileSharing_uploadFailed` no longer reaches the upload path; its only remaining reader is `HumanAgentServiceImpl.ts:1290`. Anyone who overrode that key through `PublicConfig.strings` to customize upload-failure copy loses it with no error and no warning — the announcement now composes `fileSharing_uploadErrorTitle` and `fileSharing_uploadErrorRecovery` instead. That JSDoc block already tells hosts those two keys own the on-screen copy; it should say they own the announcement too. The file isn't in this diff, so it can't take a line comment.

**Public API exceeds what #2106 and #2103 planned for.** The issue's contract says "Two public surfaces move: one language-pack key added, and the `cds-aichat-file-upload-item` element gains a labelling property." `getRemoveFileLabel` and `getFileUploadFailureText` are two more. This isn't just bookkeeping: epic #2103 rests its "No ADR is needed" specifically on "no child adds public API beyond additive language-pack keys," so adding element properties quietly removes the ground that decision stood on. Either fold the composition into the existing labels, or amend both and note why an ADR still isn't needed. Same for the contract's "two files failing in the same frame announce once each rather than coalescing," which this PR deliberately departs from — I think correctly, but it should be written down.

**PR description, Testing section.** It asks the reviewer to hand-edit `mockOnFileUpload.ts` because "No example fails an upload yet," but the main merge in `8814bee4` brought that in — `mockOnFileUpload.ts:48-51` already rejects any file whose name starts with "a". Attaching `a-reject.txt` is enough now.

## Test gaps

`role` and `aria-label` on healthy and read-only chips are unasserted (`role` is checked only in the errored case, `aria-label` nowhere), so the changed accessible name of every sent-message attachment ships unproven. Nothing covers an error transitioning from an upload failure to a `publicConfig.input.error` with no intervening clear — the path in the `AnnounceOnMount` finding above. Nothing covers re-setting `uploads` with the same errored file *not* re-announcing. And `getFileUploadFailureText` is only exercised with a single message, so the two-reason composition is proven only against the component's default formatter. The nameless-chip fallback is dead as well as untested: `types.ts:14-17` states an upload always carries its `File`.

## Verification

I applied all eight suggestions locally before proposing them. Prettier and eslint clean, `@carbon/ai-chat-components` builds, its file-uploads suite is 47/47 across Chromium, Firefox and WebKit, and `uploadFailureAnnounce_spec` is 4/4. Reverted afterwards — nothing pushed.

CI's failing `test` check isn't yours: the job died installing Playwright browsers (`Hash Sum mismatch` from `dl.google.com`).

AlexanderMelox:
Pushed the six suggestions and the dead `failures.push` branches. Two places where what landed differs from what was suggested — both worth a look before you re-review.

**`file-upload-item.ts` — the ARIA wiring now goes through a helper.** Scoping the role to non-read-only chips added a fourth set-or-remove pair to a method that already had three, so each is now one call:

```ts
syncAttribute({ element: this, attribute: 'aria-invalid', value: invalid ? 'true' : undefined });
```

AlexanderMelox:
Hey @ethanwinters also want to bring this up to your attention on some issue / PR deviation.

There is three places the code and the written issue #2106 now conflict.The first two I think we should update the issue to reflect your feedback and the PR's changes the third I will need your call on.

1. Key name. [#2106](https://github.com/AlexanderMelox/carbon-ai-chat/issues/2106)'s contract says fileSharing_removeNamedFile; the code ships fileSharing_removeButtonTitleWithName after your rename. 

2. Coalescing. [#2106](https://github.com/AlexanderMelox/carbon-ai-chat/issues/2106) says two files failing in the same frame "announce once each rather than coalescing," and the changes now coalesces. Should we amend the issue since you mentioned that the changes as a better read than what was stated on the issue?

3. Public API. [#2103](https://github.com/AlexanderMelox/carbon-ai-chat/issues/2103) rests "No ADR is needed" on "no child adds public API beyond additive language-pack keys." This PR adds removeFileNamedLabel (planned in [#2106](https://github.com/AlexanderMelox/carbon-ai-chat/issues/2106)) plus getRemoveFileLabel and getFileUploadFailureText (not planned). Two ways out: fold the formatters back into existing labels so the epic's sentence stays true, or keep them and amend #2103 with why an ADR still isn't warranted. You made the no-ADR call, so I'd rather not overturn it from an implementation PR.

ethanwinters (approved):

Issue #1731, opened 2026-07-13
Move the conversation layer to src/shared/ and make it React-free


ethanwinters:
### Background

The chat cannot run without React, even for a host that renders no React. Every entry point the package ships boots a React tree. The conversation machinery underneath it is tangled into that tree — sending, receiving, history, human agents, state. So a host on Vue, Angular, or plain elements pays for a framework it never asked for. The headless path that would let it opt out cannot be built while the tangle holds.

The tangle is smaller than it looks. But it is not only about imports. The conversation code and the UI already sit in separate folders. Only a handful of imports cross from the first into the second. What is missing is a home for the clean half. Without one, nothing stops those edges growing back.

So this issue does two things. It gives the conversation layer its own top-level folder. Then it cuts the edges that reach out of it. All of it is mechanical, and no behavior changes.

Parent epic: #1728. ADR-0002 governs it. That record sets the three-folder layout used below.

**On size.** The move touches nearly every import in the package. Each hunk is trivial, but the diff is large. Split it if review proves hard: move first, cut edges second. They are together here only because the fence in #1732 needs both.

### Goal

The conversation layer lives in `src/shared/`. It imports nothing from elsewhere in `src/`. It holds no React, no Lit, and no import that registers a custom element.

### Acceptance criteria

- [ ] `services/`, `store/`, `instance/`, `events/`, and `schema/` have moved out of `chat/` and into `src/shared/`.
- [ ] The utilities those files import have moved with them.
- [ ] The public types under `src/types/` have moved into `src/shared/`.
- [ ] Nothing in `src/shared/` imports from elsewhere in `src/`. A new spec walks the folder and fails on any such import.
- [ ] `carbonIcon.ts` has moved to `src/react/`, where React helpers live. Every importer still resolves.
- [ ] `domUtils.ts` imports nothing from `react`. Structural types stand in for `RefObject` and React's `KeyboardEvent`.
- [ ] `miscUtils.ts` imports nothing from `react`. A structural type stands in for `ErrorInfo`.
- [ ] `ServiceManager.ts` gets its three view-handle types from a new `viewHandles.ts` in `src/shared/`.
- [ ] The three view files that owned those types re-export them. No import path changes elsewhere.
- [ ] `MessageTypeComponentProps.d.ts` no longer imports the messages component or the service-manager wrapper. This is ADR-0002's fourth cross edge.
[...]

PR #2313, opened 2026-09-09
docs(adr): correct citations, lead with the decision, tighten prose

ethanwinters:
Closes #

Nine records go to final review carrying citations that no longer resolve, two counts that reproduce as different numbers, and prose that runs over the repo's grade-10 ceiling. One of them buries its own decision under 1179 words of background.

Corrections and wording only. No decision, option, or verdict changes anywhere.

Four commits, and they are the unit to review:

1. **Stale citations and counts** — factual fixes across eight records.
2. **ADR-0009 context** — the only argument that moved. It argued from a `send` contract "wrong in three places for as long as it has been published". That was fixed after the record was drafted. Reframed to ratify what the source now asserts.
3. **Reading pass on the six SDK records** — wording only.
4. **Structure** — ADR-0002's decision material moves out of context; four records gain an opening that says what is at stake.

Commits 3 and 4 were verified against the commit before them, not by eye: same options, verdicts, counts, citations, code blocks, tables, and heading set.

#### Changelog

**Changed**

- ADR-0002 — "Two layers, three ways in" and "Source tree layout" move from context into the decision outcome. They describe the three directories and the import rules, which is what the record decides. Context drops from 1179 words to 618, against 1245 for the decision. The record also gains an opening; it previously began on a subheading.
- ADR-0005, ADR-0023, ADR-0025 — each gains one opening sentence naming what is at risk, ahead of the link to the sibling record.
- ADR-0009 — Context rewritten. Verb placement restated: only `restartConversation` has two homes, `send` exists at the root alone.
- ADR-0002 — `@lit/react` wrappers: 42 sites → 28 across 24 files. 42 is the count of components-package React-barrel imports, a different thing.
- ADR-0023 — callback figures 410/14/16 → 518 messaging calls against 121 view-member sites, with the counting rule stated.
- ADR-0007 — "every reset outside the chunk path is guarded on `streamingMessageID`" was false. Four are unguarded. The conclusion holds: none is reachable from the upsert delivery path.
- ADR-0003 — `ChatContainer.tsx:378` → `:297`. The file is 317 lines.
- ADR-0004 — the eight `deepFreeze` sites and the `history` spread range re-cited. One of the two examples does not seed from `getState()`.
- ADR-0005, ADR-0025 — drifted line numbers.
- All nine records now score 6.9 to 9.6, from 9.3–11.3.

#### Testing / Reviewing

1. `npm run validate:adrs` exits 0.
2. Spot-check any corrected citation: open the file at the cited line, confirm it holds what the record says.
3. Confirm ADR-0009's premise really is stale. `grep -rn "gives up retrying\|too many errors" packages/ai-chat/src` returns nothing, and `types/instance/ChatInstance.ts:78-109` carries a **Settlement points** block.
4. Confirm the two replaced counts. `grep -rho "createComponent(" packages/ai-chat/src --include='*.ts' --include='*.tsx' | wc -l` gives 28.
5. For commits 3 and 4, check that nothing but wording and section order moved. No option verdict, count, citation, code block, or table should differ.

#### One open question for the review

ADR-0002's decision outcome still reads "This record decides only that the SDK is an entry point in this package, and when it may appear." With the source-tree layout now above that line, it under-claims: the record also decides the three-directory layering, and #1731 is already building against it. Left exactly as written rather than quietly widen what an unaccepted record claims. It needs a deliberate call.


davidmenendez (approved):
LGTM!

PR #2316, opened 2026-09-09
test(markdown): capture addMarkdown return values in projection test

davidmenendez:
The `projects exactly one host into each markdown element` test in `markdown.test.ts` referenced `elA` and `elB` without ever assigning them, causing a `ReferenceError` at runtime and failing CI on the release branch.

The return values of the two `addMarkdown()` calls were already used as the element references for the slot assertions — they just weren't captured. Assigning them to `const elA` and `const elB` is the complete fix.

This was introduced by the cherry-pick of `d5e90a24` (`fix(markdown): restore consumer CSS and fix line breaks #2299`) onto `release/v1.20.0`.


warrenmblood (approved):
LGTM

PR #2320, opened 2026-09-09
fix: add isSelected support to toolbar

### LINKED ISSUE #1963: Add a pressed/selected state to ToolbarAction
## Background

`ToolbarAction` (public API, exported from `packages/ai-chat/src/aiChatEntry.tsx:151`) has no way to express a toggle's on/off state. Its full field set comes from `Action` → `BaseOverflowMenuItem`: `icon`, `size`, `fixed`, `text`, `onClick`, `href`, `target`, `disabled`, `danger`, `dangerDescription`, `divider`, `testId`. Nothing carries selected/pressed.

This surfaced while reviewing the new `prompt-line-conversation-starters` examples in #1936, which use a single `input.actions` button to toggle `StartersConfig.isOn`. Because there is no pressed state, the only signal of whether starters are currently on or off is the tooltip text flipping between "Hide conversation starters" and "Show conversation starters".

That is not sufficient:

- **Sighted users** must hover the button to discover state — nothing in the resting visual conveys it. Carbon's ghost icon button renders identically in both states.
- **Screen reader users** get no toggle semantics at all. Verified on the running example: the rendered button has `aria-pressed = null`. It is announced as a plain button, so its state is not conveyed by name, role, or value.

This affects any consumer building a toggle out of `input.actions` or `header.actions`, not just the starters example.

## Goal

`ToolbarAction` can express a two-state toggle, and `cds-aichat-toolbar` renders that state with both an accessible name/role/state and a visible resting affordance — in the inline icon-button path and the overflow-menu path.

## Acceptance criteria

- [ ] `ToolbarAction` accepts an optional toggle-state field; omitting it leaves today's plain-button behavior byte-for-byte unchanged.
- [ ] `renderIconButton` (`packages/ai-chat-components/src/components/toolbar/src/toolbar.ts:236-259`) emits `aria-pressed="true"|"false"` on `<cds-icon-button>` when the field is present, and no `aria-pressed` attribute at all when it is absent.
- [ ] The overflow-menu render path (same file, ~line 336+) conveys the same state, since a `fixed: false` action can collapse into the menu.
- [ ] The pressed state has a **visible** resting treatment distinct from hover/focus, using Carbon tokens — not tooltip text alone. Coordinate with design; this may need `status: design needed`.
- [ ] Contrast of the pressed treatment against its background meets WCAG 2.1 AA (3:1 for the non-text state indicator).
- [ ] Keyboard: the toggle is reachable, operable with Enter and Space, and the state change is perceivable without hovering.
- [ ] WTR test in `packages/ai-chat-components/src/components/toolbar/__tests__/` covers pressed / not-pressed / field-absent, for both the inline and overflow paths.
- [ ] JSDoc on the new field follows `packages/ai-chat/src/types/AGENTS.md`, with a `@category` value drawn from `categoryOrder` in `packages/ai-chat/typedoc.json`.
- [ ] `npm run custom-elements` rerun and the `custom-elements.json` diff inspected.
- [ ] Verified with NVDA and VoiceOver — the toggle announces its state on focus and announces the change on activation.

## Public API / contract

Additive and optional, so no semver break. Proposed shape on `ToolbarAction` (`packages/ai-chat-components/src/react/toolbar.tsx:27`, flowing through `Action` in `packages/ai-chat-components/src/components/toolbar/src/toolbar.ts:39`):

```ts
/**
 * Marks this action as a two-state toggle and sets its current state.
 *
 * When defined, the action renders with `aria-pressed` and a visible
 * pressed treatment. When omitted, the action renders as a plain
 * button with no toggle semantics.
 */
isSelected?: boolean;
```

`undefined` (plain button) must stay distinct from `false` (toggle, currently off) — a `boolean` with a `false` default would make the two indistinguishable and force `aria-pressed="false"` onto every existing action.

Open question for the assignee: whether `text` should stay a single string that consumers swap per state (as the starters example does today), or whether the toggle needs a second label field so the accessible name can stay stable while `aria-pressed` carries the state. The latter is generally the better a11y pattern — a name that changes underneath a user is disorienting — but it widens the contract.

## Out of scope

- Radio-style / single-select action groups. This issue covers independent two-state toggles only.
- Changing `StartersConfig.isOn` or the `prompt-line-conversation-starters` examples. Once this lands, the example should adopt it in a follow-up.
- The unrelated a11y wiring gap in `cds-aichat-autocomplete` (separate issue).

## Related

- PR #1936 — the review that surfaced this.
- `packages/ai-chat-components/src/components/toolbar/src/toolbar.ts` — render paths.
- `packages/ai-chat-components/src/typings/overflow-menu.ts` — `BaseOverflowMenuItem`.
- `references/accessibility.md` — repo WCAG 2.1 AA checklist.

davidmenendez:
Closes #1963 

adds `isSelected` support in the `toolbar` component so users can enable buttons in the toolbar to have a selected state

**New**

- `ToolbarAction` / `Action` accepts an optional `isSelected?: boolean` field. When present, the action renders as a two-state toggle. Omitting it leaves existing plain-button behaviour unchanged — `undefined` is intentionally distinct from `false`.

**Changed**

- `cds-aichat-toolbar` `renderIconButton`: when `isSelected` is defined, sets `isSelected` on `<cds-icon-button>` so `@carbon/web-components` emits `aria-pressed="true"` on the inner `<button>` (requires `@carbon/web-components` ≥ 2.62.0, which ships the fix from [carbon-design-system/carbon#23009](https://github.com/carbon-design-system/carbon/pull/23009)). Sets `aria-pressed="false"` on the host element as a fallback for the toggle-off state, since upstream does not emit it.
- Overflow menu render path: overflow items now carry `aria-pressed` and `data-selected` to match the inline button path, so a `fixed: false` toggle action conveys the same state whether it renders inline or in the menu.
- `toolbar.scss`: added `cds-icon-button[data-selected]` and `cds-overflow-menu-item[data-selected]` rules using `theme.$layer-selected` / `theme.$text-primary` to give the pressed state a visible resting treatment distinct from hover and focus.
- `ToolbarAction` JSDoc in `HeaderConfig.ts` updated to document the toggle semantics and the `undefined`-vs-`false` distinction.

**Testing**

- WTR tests added to `toolbar/__tests__/toolbar.test.ts` covering: field-absent (no `aria-pressed`), `isSelected: true` (host `isselected` attribute set, `data-selected` present), `isSelected: false` (`aria-pressed="false"` fallback on host), state update (toggling the actions prop re-renders correctly), and overflow-menu equivalents of all three cases.
- Storybook: new **"Toggle action (isSelected)"** story in both the Lit (`toolbar.stories.js`) and React (`toolbar-react.stories.jsx`) Storybooks with a live click-to-toggle demo. A **"Toggle actions"** preset added to the Default story's actions dropdown.


AlexanderMelox:
Small nit from my end, since we are showcasing a "toggle" state maybe the example on storybook should toggle between star filled and not filled?

Screenshot 2026-09-10 at 8 50 41 AM

Toggling between the storybook actions doesn't do anything anymore maybe we set it as read only for the isSelected state or hide it
Screenshot 2026-09-10 at 8 52 24 AM

packages/ai-chat-components/src/components/toolbar/__stories__/story-data-react.js
```diff
@@ -77,4 +78,23 @@ export const actionLists = {
     },
   ],
   None: [],
+  'Toggle actions': [
+    {
+      text: 'Favourite (off)',
+      icon: StarFilled,
```
AlexanderMelox, line 84: Can we toggle between star and starfilled with its toggled on and off?

packages/ai-chat-components/src/components/toolbar/__stories__/story-data.js
```diff
@@ -88,4 +89,26 @@ export const actionLists = {
     },
   ],
   None: [],
+  'Toggle actions': [
+    {
+      text: 'Favourite (off)',
+      icon: StarFilled16,
```
AlexanderMelox, line 95: same here?

packages/ai-chat-components/src/components/toolbar/__stories__/toolbar-react.stories.jsx
```diff
@@ -223,3 +225,44 @@ export const Default = {
...
   },
 };
+
+const ToggleDemo = () => {
+  const [isOn, setIsOn] = useState(false);
+
+  const actions = [
+    {
+      text: isOn ? 'Favourite (on)' : 'Favourite (off)',
+      icon: StarFilled,
```
AlexanderMelox, line 235: can we add a toggle between star and star filled here also?

packages/ai-chat-components/src/components/toolbar/src/toolbar.ts
```diff
@@ -377,23 +386,32 @@ class CDSAIChatToolbar extends LitElement {
...
-                              danger-description=${
-                                item.dangerDescription || nothing
-                              }
-                              ?divider=${item.divider}
-                              data-testid=${item.testId || nothing}>
-                              ${item.text}
-                            </cds-overflow-menu-item>
-                          `
+                          (item) => {
+                            const { isSelected } = item;
```
AlexanderMelox, line 390: For the sake of consistency maybe we don't destructure this since we don't destructure the other props and just adds an extra line
```suggestion
                          (item) => {
```

packages/ai-chat-components/src/components/toolbar/src/toolbar.ts
```diff
@@ -377,23 +386,32 @@ class CDSAIChatToolbar extends LitElement {
...
+                                target=${
+                                  item.href ? item.target || '_self' : nothing
+                                }
+                                ?disabled=${item.disabled}
+                                ?danger=${item.danger}
+                                danger-description=${
+                                  item.dangerDescription || nothing
+                                }
+                                ?divider=${item.divider}
+                                ?data-selected=${isSelected === true}
```
AlexanderMelox, line 404:
```suggestion
                                ?data-selected=${item.isSelected === true}
```

packages/ai-chat-components/src/components/toolbar/src/toolbar.ts
```diff
@@ -377,23 +386,32 @@ class CDSAIChatToolbar extends LitElement {
...
+                                aria-pressed=${
+                                  isSelected !== undefined
+                                    ? String(isSelected)
+                                    : nothing
```
AlexanderMelox, line 408:
```suggestion
                                  item.isSelected !== undefined
                                    ? String(item.isSelected)
                                    : nothing
```

AlexanderMelox (changes requested):

packages/ai-chat-components/src/components/toolbar/src/toolbar.ts
```diff
@@ -270,6 +277,8 @@ class CDSAIChatToolbar extends LitElement {
         kind="ghost"
         enter-delay-ms="0"
         leave-delay-ms="0"
+        ?isSelected=${isSelected === true}
+        aria-pressed=${isSelected === false ? 'false' : nothing}
```
AlexanderMelox, line 281: **Blocker** — the off-state never announces. `aria-pressed="false"` goes on the `cds-icon-button` *host*, which has no role (nothing in Carbon 2.62's `button.js`, `icon-button.js`, or `focus.js` sets one), so it resolves to `generic` — and axe's `allowedAttr('generic')` is exactly the 24 global attributes, which `aria-pressed` is not (`getGlobalAriaAttrs()` confirms it isn't global). A screen reader gets a plain button in the off-state, which is the defect #1963 was filed for. `references/accessibility.md:16` names this trap: "declare `role` and labels on the host element explicitly."

The attribute has to land on the shadow button — set it in `updated()`, the way Carbon itself sets `aria-label` (`icon-button.js` `updated()`):

```ts
this.shadowRoot?.querySelectorAll<HTMLElement>('cds-icon-button[data-pressed]')
  .forEach((el) => el.shadowRoot?.querySelector('button')
    ?.setAttribute('aria-pressed', el.getAttribute('data-pressed')!));
```

That also covers the on-state, which is broken differently: Carbon binds `?aria-pressed="${ariaPressed}"` (`button.js:161,185`) — a *boolean* attribute binding, so `toggleAttribute` gives the button `aria-pressed=""`. axe's `validateAttrValue` rejects the empty string (`"true"`/`"false"` pass, `""` fails), so the PR description's claim that it emits `aria-pressed="true"` isn't what ships. Worth an upstream issue against carbon#23009 either way.

packages/ai-chat-components/src/components/toolbar/src/toolbar.ts
```diff
@@ -377,23 +386,32 @@ class CDSAIChatToolbar extends LitElement {
...
+                                danger-description=${
+                                  item.dangerDescription || nothing
+                                }
+                                ?divider=${item.divider}
+                                ?data-selected=${isSelected === true}
+                                aria-pressed=${
+                                  isSelected !== undefined
+                                    ? String(isSelected)
+                                    : nothing
+                                }
```
AlexanderMelox, line 409: **Blocker** — the overflow path is invalid ARIA, so a `fixed: false` toggle conveys nothing once it collapses into the menu. `cds-overflow-menu-item` assigns itself `role="menuitem"` in `connectedCallback` (`overflow-menu-item.js:50`), and axe's `allowedAttr('menuitem')` contains neither `aria-pressed` nor `aria-checked`; a live axe run flags this exact pairing as "ARIA attribute is not allowed".

Use the menu-native toggle role instead — `requiredContext('menuitemcheckbox')` is menu/menubar/group and `cds-overflow-menu-body` sets `role="menu"` (`overflow-menu-body.js:126`), so it slots in cleanly, and Carbon only assigns its default role when one is absent:

```suggestion
                                role=${
                                  isSelected !== undefined
                                    ? 'menuitemcheckbox'
                                    : nothing
                                }
                                aria-checked=${
                                  isSelected !== undefined
                                    ? String(isSelected)
                                    : nothing
                                }
```

packages/ai-chat-components/src/components/toolbar/__tests__/toolbar.test.ts
```diff
@@ -65,4 +65,173 @@ describe('toolbar', function () {
...
+        el.shadowRoot!.querySelectorAll('cds-overflow-menu-item')
+      );
+      const toggleItem = items.find(
+        (item) => item.textContent?.trim() === 'Toggle'
+      );
+      if (toggleItem) {
+        expect(toggleItem.getAttribute('aria-pressed')).to.equal('false');
+        expect(toggleItem.hasAttribute('data-selected')).to.be.false;
+      }
+    });
```
AlexanderMelox, line 235: **Blocker** — all three overflow tests are unfailable, so the path the AC calls out has no real coverage. Both state cases wrap every assertion in `if (toggleItem)`, and the absent case `forEach`s a list that is empty: overflow items render only after the `ResizeObserver` sets `containerWidth` (`toolbar.ts:143-151`; `getActions()` returns empty while `container` has zero width), but `fixtureWithOverflow` awaits only `el.updateComplete` — and `@open-wc`'s `elementUpdated` awaits that promise alone, skipping its `nextFrame()` fallback, while ResizeObserver does not deliver on a microtask.

Drop the `if` guards, assert the item exists first, and wait for the menu:

```ts
await waitUntil(() =>
  el.shadowRoot!.querySelectorAll('cds-overflow-menu-item').length > 0
);
// ...
expect(toggleItem, 'overflow item not rendered').to.exist;
```

packages/ai-chat-components/src/components/toolbar/src/toolbar.scss
```diff
@@ -108,4 +108,15 @@ $block-class: #{$prefix}-toolbar;
     min-block-size: layout.$spacing-08;
     min-inline-size: layout.$spacing-08;
   }
+
+  // Pressed / selected state for toggle actions
+  cds-icon-button[data-selected] {
+    background-color: theme.$layer-selected;
```
AlexanderMelox, line 114: **Blocker** — the pressed treatment misses the 3:1 non-text contrast #1963 requires, so the resting state stays invisible to the sighted users the issue was filed for. The inline button composites to `#cfcfcf` at **1.56:1** on `$background`; the overflow fill below is `$layer-selected` `#e0e0e0` on `$layer-01` `#f4f4f4` at **1.20:1**. It also lands 1.09–1.38:1 from Carbon's ghost hover (`$background-hover` over white is `#f1f1f1`), so it fails "distinct from hover/focus" too.

Take a treatment from design that clears 3:1 — a border or an icon-color shift carries it more cheaply than a fill — or amend #1963 with design's sign-off on the Carbon default. The issue tagged this AC `status: design needed`.

packages/ai-chat-components/src/components/toolbar/__tests__/toolbar.test.ts
```diff
@@ -65,4 +65,173 @@ describe('toolbar', function () {
...
+        html`<cds-aichat-toolbar .actions=${actions}></cds-aichat-toolbar>`
+      );
+      const btn = el.shadowRoot!.querySelector('cds-icon-button');
+      expect(btn).to.exist;
+      // Carbon handles aria-pressed="true" inside its own shadow DOM;
+      // we set the isSelected property so it does so.
+      expect(btn!.hasAttribute('isselected')).to.be.true;
+      expect(btn!.hasAttribute('data-selected')).to.be.true;
+      // No host-level aria-pressed="false" fallback needed for the on-state
+      expect(btn!.getAttribute('aria-pressed')).to.not.equal('false');
```
AlexanderMelox, line 106: **Important** — these assertions can't distinguish a working toggle from the broken one: they check only that Lit set `isselected` on the host, and `to.not.equal('false')` passes when the attribute is absent — exactly today's state. Assert what the AC names, on the element that carries it:

```ts
const inner = btn!.shadowRoot!.querySelector('button')!;
expect(inner.getAttribute('aria-pressed')).to.equal('true');
```

The `isSelected: false` test at line 141 needs the mirrored `'false'` assertion.

packages/ai-chat-components/src/components/toolbar/src/toolbar.scss
```diff
@@ -108,4 +108,15 @@ $block-class: #{$prefix}-toolbar;
...
+  }
+
+  // Pressed / selected state for overflow item
+  cds-overflow-menu-item[data-selected] {
+    background-color: theme.$layer-selected;
+    color: theme.$text-primary;
```
AlexanderMelox, line 120: **Important** — both new rules fight the shadow DOM they target. `color: theme.$text-primary` never applies at all: the label lives in `cds-overflow-menu-item`'s shadow root, where Carbon sets `color:var(--cds-text-secondary,#525252)` on `.cds--overflow-menu-options__btn` explicitly, which beats an inherited value. And on the inline button above, Carbon already paints the selected state (`.cds--btn--icon-only.cds--btn--selected{background:var(--cds-background-selected,hsla(0,0%,55.3%,.2))}`) — because that fill is translucent, the host fill stacks under it and renders `#cfcfcf`, neither token.

Style through the part Carbon exposes (`cds-icon-button::part(button)`), or let `cds--btn--selected` own the inline treatment and drop the host rule.

packages/ai-chat-components/src/components/toolbar/__stories__/toolbar.stories.js
```diff
@@ -224,3 +231,59 @@ export const Default = {
...
+        icon: StarFilled16,
+        size: 'md',
+        isSelected: isOn,
+        onClick: () => {
+          isOn = !isOn;
+          // Re-render by dispatching a custom event the story can react to
+          document
+            .querySelector('cds-aichat-toolbar')
+            ?.dispatchEvent(new CustomEvent('_story-rerender'));
+        },
```
AlexanderMelox, line 262: **Important** — `document.querySelector('cds-aichat-toolbar')` returns the *first* toolbar on the page, which in autodocs is the Default story's, so this story's `_story-rerender` listener never fires and the star stops toggling — in the one place a reviewer goes to see the feature work. `toolbar` is already in scope below; drop the event round-trip:

```js
onClick: () => {
  isOn = !isOn;
  toolbar.actions = getActions();
},
```

That makes the `addEventListener('_story-rerender')` block at line 283 dead too.

packages/ai-chat-components/src/components/toolbar/src/toolbar.ts
```diff
@@ -51,6 +51,11 @@ export interface Action extends BaseOverflowMenuItem {
    * When overflow handling is enabled, setting fixed to true will force this action out of the overflow menu.
    */
   fixed?: boolean;
+
+  /**
+   * Determines if a button has toggle functionality
+   */
```
AlexanderMelox, line 57: **Important** — this is the property doc that renders on the public docs site, not an internal note: `cross-package-types.md:37` says `@interface` resolves members so each carries "the upstream property's own JSDoc", meaning the paragraph you added at `HeaderConfig.ts:20-25` documents the interface, not this field. As written it's a fragment with no period that omits the `undefined`-vs-`false` distinction the whole design turns on, against the bar in `packages/ai-chat/src/types/AGENTS.md`.

```suggestion
  /**
   * Renders this action as a two-state toggle and sets its current state.
   * When `true` or `false`, the button exposes `aria-pressed` and a visible
   * pressed treatment. Omit the field for a plain button with no toggle
   * semantics — `undefined` is intentionally distinct from `false`.
   */
```

AlexanderMelox (commented):
Rework — the toggle's state is not exposed to assistive tech in either render path, and the overflow tests cannot fail.

Both `aria-pressed` placements sit on roles that do not support the attribute, checked against axe-core 4.10.2's ARIA standards tables (`getGlobalAriaAttrs()`, `allowedAttr()`) rather than inferred; the three overflow tests wrap every assertion in `if (toggleItem)`; the pressed treatment misses the 3:1 contrast AC by 2–3×. Verified against `@carbon/web-components` 2.62.0, which the lockfile resolves — not 2.61.0, so the on-state path itself is sound in CI.

**One finding has no diff line to sit on:**

**Important** — `packages/ai-chat-components/package.json:56` and `packages/ai-chat/package.json:129` — the lockfile resolves 2.62.0 so this repo exercises the fixed Carbon, but the declared ranges don't require it: `"@carbon/web-components": "^2.58.1"` and peer `>=2.54.0 <3.0.0` are both satisfied by 2.58–2.61, where `isSelected` exists but emits no `aria-pressed` at all. A consumer whose tree dedupes there gets a silent no-op with no warning. Raise the dependency to `^2.62.0` and the peer floor to `>=2.62.0 <3.0.0`.

### Test / verification gaps

- **The AC's assertion is untested in both paths** — no test reads `aria-pressed` off a rendered `<button>`, so the suite stays green while the feature emits nothing valid.
- **Keyboard AC uncovered** — nothing exercises Enter/Space, or checks the state change is perceivable without hover.
- **Screen-reader AC unclaimed** — #1963 requires NVDA and VoiceOver; the PR doesn't mention either.
- **Not run on my side** — the WTR suite and the lint gates. The contrast figures are computed from Carbon's compiled CSS and the white-theme token values, not sampled from a rendered Storybook.

### Questions

1. #1963's open question is still open: should `text` stay one string that swaps per state? The stories demo exactly the swapping the issue called disorienting (`Favourite (off)` → `Favourite (on)`) — the accessible name changes underneath the user, which is what `aria-pressed` exists to avoid.
2. Did the pressed treatment go through design, or is `$layer-selected` a placeholder? The issue flagged this AC as `status: design needed`.
3. Was the ≥2.62.0 floor left out of `package.json` deliberately — waiting on a separate Carbon bump?

packages/ai-chat-components/src/components/toolbar/src/toolbar.ts
AlexanderMelox, line 281: **Blocker** — the off-state never announces. `aria-pressed="false"` goes on the `cds-icon-button` *host*, which has no role (nothing in Carbon 2.62's `button.js`, `icon-button.js`, or `focus.js` sets one), so it resolves to `generic` — and axe's `allowedAttr('generic')` is exactly the 24 global attributes, which `aria-pressed` is not (`getGlobalAriaAttrs()` confirms it isn't global). A screen reader gets a plain button in the off-state, which is the defect #1963 was filed for. `references/accessibility.md:16` names this trap: "declare `role` and labels on the host element explicitly."

The attribute has to land on the shadow button — set it in `updated()`, the way Carbon itself sets `aria-label` (`icon-button.js` `updated()`):

```ts
this.shadowRoot?.querySelectorAll<HTMLElement>('cds-icon-button[data-pressed]')
  .forEach((el) => el.shadowRoot?.querySelector('button')
    ?.setAttribute('aria-pressed', el.getAttribute('data-pressed')!));
```

That also covers the on-state, which is broken differently: Carbon binds `?aria-pressed="${ariaPressed}"` (`button.js:161,185`) — a *boolean* attribute binding, so `toggleAttribute` gives the button `aria-pressed=""`. axe's `validateAttrValue` rejects the empty string (`"true"`/`"false"` pass, `""` fails), so the PR description's claim that it emits `aria-pressed="true"` isn't what ships. Worth an upstream issue against carbon#23009 either way.

packages/ai-chat-components/src/components/toolbar/src/toolbar.ts
AlexanderMelox, line 409: **Blocker** — the overflow path is invalid ARIA, so a `fixed: false` toggle conveys nothing once it collapses into the menu. `cds-overflow-menu-item` assigns itself `role="menuitem"` in `connectedCallback` (`overflow-menu-item.js:50`), and axe's `allowedAttr('menuitem')` contains neither `aria-pressed` nor `aria-checked`; a live axe run flags this exact pairing as "ARIA attribute is not allowed".

Use the menu-native toggle role instead — `requiredContext('menuitemcheckbox')` is menu/menubar/group and `cds-overflow-menu-body` sets `role="menu"` (`overflow-menu-body.js:126`), so it slots in cleanly, and Carbon only assigns its default role when one is absent:

```suggestion
                                role=${
                                  isSelected !== undefined
                                    ? 'menuitemcheckbox'
                                    : nothing
                                }
                                aria-checked=${
                                  isSelected !== undefined
                                    ? String(isSelected)
                                    : nothing
                                }
```

packages/ai-chat-components/src/components/toolbar/__tests__/toolbar.test.ts
AlexanderMelox, line 235: **Blocker** — all three overflow tests are unfailable, so the path the AC calls out has no real coverage. Both state cases wrap every assertion in `if (toggleItem)`, and the absent case `forEach`s a list that is empty: overflow items render only after the `ResizeObserver` sets `containerWidth` (`toolbar.ts:143-151`; `getActions()` returns empty while `container` has zero width), but `fixtureWithOverflow` awaits only `el.updateComplete` — and `@open-wc`'s `elementUpdated` awaits that promise alone, skipping its `nextFrame()` fallback, while ResizeObserver does not deliver on a microtask.

Drop the `if` guards, assert the item exists first, and wait for the menu:

```ts
await waitUntil(() =>
  el.shadowRoot!.querySelectorAll('cds-overflow-menu-item').length > 0
);
// ...
expect(toggleItem, 'overflow item not rendered').to.exist;
```

packages/ai-chat-components/src/components/toolbar/src/toolbar.scss
AlexanderMelox, line 114: **Blocker** — the pressed treatment misses the 3:1 non-text contrast #1963 requires, so the resting state stays invisible to the sighted users the issue was filed for. The inline button composites to `#cfcfcf` at **1.56:1** on `$background`; the overflow fill below is `$layer-selected` `#e0e0e0` on `$layer-01` `#f4f4f4` at **1.20:1**. It also lands 1.09–1.38:1 from Carbon's ghost hover (`$background-hover` over white is `#f1f1f1`), so it fails "distinct from hover/focus" too.

Take a treatment from design that clears 3:1 — a border or an icon-color shift carries it more cheaply than a fill — or amend #1963 with design's sign-off on the Carbon default. The issue tagged this AC `status: design needed`.

packages/ai-chat-components/src/components/toolbar/__tests__/toolbar.test.ts
AlexanderMelox, line 106: **Important** — these assertions can't distinguish a working toggle from the broken one: they check only that Lit set `isselected` on the host, and `to.not.equal('false')` passes when the attribute is absent — exactly today's state. Assert what the AC names, on the element that carries it:

```ts
const inner = btn!.shadowRoot!.querySelector('button')!;
expect(inner.getAttribute('aria-pressed')).to.equal('true');
```

The `isSelected: false` test at line 141 needs the mirrored `'false'` assertion.

packages/ai-chat-components/src/components/toolbar/src/toolbar.scss
AlexanderMelox, line 120: **Important** — both new rules fight the shadow DOM they target. `color: theme.$text-primary` never applies at all: the label lives in `cds-overflow-menu-item`'s shadow root, where Carbon sets `color:var(--cds-text-secondary,#525252)` on `.cds--overflow-menu-options__btn` explicitly, which beats an inherited value. And on the inline button above, Carbon already paints the selected state (`.cds--btn--icon-only.cds--btn--selected{background:var(--cds-background-selected,hsla(0,0%,55.3%,.2))}`) — because that fill is translucent, the host fill stacks under it and renders `#cfcfcf`, neither token.

Style through the part Carbon exposes (`cds-icon-button::part(button)`), or let `cds--btn--selected` own the inline treatment and drop the host rule.

packages/ai-chat-components/src/components/toolbar/__stories__/toolbar.stories.js
AlexanderMelox, line 262: **Important** — `document.querySelector('cds-aichat-toolbar')` returns the *first* toolbar on the page, which in autodocs is the Default story's, so this story's `_story-rerender` listener never fires and the star stops toggling — in the one place a reviewer goes to see the feature work. `toolbar` is already in scope below; drop the event round-trip:

```js
onClick: () => {
  isOn = !isOn;
  toolbar.actions = getActions();
},
```

That makes the `addEventListener('_story-rerender')` block at line 283 dead too.

packages/ai-chat-components/src/components/toolbar/src/toolbar.ts
AlexanderMelox, line 57: **Important** — this is the property doc that renders on the public docs site, not an internal note: `cross-package-types.md:37` says `@interface` resolves members so each carries "the upstream property's own JSDoc", meaning the paragraph you added at `HeaderConfig.ts:20-25` documents the interface, not this field. As written it's a fragment with no period that omits the `undefined`-vs-`false` distinction the whole design turns on, against the bar in `packages/ai-chat/src/types/AGENTS.md`.

```suggestion
  /**
   * Renders this action as a two-state toggle and sets its current state.
   * When `true` or `false`, the button exposes `aria-pressed` and a visible
   * pressed treatment. Omit the field for a plain button with no toggle
   * semantics — `undefined` is intentionally distinct from `false`.
   */
```

AlexanderMelox (commented):
Rework — the toggle's state is not exposed to assistive tech in either render path, and the overflow tests cannot fail.

Both `aria-pressed` placements sit on roles that do not support the attribute, checked against axe-core 4.10.2's ARIA standards tables (`getGlobalAriaAttrs()`, `allowedAttr()`) rather than inferred; the three overflow tests wrap every assertion in `if (toggleItem)`; the pressed treatment misses the 3:1 contrast AC by 2–3×. Verified against `@carbon/web-components` 2.62.0, which the lockfile resolves — not 2.61.0, so the on-state path itself is sound in CI.

**One finding has no diff line to sit on:**

**Important** — `packages/ai-chat-components/package.json:56` and `packages/ai-chat/package.json:129` — the lockfile resolves 2.62.0 so this repo exercises the fixed Carbon, but the declared ranges don't require it: `"@carbon/web-components": "^2.58.1"` and peer `>=2.54.0 <3.0.0` are both satisfied by 2.58–2.61, where `isSelected` exists but emits no `aria-pressed` at all. A consumer whose tree dedupes there gets a silent no-op with no warning. Raise the dependency to `^2.62.0` and the peer floor to `>=2.62.0 <3.0.0`.

### Test / verification gaps

- **The AC's assertion is untested in both paths** — no test reads `aria-pressed` off a rendered `<button>`, so the suite stays green while the feature emits nothing valid.
- **Keyboard AC uncovered** — nothing exercises Enter/Space, or checks the state change is perceivable without hover.
- **Screen-reader AC unclaimed** — #1963 requires NVDA and VoiceOver; the PR doesn't mention either.
- **Not run on my side** — the WTR suite and the lint gates. The contrast figures are computed from Carbon's compiled CSS and the white-theme token values, not sampled from a rendered Storybook.

### Questions

1. #1963's open question is still open: should `text` stay one string that swaps per state? The stories demo exactly the swapping the issue called disorienting (`Favourite (off)` → `Favourite (on)`) — the accessible name changes underneath the user, which is what `aria-pressed` exists to avoid.
2. Did the pressed treatment go through design, or is `$layer-selected` a placeholder? The issue flagged this AC as `status: design needed`.
3. Was the ≥2.62.0 floor left out of `package.json` deliberately — waiting on a separate Carbon bump?

AlexanderMelox:
@davidmenendez I did a pass myself and got some help from bob to see any potential blockers or pitfalls. I refined them as much as I can but feel free to close any requests that seem ridiculous

davidmenendez:
@AlexanderMelox changes made 👍

davidmenendez:
I pulled out the custom `$layer-selected` fills because they were failing the 3:1 non-text contrast requirement from the AC — `#e0e0e0` on white only gets you to about 1.38:1. The button is currently just relying on Carbon's built-in `.cds--btn--selected` treatment, but that one has the same problem (it uses a translucent fill). The issue was already flagged `status: design needed` for this, so we need a designer to weigh in on what the selected state should actually look like. A border or an icon color change would probably be the easiest path to hitting 3:1.

AlexanderMelox:
@davidmenendez these changes looks good, just one last thing and it's good to go!

**In the overflow menu, a screen reader won't announce the toggle's state.**
When the toggle action collapses into the "…" menu, the code marks the menu item as "this is a checkbox, and it's checked." But the thing that actually receives keyboard focus is a <button> hidden one level deeper inside Carbon's shadow DOM — and that button carries none of that information. So the screen reader reads the focused button and says "Favourite, button," with no on/off state. That's the exact problem the issue was filed to fix, just relocated from the inline button (where it's now fixed) to the menu.

AlexanderMelox (approved):

PR #2331, opened 2026-09-11
fix(deps): make a clean install work on Apple Silicon Macs

### LINKED ISSUE #2330: Make a clean install work on Apple Silicon Macs
### Background

On an Apple Silicon Mac, installing dependencies does not get you a working checkout. The install reports success. The first build then fails on a missing binary. So does the next one, in another package, for another tool. Recovery takes several manual steps. The next install undoes them.

Two separate faults produce this.

**The lockfile records Linux binaries only.** Five tools ship one compiled binary per platform. For each, `package-lock.json` lists the `linux-x64` copy and nothing else. An install on macOS arm64 does what the lockfile says. So it skips every binary the machine can run. Each one then fails in a different place:

| Tool | Where it fails | Message |
| --- | --- | --- |
| rollup | `npm run build --workspace=@carbon/ai-chat` | `Cannot find module @rollup/rollup-darwin-arm64` |
| oxc-resolver | `npm run build --workspace=@carbon/ai-chat-components`, at `cem analyze` | `Cannot find native binding` |
| rolldown | `npm run test --workspace=@carbon/ai-chat-examples-demo`, starting the web server | `Cannot find module '@rolldown/binding-darwin-arm64'` |
| lightningcss | the same demo build | `Cannot find module '../lightningcss.darwin-arm64.node'` |
| esbuild | not observed to fail | — |

The esbuild binary is missing the same way. No gate failed because of it. It is listed here because the lockfile treats it the same.

**Nothing pins the compiler.** No manifest asks for `typescript` directly. Not the root, not `packages/ai-chat`. Three packages deeper in the tree ask for it as an open-ended peer: `@rollup/plugin-typescript` wants `>=3.7.0`, `@storybook/react` wants `>= 4.9.x`, `react-docgen-typescript` wants `>= 4.3.x`. npm installs that peer and picks the newest match. A plain `npm install` therefore puts TypeScript 6.0.3 at the root, hiding the 5.9.3 the lockfile pins. But `packages/ai-chat/tsconfig.json` sets `"moduleResolution": "node"`, and TypeScript 6 rejects it:

```
error TS5107: Option 'moduleResolution=node10' is deprecated and will stop
functioning in TypeScript 7.0.
```

The build dies there. `npm ci` restores 5.9.3. It cannot fix the binaries, though: they are absent from the lockfile, not dropped on the way in. And adding one with `npm install --no-save` re-runs resolution, which brings TypeScript 6 back. Fix either fault alone and the other returns.

Seen on Node 24.14.0 with npm 11.19.0. The repo's `.nvmrc` asks for 24.20.0.

Why the lockfile holds Linux entries only is not yet known. Rebuild it on macOS and read the diff to find out.

### Goal

One install on an Apple Silicon Mac gives a working tree. Both packages build, the tests run, and the installed TypeScript matches the lockfile.

### Acceptance criteria

- [ ] The lockfile records the macOS arm64 binary for rollup, esbuild, oxc-resolver, rolldown, and lightningcss, next to the Linux ones. Proof: `grep darwin-arm64 package-lock.json` returns an entry for each of the five.
- [ ] A clean install on macOS arm64 puts all five binaries on disk. Proof: run `rm -rf node_modules && npm ci`, then check that each of the five directories exists under `node_modules`.
- [ ] The root pins TypeScript to the version the lockfile carries, so npm cannot pick a newer major. Proof: run `npm install`, then `node -p "require('./node_modules/typescript/package.json').version"` prints the pinned version.
- [ ] `npm install` and `npm ci` leave the lockfile unchanged. Proof: `git diff --exit-code package-lock.json` after each.
- [ ] Linux installs still work, so CI is unaffected. Proof: CI on this branch is green.
- [ ] The full gate passes on macOS arm64 from a clean install. Proof: `npm run build`, `npm run ci-check`, `npm run test --workspace=@carbon/ai-chat`, and `npm run test --workspace=@carbon/ai-chat-examples-demo` all exit 0.

### Out of scope

- Moving `moduleResolution` off `node10`, or any other step toward TypeScript 6. That is its own task. This one keeps the compiler the build already expects.
- Windows and Linux arm64. Nobody has reported either. Adding platforms nobody runs makes the lockfile harder to keep honest.
- The gap between the running Node version and `.nvmrc`.

### Related

Found while rebasing #2329. A plain `npm install` broke the build in four places before the branch could be checked.

ethanwinters:
Closes #2330

A clean install on an Apple Silicon Mac now gives a working tree. Two faults had to go together: fix either alone and the other still breaks the build.

The lockfile diff is large but purely additive. Read it with `git diff --diff-algorithm=histogram` and it is 2,639 additions and nothing else. The default Myers algorithm renders the same diff as 7,657 insertions and 5,018 deletions. That is pairing noise, not churn — key order is unchanged and no version moves.

#### How this happened

**The platform binaries — a week old.** Each native tool ships one package per platform. Through v1.20.0 the lockfile held all of them. #2307, the Webpack to Vite migration, rewrote the lockfile whole: 1,928 insertions against 16,197 deletions. Only the `linux-x64-gnu` entries survived. That left 1 of 27 for rollup, 1 of 26 for esbuild, and the same shape for oxc-resolver, rolldown, lightningcss, and fsevents. 123 platform entries went and none came back.

The musl variants went with the darwin ones. That is the tell: the lockfile was rebuilt on a Linux x64 glibc machine, with optional dependencies pruned to that host. Nothing in the repo asked for that. There is no `.npmrc`, and no `omit=optional` anywhere in tracked files.

CI could not catch it. All 24 jobs run `ubuntu-latest` and install with `npm ci`. A lockfile holding only `linux-x64-gnu` is complete for that. No job runs macOS or Windows.

The Mend `chore(deps): lock file maintenance` commits are not involved. The last one before the loss, #2298 on Sep 8, still carried every darwin entry.

**The compiler — older, and latent.** No manifest asked for `typescript` directly. So npm resolved it from open-ended peer ranges, `>=3.7.0` from `@rollup/plugin-typescript` and `>= 4.9.x` from `@storybook/react`, and took the newest major. TypeScript 6.0.2 shipped 2026-03-23, so `npm install` could hoist a 6.x for six months. `npm ci` pins 5.9.3 from the lockfile, so nobody hit it. The missing binaries are what pushed people to `npm install`, where TypeScript 6 rejects `moduleResolution: node` in `packages/ai-chat`.

#### Changelog

**New**

- root pins `typescript` to `5.9.3`, exact, so npm cannot reach a newer major

**Changed**

- lockfile records the full platform matrix for rollup, esbuild, oxc-resolver, rolldown, lightningcss, and fsevents — 157 entries added, no version changed

#### Testing / Reviewing

On macOS arm64, from a clean tree:

1. `grep -c darwin-arm64 package-lock.json` — 38, up from 22 on `main`.
2. `rm -rf node_modules && npm ci`, then check the five directories exist under `node_modules`. All five land, plus three nested copies and `fsevents`.
3. `node -p "require('./node_modules/typescript/package.json').version"` — `5.9.3`.
4. `npm install`, then `git diff --exit-code package-lock.json`. Clean, and TypeScript stays 5.9.3.
5. `npm run build` — 89 projects.
6. `npm run test --workspace=@carbon/ai-chat` — 1275 tests, 118 suites.
7. `npm run test --workspace=@carbon/ai-chat-examples-demo` — 44 Playwright tests.

Linux is unaffected by construction. Filtering both lockfiles by `os`, `cpu`, and `libc` for linux/x64/glibc gives the same 2,773 packages before and after, nothing gained or lost. Every linux entry added is `libc: ["musl"]`, which a glibc runner skips. CI on this branch is the real proof.

`npm run ci-check` fails before it reaches the tests, on pre-existing skill-mirror drift in `.agents/skills/` — four files differ from `.bob/skills/`. That is on `main` already and unrelated to this branch. Every stage before it passes, and the two stages after it were run directly and pass.

`npm run measure -- --changed main` is clean: no new code files, no added lines, one new dependency.


warrenmblood (approved):
LGTM

Issue #1424, opened 2026-05-21
Roll out Playwright tests to all remaining examples


ethanwinters:
## Background

Parent epic: #1421.

#1422 and #1423 deliver three golden examples and an `examples/references/playwright.md` hardened against a trivial case (`basic-custom-element-fullscreen`), a rich-interaction one (`prompt-line-mentions-and-commands`), and a launcher plus host-state one (`watch-state`). This issue applies that guide to every remaining deterministic example so the whole catalog has per-example end-to-end coverage.

Counts below were verified against `main` on 2026-08-11.

## Goal

Every deterministic example has a Playwright test covering its unique functionality, generated by following the guide. The guide gets refined whenever an example exposes a gap. As the closing step, `examples/AGENTS.md` is updated to *require* a Playwright test for every example.

## Scope of the rollout

88 examples exist — 47 React, 41 web-components. Web-components is a strict subset: all 41 of its slugs also exist in React, and 6 are React-only (`frameworks-next`, `frameworks-react-17`, `frameworks-vite`, `tests-jest-happydom`, `tests-jest-jsdom`, `watch-state-redux`).

Subtract the 6 test setups delivered by #1422 and #1423, the 2 skipped `integrations-watsonx` setups, and the 2 skipped `tests-jest-*` setups, and this issue covers **78 remaining test setups** — 74 across 37 shared slugs, plus 4 React-only.

The catalog is actively growing (three slug pairs landed in the last month, one pair was deleted). Re-derive these counts at the start of the rollout rather than trusting the number above; the tightened Definition of done keeps new examples in step once it lands.

### React and web-component coverage are not symmetric

#1941 rebases `ChatContainer` and `ChatCustomElement` onto the web-component chain, so once it lands the chat DOM is identical across flavors — but it replaces the old duplication with a new React-only bridge layer: four render props portaling consumer ReactNodes into web-component-managed hosts, with microtask-deferred state updates and stable per-slot host elements. That bridge is where React-specific breakage will live, and #1941's own acceptance leans on React examples smoke-testing clean. So React coverage is kept, but weighted:

**Tier A — full behavioral spec in both flavors.** The 16 shared slugs whose demonstrated concern runs through React-specific code, plus the 4 React-only slugs, which have no web-component twin at all.

[...]

Issue #1421, opened 2026-05-21
Epic: Playwright testing for examples


ethanwinters:
## Value proposition

The examples catalog is how developers learn Carbon AI Chat, and it has almost no end-to-end coverage. There are 88 examples — 47 React, 41 web-components — sharing exactly one Playwright setup, in `examples/react/frameworks-react-17/`. A broken example ships silently, and the first person to find out is a developer copying it.

This epic gives every deterministic example a Playwright test that exercises the one thing that example demonstrates, governed by a guide terse enough for a small-context agent to follow unattended.

## Expected outcomes

- [ ] Every deterministic, servable example has a Playwright test covering its unique functionality, in the flavor(s) where that functionality can break.
- [ ] `examples/references/playwright.md` is a complete authoring guide an agent can follow without supervision.
- [ ] The "Definition of done" in `examples/AGENTS.md` *requires* a Playwright test for every example, carving out the four deliberately skipped.
- [ ] The example suite runs in CI on a decided cadence, in a documented shape that holds at ~78 suites, not just at three.

## Out of scope

- **Chromatic snapshots.** The guide today is one sentence: "All examples will eventually require playwright tests with Chromatic snapshots. For now, they can be skipped." That Chromatic requirement is dropped. This epic delivers functional Playwright tests only, and #1422 removes the sentence when it rewrites the guide. Visual regression is a separate decision, not a prerequisite here.
- `integrations-watsonx` — non-deterministic (live watsonx.ai, through a token proxy). Skipped in both flavors.
- `tests-jest-happydom` and `tests-jest-jsdom` — the only two examples with no `start` script, no `build` script, no bundler config, and no HTML entry. There is no page for Playwright to open, and giving them a dev server would give each a second concern. Their Jest spec is their coverage.
- The `demo/` Playwright suite — unchanged. Its helpers are prior art to copy from, not a target.
- Unit, Jest, and vitest test changes beyond two things: reconciling smoke-test guidance, and renaming the `test` script where Playwright has to coexist with a runner an example already owns.

## Details

The sub-issue list is the child record, so it is not restated here.

[...]

PR #2341, opened 2026-09-14
test(examples): add playwright goldens and authoring guide

### LINKED ISSUE #1422: Golden Playwright example: basic-custom-element-fullscreen
## Background

Parent epic: #1421.

`examples/references/playwright.md` is one sentence today, and only `examples/react/frameworks-react-17/` has a Playwright setup. Before tests roll out across 88 examples (47 React, 41 web-components), the repo needs one canonical, copy-pasteable reference per flavor and a guide an agent can follow without supervision.

`basic-custom-element-fullscreen` is the right first target. It's the canonical non-float baseline that `examples/AGENTS.md` names in its base-template rule — `ChatCustomElement` / `<cds-aichat-custom-element>` + `layout.showFrame: false` + `openChatByDefault: true`. The chat is open on load, so there's no launcher step; the mock `customSendMessage` echoes input deterministically; and the React and web-component versions are structurally parallel. A feature-rich golden example is deliberately deferred to #1423.

This issue also settles two conventions the rollout can't discover on its own: what the test script is called, and how each example gets a port to serve on.

## Goal

A golden Playwright test for `basic-custom-element-fullscreen` in both flavors, plus a complete `examples/references/playwright.md` that contributors and agents copy from.

### 1. Settle the `test` script name

Three examples already own a `test` script: `frameworks-vite` runs `vitest run`, and `tests-jest-happydom` / `tests-jest-jsdom` run `jest`. The rollout adds a Playwright script to ~78 more, so the naming has to be decided here rather than discovered mid-rollout when the first collision clobbers a pre-existing suite.

Pick one and record it in the guide:

- Playwright lands on `test:e2e` everywhere, leaving `test` to whatever unit runner an example has, or
- `test` runs Playwright, and the three examples above get a composite that runs both runners.

Whichever wins, `npm run test --workspace=frameworks-vite` and both `tests-jest-*` workspaces must still run their existing suites afterward.

### 2. Settle port allocation — programmatically, not by table

All 84 webpack examples resolve `const port = process.env.PORT || 3000` and hand it to `devServer.port`. So `PORT=<port> npm run start` works everywhere, and every example collides on 3000 if nothing sets it.

What that needs is a unique port **at run time**. It does not need a checked-in registry: an 88-row table every new-example PR has to edit is a permanent maintenance tax and a merge-conflict magnet, and it grows with the catalog forever. Do not introduce one.

Pick a programmatic mechanism and implement it in the shared config template:

- **Probe a free port at config load.** Playwright config files are ESM, so top-level `await` works — find an open port, feed it to both `webServer.command` and `baseURL`. Note the small window between probing and webpack binding, and say whether it matters at the concurrency #2127 lands on.
- **Derive it from the workspace name**, with a CI check asserting uniqueness. A generated table is fine; a hand-edited one is not.

Two examples can't take an outer `PORT=` at all: `frameworks-vite` (`cross-env PORT=3016 vite`) and `frameworks-next` (`cross-env PORT=3018 next start`) pin it from inside the start script, and `frameworks-next`'s `start` is the production server. Whatever mechanism wins has to say what those two do instead.

### 3. React golden — `examples/react/basic-custom-element-fullscreen/`

- Add `playwright.config.ts` modeled on `examples/react/frameworks-react-17/playwright.config.ts`: `testDir: "./tests"`, 60s timeout, `retries: process.env.CI ? 1 : 0`, `webServer` auto-start (`PORT=<port> npm run start`, `reuseExistingServer: !process.env.CI`), **chromium only**, and the port from task 2.
- Set `workers: 1`. Each example has one or two specs, so a per-example worker pool buys nothing and multiplies against whatever concurrency runs the examples themselves. #2127 may tune this; the template should not inherit two defaults that multiply.
- Add a `tests/` spec that asserts the chat mounts with no console errors, sends one message through the mock backend, and asserts the echoed response.
- Select with Playwright's recommended priority: `getByRole` / `getByLabel` with a stable accessible name first, falling back to `PageObjectId` test IDs (`getByTestId`, from `@carbon/ai-chat/server`) where there's no reliable role or name. No CSS or structural selectors.
- Add the test script under the name settled in task 1; add `@playwright/test` to `devDependencies`, matching the `frameworks-react-17` version.
- Add top-of-file purpose comments per the inline-comments rule in `examples/AGENTS.md`.

Illustrative shape — final code is the implementer's:

```ts
import { test, expect } from "@playwright/test";
import { PageObjectId } from "@carbon/ai-chat/server";

test("opens fullscreen and echoes one message", async ({ page }) => {
  const errors: string[] = [];
  page.on("console", (m) => m.type() === "error" && errors.push(m.text()));

  await page.goto("/");

  // The input is a contenteditable with no reliable ARIA role/name, so use the
  // maintained PageObjectId test-id contract here.
  const input = page.getByTestId(PageObjectId.INPUT);
  await expect(input).toBeVisible({ timeout: 10000 });

  await input.click();
  await input.fill("hello");

  // The send button has a stable accessible name — prefer a user-facing locator.
  await page.getByRole("button", { name: /send/i }).click();
  await expect(page.getByText("hello")).toBeVisible();

  expect(errors).toEqual([]);
});
```

### 4. Web-component golden — `examples/web-components/basic-custom-element-fullscreen/`

Same deliverables. The chat UI inside `<cds-aichat-custom-element>` renders identically to the React host, so the selectors and spec body should be near-identical — diverge only in how the page mounts. This parity check decides task 7's single-file-vs-split question, and it is also the first evidence for the epic's decision that React and web-component coverage need not be symmetric.

Cover default hide-on-close here, in the React flavor. It is one shared behavior across every `ChatCustomElement` example, so covering it once in the canonical baseline is what lets #1424 leave it off its tiering axis. #1941 later moves that behavior onto the web-component path; this test is what proves it didn't change.

### 5. Confirm the goldens run in CI at all

Root `test` is `lerna run test --stream` and root workspaces include `examples/*/*`, so a test script added to an example is picked up by the existing `test` job automatically — confirm that actually happens, that `npx playwright install --with-deps` covers the new suites, and that the job stays green.

That is the whole CI obligation here. **Do not design the run policy from two examples.** #2127 owns when and how the suite runs, and it uses these goldens as the thing it measures.

### 6. Reconcile smoke-test guidance

`examples/AGENTS.md` says examples should have Playwright tests, but `examples/react/AGENTS.md` and `examples/web-components/AGENTS.md` still point their "Smoke-test setup" sections at Jest (`tests-jest-jsdom`, `tests-jest-happydom`) — and the web-component file points at a React example for its Jest wiring. Neither mentions the vitest setup in `frameworks-vite`. Reconcile all three: Playwright is the smoke-test mechanism going forward, the `tests-jest-*` examples stay framework-integration examples only, and `frameworks-vite`'s vitest suite is named for what it is rather than left undocumented.

Do **not** make a Playwright test a hard requirement in the `examples/AGENTS.md` Definition of done yet — most examples have no tests, so the requirement would fail repo-wide. That hardening is the closing step of #1424. Note the intent here so it's on record.

### 7. Rewrite `examples/references/playwright.md`

Replace the one-line file. **Drop the Chromatic requirement** — the current sentence mandates "playwright tests with Chromatic snapshots", and Chromatic is out of scope for this epic (see #1421). The rewritten guide covers functional Playwright tests only, with no visual-snapshot step.

After both goldens exist, decide the guide's location:

- If the React and web-component setups turn out **markedly different**, split into `examples/react/references/playwright.md` and `examples/web-components/references/playwright.md`, each linked from that flavor's `AGENTS.md`.
- Otherwise keep the single `examples/references/playwright.md` already linked from `examples/AGENTS.md`. This is the expected outcome — the setups share the webpack dev server, `@playwright/test`, and the same selector approach.

The guide must contain:

1. **Quick start** — "copy the golden example", linking the two concrete golden directories.
2. **`playwright.config.ts` conventions** — testDir, timeout, retries, `webServer` auto-start, chromium-only rationale, `workers: 1`, the script name settled in task 1, and the port mechanism settled in task 2. Explain *why* a port has to be supplied — every webpack example falls back to 3000 — and state plainly that ports are allocated programmatically and no table is to be introduced. A future contributor who hardcodes a port into a `webpack.config.js` breaks `PORT=` for that example; say so.
3. **Selector strategy** — prefer user-facing locators (`getByRole` with an accessible name, `getByLabel`) where the chat exposes a stable, correct accessible name; that doubles as a free a11y check. Fall back to `PageObjectId` test IDs where there's no reliable role or name — dynamic or localized text, the contenteditable input, or anywhere pinning to the library's maintained testing contract is safer. Avoid CSS and structural selectors, and don't match raw model-output text. Note that Playwright locators pierce *open* shadow DOM either way, so shadow DOM isn't the deciding factor; the real web-component caveat is that ARIA IDREFs (`aria-labelledby`, `for`) don't cross shadow-root boundaries, which makes accessible names unreliable and is part of why `PageObjectId` exists.
4. **What to test** — the example's *one* unique concern, per the single-purpose rule, plus a baseline: mounts, no console errors, one message round-trip.
5. **Determinism and what to skip** — disable animations, settle streaming before asserting, no real timestamps, how to recognize a non-deterministic example, and the explicit skips: `integrations-watsonx` (non-deterministic) and `tests-jest-happydom` / `tests-jest-jsdom` (nothing to serve). Name all four with their reasons so the next contributor doesn't re-litigate them.
6. **Examples that deviate** — `frameworks-vite` and `frameworks-next` pin their port inside the start script with `cross-env`, which overrides the `PORT=<port> npm run start` webServer command; `frameworks-next`'s `start` is the production server. Say what to do instead.
7. **Naming conventions** — spec filenames.
8. **Running** — locally. Leave CI cadence to #2127 and link it rather than guessing.
9. **Definition of done** for an individual example's Playwright tests.
10. **React vs. web-component deltas** — a short section, if the guide stays a single file. This is where the epic's asymmetry decision gets its operational form: what a mount-level React spec contains when the full behavioral spec lives on the web-component side.

### How to author the guide

`AGENTS.md` and its `references/` docs are not human documentation. They're terse instruction files an AI coding agent reads as context and then obeys. Write for a literal-minded teammate, not a tutorial reader.

- **The repo's own files are the style reference.** Read `examples/AGENTS.md`, `examples/react/AGENTS.md`, `examples/web-components/AGENTS.md`, `demo/AGENTS.md`, the root `AGENTS.md`, and `examples/references/indexer-contract.md`. Mirror their voice and structure — especially the "Authoring rules" + "Definition of done" shape of `examples/AGENTS.md`.
- **Keep it short.** This file gets loaded into IBM Bob, which has a small context window. Terse imperative bullets, no prose padding, one tight snippet over a paragraph, say each thing once. Anything that doesn't change what the agent does is cuttable.
- **Lead with "copy the golden example"** and link the concrete directories, so the file still works if a small-context agent reads only the top.
- **Make it self-contained.** #1423 and #1424 turn agents loose on this guide unattended, so every step needs explicit paths, unambiguous instructions, and explicit skip criteria. No "see the other example".
- **Cross-link, don't duplicate** — `examples/AGENTS.md`, both flavor `AGENTS.md` files, and `demo/tests/README.md` for existing Playwright helper patterns.

External references: the [AGENTS.md format](https://agents.md/), Anthropic's [memory docs](https://docs.anthropic.com/en/docs/claude-code/memory), and the `CLAUDE.md` section of [Claude Code best practices](https://www.anthropic.com/engineering/claude-code-best-practices).

## Acceptance criteria

- [ ] The `test` script convention is decided, recorded in the guide, and applied to both goldens.
- [ ] `frameworks-vite`'s vitest suite and both `tests-jest-*` suites still run and pass under that convention.
- [ ] Ports are allocated programmatically: both goldens run concurrently on distinct ports with nothing checked in listing them, and the guide contains no port table.
- [ ] Both goldens have `playwright.config.ts` with `workers: 1`, a `tests/` spec, and updated `package.json` (test script + devDeps).
- [ ] Each golden's suite passes locally on chromium, and `npm run build --workspace=<each golden>` exits 0.
- [ ] The React golden asserts default hide-on-close with no `onViewChange` supplied.
- [ ] Both golden suites run in CI on a green `main` run, with browsers installed and no added flakiness.
- [ ] `examples/references/playwright.md` is complete, carries no Chromatic requirement, and its location is decided with all cross-references updated to match.
- [ ] The guide names all four skipped examples with their reasons.
- [ ] Smoke-test guidance is consistent across `examples/AGENTS.md`, `examples/react/AGENTS.md`, and `examples/web-components/AGENTS.md`, and accounts for the vitest setup in `frameworks-vite`.
- [ ] The `examples/AGENTS.md` Definition of done is *not* yet hardened, and the intent to harden it in #1424 is recorded.
- [ ] `npm run ci-check` passes, including the `validate-agents-files` job.

## Out of scope

- Every other example — #1424.
- A complex or feature-rich golden — #1423.
- When and how the suite runs in CI at scale — #2127.
- Chromatic or any visual-snapshot step — dropped epic-wide, see #1421.
- `integrations-watsonx`, `tests-jest-happydom`, `tests-jest-jsdom` — see #1421.

## Related

Parent epic: #1421.

Blocks #1423, which dogfoods the guide this issue writes, and #2127, which measures the suites this issue produces.

AlexanderMelox:
Closes #1422

Two golden Playwright examples for `basic-custom-element-fullscreen`, React and Web Components, plus the authoring guide the rollout in #1424 turns agents loose on.

- `examples/*/basic-custom-element-fullscreen/playwright.config.ts` — byte-identical between flavors, so "copy the golden" is literal. The port probe publishes to the environment; read the comment before changing it.

#### Changelog

**New**

- Playwright smoke tests for `basic-custom-element-fullscreen`, React and Web Components.
- `examples/references/playwright.md` — authoring guide: golden examples to copy, port rule, selector order, the four deliberate skips, per-example definition of done.
- Root `test:e2e` script, and a CI step that runs it.

**Changed**

- Playwright takes `test:e2e`, never `test`. `frameworks-vite` and both `tests-jest-*` keep `test` for their own runner.
- Ports are probed at run time.
- Smoke-test guidance in the three `AGENTS.md` files points at Playwright instead of Jest, and names the vitest suite in `frameworks-vite`.

**Removed**

- The Chromatic requirement, dropped epic-wide in #1421.

#### Testing / Reviewing

Not reachable from the demo site — these are example suites.

1. `npm run test:e2e` from the root. Two projects, 5 tests, all chromium.
2. Both concurrently, on distinct ports: `npm run test:e2e` runs them under lerna. Nothing lists a port.
3. Confirm the hide-on-close test discriminates. Add `onViewChange: async () => {}` to `examples/react/basic-custom-element-fullscreen/src/App.tsx` and rerun — the third test must fail with `Expected: 0, Received: 1280`. Revert.
4. Confirm the pre-existing suites still pass: `npm run test --workspace=@carbon/ai-chat-examples-react-frameworks-vite`, and the same for both `tests-jest-*`.
5. `npm run build --workspace=<each golden>` exits 0.


examples/react/basic-custom-element-fullscreen/playwright.config.ts
```diff
@@ -0,0 +1,68 @@
...
+ * Each example's suite is its own process tree, so the environment variable
+ * below never leaks between examples running in parallel.
+ */
+
+import { defineConfig, devices } from '@playwright/test';
+import { createServer } from 'node:net';
+import type { AddressInfo } from 'node:net';
+
+/** Ask the OS for an unused port by binding port 0 and reading it back. */
+async function probeFreePort(): Promise<number> {
```
ethanwinters, line 23: This should probably be a shared utility of some kind rather than copied in each example over and over. Also worry about race conditions on this if we have a bunch of example test suites running concurrently.

examples/web-components/AGENTS.md
```diff
@@ -13,7 +13,9 @@ Workspace naming: `@carbon/ai-chat-examples-web-components-<slug>`.
 
 ## Smoke-test setup
 
-Reuse Jest wiring from [`../react/tests-jest-jsdom/`](../react/tests-jest-jsdom/) — nearest reference for workspace-dep resolution. A sufficient spec instantiates the custom element, asserts it upgrades without error, and does one round-trip behavior (e.g. send one message via `customSendMessage`).
+Playwright is the smoke-test mechanism. Copy [`./basic-custom-element-fullscreen/`](./basic-custom-element-fullscreen/) — it is the web-component golden. Conventions and steps: [playwright.md](../references/playwright.md).
+
+This tree has no Jest wiring and needs none. Jest and vitest integration are the subject of React examples, not a pattern to reproduce here.
```
ethanwinters, line 18: This is probably info we don't need.

.github/workflows/ci.yml
```diff
@@ -45,6 +45,11 @@ jobs:
         run: cd packages/ai-chat-components && npm run build
       - name: Run test
         run: npm run test
+      # Playwright suites live under `test:e2e` so they never displace the unit
+      # runner an example already owns, which means `npm run test` does not
+      # reach them.
+      - name: Run example e2e tests
```
ethanwinters, line 51: I suspect we should only run these when we cut releases.

examples/react/basic-custom-element-fullscreen/playwright.config.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// Playwright evaluates this config once in the runner and again in every
+// worker process. Probing per evaluation would hand each process a different
+// port, so the first probe is published to the environment the workers
+// inherit and every later evaluation reuses it.
+const PORT = Number(process.env.CAIC_EXAMPLE_PORT) || (await probeFreePort());
+process.env.CAIC_EXAMPLE_PORT = String(PORT);
+
+export default defineConfig({
```
ethanwinters, line 42: Similarly, lets consider putting this somewhere shared that maybe takes the testDir as an argument.

examples/react/AGENTS.md
```diff
@@ -13,9 +13,12 @@ Workspace naming: `@carbon/ai-chat-examples-react-<slug>`.
...
 ## Smoke-test setup
 
-Reference setups: [`./tests-jest-jsdom/`](./tests-jest-jsdom/) and [`./tests-jest-happydom/`](./tests-jest-happydom/) — `jest.config.js`, `babel.config.js`, plus a spec under `src/`. **Default to jsdom**; use happy-dom only when the example's APIs need it (layout measurement, certain form behaviors).
+Playwright is the smoke-test mechanism. Copy [`./basic-custom-element-fullscreen/`](./basic-custom-element-fullscreen/) — it is the React golden. Conventions and steps: [playwright.md](../references/playwright.md).
 
-A sufficient spec mounts root, asserts no error, and exercises the main claimed API (e.g. one message via `customSendMessage`).
+The other runners here are what their example demonstrates, not a pattern to copy:
+
+- [`./tests-jest-jsdom/`](./tests-jest-jsdom/) and [`./tests-jest-happydom/`](./tests-jest-happydom/) show Jest integration. Both keep `test` for Jest, and both are skipped for Playwright — they have no page to serve.
+- [`./frameworks-vite/`](./frameworks-vite/) shows Vite integration and keeps `test` for its vitest suite.
```
ethanwinters, line 21: Since all the examples are on vite now, we can probably just delete this example.

examples/references/playwright.md
```diff
@@ -1 +1,137 @@
...
+"devDependencies": { "@playwright/test": "^1.63.0" }
+```
+
+**Name the script `test:e2e`, not `test`.** Three examples already use `test` for a different test runner: `frameworks-vite` runs vitest, and `tests-jest-happydom` and `tests-jest-jsdom` run jest. Calling yours `test` would replace theirs.
+
+## Config conventions
+
+Copy [playwright.config.ts](../react/basic-custom-element-fullscreen/playwright.config.ts) as-is. What it sets, and why:
+
+| Setting | Value |
```
ethanwinters, line 21: If we do make the config shared from somewhere we can update this.

examples/references/playwright.md
```diff
@@ -1 +1,137 @@
...
+| `testDir` | `./tests` |
+| `timeout` | 60s per test |
+| `workers` | `1` — an example has one or two specs, so a pool buys nothing and multiplies against whatever concurrency runs the examples |
+| `retries` | `process.env.CI ? 1 : 0` |
+| `projects` | chromium only — webkit has shadow-DOM gaps, see [demo/playwright.config.ts](../../demo/playwright.config.ts) |
+| `webServer.command` | `PORT=<probed> npm run start` |
+| `reuseExistingServer` | `!process.env.CI` |
+
+### Ports are allocated at run time. Never add a port table.
+
```
ethanwinters, line 32: Generally concerned about race conditions here with ports being available when its checked but not available by the time we boot up the tests if these tests all get run concurrently.

examples/react/basic-custom-element-fullscreen/playwright.config.ts
AlexanderMelox: @ethanwinters should we run one at a time then?

examples/web-components/AGENTS.md
AlexanderMelox: Removed 👍🏼

.github/workflows/ci.yml
AlexanderMelox: Good call, I wasn't sure how often we wanted to run these. Should i move this action into the release actions?

AlexanderMelox: Removed from `ci.yml` i think we should run the e2e offline (during test fest) or only on releases

examples/react/basic-custom-element-fullscreen/playwright.config.ts
ethanwinters: I think the the shared utility probably should keep track of the ports in use. If we spin up all the tests at once with "check if port is available" we will certainly run into "it was available at the time of check, but then a test spun up and took it" if we run them all concurrently. 

We 1000% want to run them concurrently. I think we just need a safer way to assign ports.

ethanwinters: Like perhaps asking for the port needs to be in a queue and we don't return a port until we have returned the port to the last request first, etc..

Issue #2322, opened 2026-09-10
[A11y]: Screen reader does not properly read chat window content


SrfHead:
### Browser

Firefox

### Operating system

Windows

### Package version

1.15

### React version (if applicable)

18.2.0

### Automated testing tool and ruleset

Latest

### Assistive technology

Windows Narrator

### Description

Using the Windows Narrator screen reader does not give accept results. There is a written notification that describes navigating with Tab/Arrow keys or F6, which is only helpful if you know it's there or are scanning around due to lack of response.
String changes that give a visual queue that the AI response is taking time are not read out. This means there is no audio feedback for when a query has completed.
Even with the F6 switch between input and response, this only works reliably if you wait until you see the response return. If you switch to the chat window before the response returns, you usually end up on a time stamp when you switch back and forth.
Using Tab and arrow keys to navigate the chat window will either focus on time stamps or individual characters. In very few occasions, the chat messages actually read out. 

### WCAG 2.2 violation

503.4.2

### Reproduction / example

N/A

### Steps to reproduce

Enable a screen reader. Attempt to use the AI agent.

### Suggested severity

Severity 2 = User cannot complete task, and/or no workaround within the user experience of a given component.

### Code of conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/carbon-ai-chat/issues) for duplicate problems

davidmenendez:
hey @SrfHead thanks for bringing this up. i noticed this same behavior while working on another issue and have opened a PR that i believe will resolve this https://github.com/carbon-design-system/carbon-ai-chat/pull/2347

can you check the [demo app in the deploy preview](https://deploy-preview-2347--carbon-ai-chat-demo.netlify.app/) and verify if the behavior persists? i was able to have the content properly read to me using mac voice over

SrfHead:
@davidmenendez Hi David, thank you for the update. It definitely works better but I found that it stops responding after a few repeat messages. I don't know if this is a demo/narrator/code issue, but entering "Help" multiple times, the narrator stops reading the response.
If you are confident that the screen reader does not get locked in other circumstances, this should resolve the issue I saw.

davidmenendez:
[...]

PR #2347, opened 2026-09-15
fix: decouple aria announcer from chat into provider

### LINKED ISSUE #1933: Move ARIA live-region rendering to a shared host and delete the ai-chat aria directory
# Move ARIA live-region rendering to a shared host and delete the ai-chat aria directory

## Background

The screen-reader announcer logic is already shared: `AriaAnnouncerManager` (`packages/ai-chat-components/src/globals/utils/aria-announcer-manager.ts`) owns rotation, the dual-clear pattern, and the 250 ms NVDA debounce, and both the Lit `chat-shell` and the React app already construct and drive it. What is _not_ shared is the layer around it. `@carbon/ai-chat` keeps a bespoke `packages/ai-chat/src/chat/components/aria/` directory whose `AriaAnnouncerComponent.tsx` hand-rolls the visually-hidden `aria-live` regions (3 polite + 2 assertive), constructs the manager, and adapts React-specific inputs; and every Lit consumer (`chat-shell`, `file-uploads`, `autocomplete`, `audio-player`, `video-player`) independently hand-rolls its own `aria-live` divs plus `new AriaAnnouncerManager()`.

This task lifts the region-creation boilerplate into a small framework-agnostic host beside the manager and adopts it in the React app, so the chat no longer owns a private aria component directory. It extends this epic's "lift the chat-agnostic piece into `@carbon/ai-chat-components`, adopt it, delete the React component" pattern to a piece that is global chat chrome rather than a response type — the same structural move, applied outside the presentational-element set.

The shared piece is an **imperative util** beside the existing `AriaAnnouncerManager` rather than a published `cds-aichat-*` element with a generated React wrapper: it manages a set of live-region DOM nodes plus coalescing timers, with no rendered content to slot or attribute to forward, so an element (and its wrapper/story/CEM overhead) would buy nothing. The React/intl-specific adapter stays behind in a thin wiring layer, so nothing React or store-facing crosses into the shared package.

## Goal

A single framework-agnostic helper in `@carbon/ai-chat-components` creates the visually-hidden live regions and wires them to an `AriaAnnouncerManager`. The React app calls it instead of rendering its own regions, `packages/ai-chat/src/chat/components/aria/` is deleted, and screen-reader behavior is unchanged (3 polite + 2 assertive regions, aria-atomic off, back-to-back identical messages still re-read).

## Public API / contract

New export added to `packages/ai-chat-components/src/globals/utils/aria-announcer-manager.ts` — consumed via the existing deep-import path (`@carbon/ai-chat-components/es/globals/utils/aria-announcer-manager.js`), so no change to `@carbon/ai-chat`'s public surface (`aiChatEntry.tsx` / `serverEntry.ts` / `types/`).

```ts
interface AriaAnnouncerOptions {
  /** Number of polite live regions to create and rotate through. Default 1. */
  politeCount?: number;
  /** Number of assertive live regions to create. Default 0 (polite-only). */
  assertiveCount?: number;
  /** Sets aria-atomic on every created region. Default false. */
  ariaAtomic?: boolean;
}

interface AriaAnnouncerHandle {
  /** Queue a message; same-tick calls coalesce and route to the given channel. */
  announce(message: string, politeness?: "polite" | "assertive"): void;
  /** Cancel pending announcements, disconnect the manager, and remove the created regions. */
  disconnect(): void;
}

/**
 * Create visually-hidden `aria-live` regions inside `container`, wire them to a
 * new `AriaAnnouncerManager`, and return a handle. The caller owns hiding the
 * container (e.g. a visually-hidden wrapper); this only creates the live regions.
 */
function mountAriaAnnouncer(
  container: HTMLElement,
  options?: AriaAnnouncerOptions,
): AriaAnnouncerHandle;
```

The React/intl/DOM-timing adapter stays in `@carbon/ai-chat` — resolving an `AnnounceMessage`/`Node`/`string` to text (`intl.formatMessage`, `nodeToText`) and the microtask node-flush that lets same-tick event-bus-populated Lit descendants settle before conversion are inherently React-specific and do not belong in the shared package.

## Acceptance criteria

- [ ] `mountAriaAnnouncer` (+ `AriaAnnouncerOptions`, `AriaAnnouncerHandle`) added beside `AriaAnnouncerManager`, with WTR coverage: regions created at the requested counts, `announce` routes polite vs assertive to the right regions, `disconnect` cancels pending timers and removes the created regions.
- [ ] `packages/ai-chat/src/chat/components/aria/` is deleted (both `AriaAnnouncerComponent.tsx` and `AriaLiveMessage.tsx`).
- [ ] `AriaAnnouncerProvider` renders a single visually-hidden container and drives it through `mountAriaAnnouncer({ politeCount: 3, assertiveCount: 2 })`; the intl/`Node`/`string` adapter and same-tick node-flush move into the provider (or a React-local util), preserving today's region counts and the deliberate aria-atomic-off choice.
- [ ] `AriaLiveMessage`'s two callers (`ProcessingWithText`, `ConfirmModal`) are rewired to `useAriaAnnouncer` (via an effect or a small hook relocated under `components/util/`); no import of the removed directory remains anywhere.
- [ ] Screen-reader smoke test (NVDA + VoiceOver): polite status updates and assertive blocking-error announcements still fire, and back-to-back identical messages are still re-read.
- [ ] `npm test` + `npm run build` green for both `@carbon/ai-chat` and `@carbon/ai-chat-components`; the SDK-boundary import specs are unaffected (the change is confined to view-layer modules and the already-allowed manager import).

## Out of scope

- Adopting `mountAriaAnnouncer` in the 5 Lit consumers (`chat-shell`, `file-uploads`, `autocomplete`, `audio-player`, `video-player`) — they keep their hand-rolled regions; a follow-up can collapse them onto the shared host.
- Any change to `AriaAnnouncerManager`'s existing rotation, dual-clear, or 250 ms behavior.
- Any change to `AnnounceOnMountComponent` or to `@carbon/ai-chat`'s public API.

## Related

- Parent epic: #1803 (1.x — Response types: publish the presentational elements). Lands on 1.x; no public-API change and no dependency on the 2.0 cut. Sibling reference: the post-2.0 React→Lit sweep #1863 covers the same kind of move for the response-type tree.
- Shared manager: `packages/ai-chat-components/src/globals/utils/aria-announcer-manager.ts` and its WTR test.
- React glue that survives: `AriaAnnouncerProvider`, `AriaAnnouncerContext`, `useAriaAnnouncer`, `withAriaAnnouncer`.

davidmenendez:
Closes #1933 & #2322 

Decouples the aria announcer functionality to be usable globally in different environments / frameworks.

Also adds a fix I noticed that persisted on `main` where the entire message response in the demo app wasn't being read due to a timing issue where the markdown renderer was firing after the announcer fired.

#### Changelog

**Changed**

- `packages/ai-chat-components/src/globals/utils/aria-announcer-manager.ts` refactors aria announcer util to be more universal
- `packages/ai-chat/src/chat/providers/AriaAnnouncerProvider.tsx` implements the new `aria-announcer-manager` and includes `waitForMarkdownElements` to handle ensuring that the message is fully rendered before being sent to the announcer.

**Removed**

- `packages/ai-chat/src/chat/components/aria/AriaAnnouncerComponent.tsx`
- `packages/ai-chat/src/chat/components/aria/AriaLiveMessage.tsx`

#### Testing / Reviewing

run demo app. search for `cds-aichat--aria-announcer` in the element console. verify that the correct number of children are present, which should be 5

```
    <div aria-live="polite"></div>
    <div aria-live="polite"></div>
    <div aria-live="polite"></div>
    <div aria-live="assertive"></div>
    <div aria-live="assertive"></div>
```


packages/ai-chat/src/chat/components/modals/ConfirmModal.tsx
```diff
@@ -210,4 +216,19 @@ class ConfirmModal extends Component<
   }
 }
 
+/**
+ * Thin functional helper that announces the modal's appearance to screen readers
+ * on mount and whenever the message changes. Uses `useAriaAnnouncer` in place of
+ * the removed `AriaLiveMessage` component so no separate aria directory is needed.
```
warrenmblood: Might be better to only describe the way this currently works without explaining the history.

warrenmblood (approved):
LGTM!

PR #2361, opened 2026-09-15
feat: custom header writeable element

### LINKED ISSUE #1956: Add the CUSTOM_HEADER writeable element and the replacement-slot mechanism
### Background

Parent epic: #1955, itself under the AI Chat 2.0 epic #1486.

`HeaderConfig` is the only way to shape the chat header. A host whose design doesn't fit it has to set `isOn: false` and rebuild the header outside the chat, losing the shell's layout, its max-width handling, and the way panels and the workspace slide over it. The writeable elements that touch the header today only augment the one we render — `HEADER_BOTTOM_ELEMENT` adds a row beneath it, `HEADER_FIXED_ACTIONS_ELEMENT` adds buttons inside it.

This issue adds a whole-surface replacement, and lands the replacement-slot mechanism #1957 builds on. See the parent epic for the decisions already settled — naming, replacement-not-composition, which config survives, how nested slots behave, and where presence detection lives.

### Goal

`WriteableElementName.CUSTOM_HEADER` exists. When the host supplies content for it, that content fills the shell's `header` slot and the framework mounts no header of its own; every `HeaderConfig` field is ignored except `isOn`. When the host supplies nothing, today's behavior is unchanged.

### Acceptance criteria

- [x] `WriteableElementName.CUSTOM_HEADER = "customHeader"` is exported, created in `packages/ai-chat/src/chat/services/loadServices.ts`, and reachable through both `instance.writeableElements` (web components) and `renderWriteableElements` (React).
- [x] With content in the slot, the shell's `header` slot holds only the host content and no `Header` is mounted (`packages/ai-chat/src/chat/AppShell.tsx:873`).
- [x] With no content in the slot, the out-of-the-box header renders exactly as it does today — verified on both host paths, including a web-component host that never touches the slot.
- [x] `HeaderConfig.isOn: false` hides the header whether or not host content is present, including when the history panel is in its mobile layout — see the mobile-history criteria below.
- [x] Every other `HeaderConfig` field is inert while host content is present: `title`, `name`, `actions`, `menuOptions`, `showAiLabel`, `hideDefaultAiLabelContent`, `showRestartButton`, `hideMinimizeButton`, `minimizeButtonIconType`, and `hasContentMaxWidth`.
- [x] `HEADER_FIXED_ACTIONS_ELEMENT` stops rendering when the header is replaced — it lives inside the replaced surface (`packages/ai-chat/src/chat/components/header/Header.tsx:451`). `HEADER_BOTTOM_ELEMENT` and `HOME_SCREEN_HEADER_BOTTOM_ELEMENT` keep rendering — they occupy the shell's separate `header-after` slot.
**The mobile-history override**

- [x] The override does not outrank the host. `headerConfigOverride` forces `isOn: true` and injects "new chat" / "view chats" menu options when the history panel is in its mobile layout (`packages/ai-chat/src/chat/AppShell.tsx:492`). With host content present the forced `isOn` no longer applies — `HeaderConfig.isOn` alone decides whether the header area exists (`packages/ai-chat/src/chat/AppShell.tsx:870`) — and the injected menu options go inert with the rest of the config, since there is no framework menu to inject into.
- [x] The host owns the mobile history affordances when it replaces the header, and `packages/ai-chat/docs/Header.md` says so. Point at the controls already documented in `packages/ai-chat/docs/CustomHistory.md` rather than restating them: read `instance.getState().customPanels.history.isMobile` to know when to render them, and call `instance.customPanels.getPanel(PanelType.HISTORY)?.open()` to open the panel. `BusEventType.HISTORY_PANEL_NEW_CHAT` and `HISTORY_PANEL_PRE_OPEN` are notifications the framework fires from its own menu — nothing listens to them and firing them opens nothing, so they are not the host's route.
- [x] A replaced header reads as an implicit `HistoryConfig.showMobileMenu: false`. Setting that field explicitly stays the documented way to say "I supply my own controls" (`packages/ai-chat/docs/CustomHistory.md:128`).
- [x] The chat warns when `history.isOn` holds, the panel is in its mobile layout, `showMobileMenu` is not `false`, and `CUSTOM_HEADER` has content — the built-in mobile history menu no longer renders and the host needs its own. Fire it when that combination first holds, not every render. The warning is debug-only: gate `consoleWarn` on `isEnableDebugLog()` so it appears only under `PublicConfig.debug` and never in a default production build (`packages/ai-chat/src/chat/utils/miscUtils.ts:69`, `:84`). Setting `showMobileMenu: false` acknowledges the swap and silences it.

**The replacement-slot mechanism**

- [x] Presence detection is content-based, not key-based: an eagerly-created empty wrapper div does not count as content. The check runs against the writeable-element node the framework creates in `loadServices.ts` — the one node both host paths write into. React hosts portal into it (`packages/ai-chat/src/chat/components/portals/WriteableElementsPortalsContainer.tsx:75`); web-component hosts append into it. Both containers attach that node into the light DOM (`packages/ai-chat/src/web-components/cds-aichat-container/index.ts:642`, `packages/ai-chat/src/react/ChatContainer.tsx:193`).
- [x] "Meaningful content" matches what `SlotObserver.hasSlotContent` already treats as content (`packages/ai-chat-components/src/components/chat-shell/src/slot-observer.ts:64`): an empty wrapper, a whitespace-only text node, and a comment all read as no content. Match the rules; do not reach into the shell for them — see the shell AC below.
- [x] The swap reacts to content arriving or leaving after boot, not only at first render, via a `MutationObserver` on that node. The framework header remounts cleanly when host content is removed.
- [x] No first-paint flash on either host path. On the web-component path `attachWriteableElements()` runs before `onBeforeRender`, so a host that populates the node there has content in place before the React tree first renders. On the React path portal content lands after mount, so `writeableElementsPresentKeys` (`packages/ai-chat/src/chat/ChatAppEntry.tsx:386`) seeds the first render — as a hint only. It is not a second detection path and does not override the content check after mount.
- [x] The mechanism ships as one shared helper — a hook under `packages/ai-chat/src/chat/hooks/` plus the content predicate — that #1957 imports unchanged for `CUSTOM_PROMPT_LINE`. Nothing about it is header-specific: it takes a `WriteableElementName` and returns whether the host has supplied content for it.
- [x] `@carbon/ai-chat-components` needs no change, and the PR touches no file in it. `cds-aichat-shell` stays agnostic about who fills its `header` slot: its `SlotObserver` answers "is this slot occupied" for layout classes and the initialization gate (`packages/ai-chat-components/src/components/chat-shell/src/shell.ts:553`, `:673`), which the framework's own `Header` already satisfies, so it cannot distinguish host content from ours. Rendering the writeable-element wrapper into `slot="header"` keeps that sniff correct without touching the shell — the recursion in `hasSlotContent` reports an empty wrapper as no content, so `has-header-content` stays off.

**Quality**

- [x] Accessibility: with the header replaced there is no empty or duplicated landmark, the shell's header region labeling still holds, and focus order through header → messages → input is unchanged. Docs tell the host it owns the accessible name and any controls it renders. See `references/accessibility.md`.
- [x] Jest coverage for: replaced vs. not replaced on both host paths, `isOn: false` with content present, the nested-slot expectations above, the post-boot add/remove flip, an empty wrapper reading as no content, and the mobile-history override with content present — `isOn: false` still hides the header, and the debug warning fires only with `debug` on.
- [x] `packages/ai-chat/docs/WriteableElements.md` and `packages/ai-chat/docs/Header.md` document the replacement semantics, the ignored config, and the nested-slot consequences.
- [x] A React example and a web-component example under `examples/` render a custom header, following `examples/AGENTS.md`.

### Public API / contract

```typescript
export enum WriteableElementName {
  // ...existing members unchanged...

  /**
   * An element that replaces the out-of-the-box chat header. When the host
   * supplies content for this slot, the chat renders no header of its own and
   * the content fills the header area directly.
   *
   * Every {@link HeaderConfig} field is ignored while this element has content,
   * except {@link HeaderConfig.isOn} — set it to `false` to hide the header
   * area entirely, host content included. Writeable elements rendered inside
   * the out-of-the-box header are replaced along with it, so
   * {@link WriteableElementName.HEADER_FIXED_ACTIONS_ELEMENT} does not render.
   * {@link WriteableElementName.HEADER_BOTTOM_ELEMENT} and
   * {@link WriteableElementName.HOME_SCREEN_HEADER_BOTTOM_ELEMENT} are
   * unaffected — they sit below the header, not inside it.
   */
  CUSTOM_HEADER = "customHeader",
}
```

`HeaderConfig.isOn`'s JSDoc gains a sentence naming `CUSTOM_HEADER` as the field that still applies when the header is replaced. No new methods, no signature changes. Additive: a `feat`.

The replacement-slot helper is internal. It ships unexported from `aiChatEntry.tsx` and carries no semver commitment. `HistoryConfig` gains no field — `showMobileMenu` already covers "I supply my own controls" — and the debug warning is internal behavior, not API.

### Out of scope

- `CUSTOM_PROMPT_LINE` — #1957, which consumes the mechanism this one lands.
- Changes to `cds-aichat-shell` or `SlotObserver`. The shell stays agnostic; see the epic.
- Panel headers. `cds-aichat-panel`'s `show-chat-header` chrome and the workspace panel header are a separate surface.
- Deprecating `HeaderConfig` or any of its fields. Config stays the supported path for hosts using the header we ship.
- The 2.0 writeable-element renames and removals — #1930.

### Related

- Parent epic: #1955
- Blocks #1957 (shared replacement-slot mechanism; both edit `WriteableElements.ts` and `loadServices.ts`).
- #1930 — the 2.0 writeable-element reshape. Additive here; nothing this issue adds is renamed there.
- #1929 — the 2.0 "host owns the chrome" sub-epic. Same direction, breaking half.

warrenmblood:
Closes #1956

Adds a new writeable element `CUSTOM_HEADER` which allows users to provide a custom header instead of the out-of-the-box one configurable via `HeaderConfig`. When `CUSTOM_HEADER` contains "meaningful" content, the default header won't mount and will be replaced by the writeable element.

#### Changelog

**New**

- Example (React and WC versions) showcasing replacing the default header with a custom element via `WriteableElementName.CUSTOM_HEADER`.
- Like the existing default header, the custom header is removed with `PublicConfig.header.isOn` set to false. However, the effect of `header.isOn` on the default header is overridden when the history panel is present in mobile layout because the header contains the history panel open/close button. When using a custom header, the user becomes responsible for history panel controls and the override logic doesn't apply. In that situation, we simply send a debug warning message (`packages/ai-chat/src/chat/AppShell.tsx`).
- `packages/ai-chat/docs/WriteableElements.md` and `packages/ai-chat/docs/Header.md` document the custom header and its replacement behavior, `HeaderConfig.isOn`, ignored config fields, mobile override logic, nested-slot consequences,  and accessibility requirements.
- `useWriteableElementPresence` hook used to detect if a writeable element contains meaningful or non-empty content. Mirrors what `SlotObserver.hasSlotContent` detects.
- `packages/ai-chat/tests/config/spec/customHeader_spec.tsx`: Coverage for header replaced vs. not replaced, `isOn: false` with content present, the nested-slot expectations, the post-boot add/remove flip, an empty wrapper reading as no content, and the mobile-history override with content present.

**Changed**

- `packages/ai-chat/src/web-components/cds-aichat-container/index.ts`: `attachWriteableElements()` uses `prepend()` instead of `appendChild()` to attach writeable elements to the container element. This places those elements outside of Lit's start/end markers for dynamic content so they survive re-renders of the parent element unchanged. Fixes an issue where the custom header was unexpectedly being removed.
- updates test snapshots with new writeable element slot.

#### Testing / Reviewing

- Run two new examples:
  - `npm run start --workspace=@carbon/ai-chat-examples-react-custom-header`
  - `npm run start --workspace=@carbon/ai-chat-examples-web-components-custom-header`
- A custom header should replace the default one at the top of the chat.
- Verify setting `PublicConfig.header.isOn` to `false` hides the header (neither the custom or default one should render).

- Optional: Check that mobile override does not apply when content added to custom header. Run the `history-float` example and add a `CUSTOM_HEADER` writeable element to replace the default header. Then set `PublicConfig.header.isOn` to `false` and verify that no header renders. Already verified and covered in `customHeader_spec.tsx`.


packages/ai-chat/src/chat/AppShell.tsx
```diff
@@ -543,6 +555,32 @@ function AppShell({
...
+  const customHeaderMobileWarningFired = useRef(false);
+
+  useEffect(() => {
+    if (
+      customHeaderPresent &&
+      publicConfig.history?.isOn &&
+      historyPanelState.isMobile &&
+      (publicConfig.history?.showMobileMenu ?? true) &&
+      !customHeaderMobileWarningFired.current &&
+      isEnableDebugLog()
```
davidmenendez: not major, but i feel like some of these `if` statements are becoming unruly and redundant. i think it could be helpful from a code readability standpoint to refactor these to be more reusable with getter functions or something.

packages/ai-chat/src/web-components/cds-aichat-container/index.ts
```diff
@@ -548,7 +553,7 @@ class ChatContainer extends FlattenedConfigElement {
       element.setAttribute('slot', slot);
 
       if (!element.isConnected) {
-        this.appendChild(element);
+        this.prepend(element);
```
davidmenendez, line 556: with `prepend` the children are actually going to show up in the reverse order of `writeableElements` not sure if that matters or not, but wanted to point that out.

packages/ai-chat/src/chat/hooks/useWriteableElementPresence.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    // Re-check synchronously in case content arrived between the lazy-init and mount.
+    setPresent(hasMeaningfulContent(node));
+
+    const observer = new MutationObserver(() => {
+      setPresent(hasMeaningfulContent(node));
+    });
+
+    observer.observe(node, { childList: true, subtree: true });
```
davidmenendez, line 45: `hasMeaningfulContent` only checks the direct children of the node, but `subtree: true` means that this could fire on deeper child changes. may be relevant. something to keep in mind.

davidmenendez (approved):
LGTM. had a few comments, but nothing that's a deal breaker

PR #2372, opened 2026-09-18
feat(storybook): chat shell input

### LINKED ISSUE #2054: Expand chat shell story to include Prompt line
Update "Input" in Chat shell [story](https://chat.carbondesignsystem.com/components/storybook/tag/latest/index.html?path=/story/preview-chat-shell--default) with Prompt line and the new layout/slots it introduces.

warrenmblood:
Closes #2054

Adds Input stories under `ChatShell` to showcase using the new Prompt Line component inside of Chat shell's "input" slot.

#### Changelog

**New**

- Chat shell stories with "Default" and "Expanded" Prompt line layouts
- Overview page for using Prompt line inside the Chat shell's "input" slot
- `packages/ai-chat-components/src/globals/utils/iconTransform.tsx`: Adds utility `carbonIconToReact` that mirrors util of the same name in the `ai-chat` package. Imports for use in the new stories and existing prompt line ones to reduce duplicate code.

#### Testing / Reviewing

- Go to storybook deploy preview
- Preview -> Chat shell -> Input
- Look through Overview, Default, and Expanded pages
- Check both React & WC versions


ethanwinters (commented):
I think if we are going to make a "Header" and "Input" subsection, which seems reasonable, we should change https://deploy-preview-2372--carbon-ai-chat-components.netlify.app/?path=/story/preview-chat-shell--default to not make use of the header component and perhaps get rid of the "Slot" example since default would then be the same as the slots storybook.

warrenmblood:
> I think if we are going to make a "Header" and "Input" subsection, which seems reasonable, we should change https://deploy-preview-2372--carbon-ai-chat-components.netlify.app/?path=/story/preview-chat-shell--default to not make use of the header component and perhaps get rid of the "Slot" example since default would then be the same as the slots storybook.

I removed the Default story and made "Slots" the new default, since that shows all Chat shell slots without rendering content inside them.

ethanwinters (approved):

